PR #2, opened 2020-11-17
Deploy using development label for osx and linux

OwenArnold:
Creates and pushes `ess` packages to anaconda cloud https://anaconda.org/scipp/scipp_widgets/files


conda/build.sh
```diff
@@ -0,0 +1 @@
+cp -r ess "$CONDA_PREFIX"/lib/python*/
```
Matthew-Andrew, line 1: I think a move will work here but this don't think this matters much.

Matthew-Andrew (approved):
Looks good to me

PR #4, opened 2021-01-05
Imaging loaders should handle directory of FITS files

OwenArnold:
As well as TIFF files currently handled

### DIFF
--- a/ess/v20/imaging/helper_funcs.py
+++ b/ess/v20/imaging/helper_funcs.py
@@ -87,30 +87,32 @@ def export_tiff_stack(dataset, key, base_name, output_dir, x_len, y_len,
     print('Saved tof_of_tiff_{}.txt.'.format(base_name))
 
 
-def tiffs_to_variable(tiff_dir, dtype=np.float64, with_variances=True):
+def _image_to_variable(image_dir, loader, dtype=np.float64,
+                       with_variances=True):
     """
-    Loads all tiff images from the directory into a scipp Variable.
+    Loads all images from the directory into a scipp Variable.
     """
-    stack = _load_tiffs(tiff_dir)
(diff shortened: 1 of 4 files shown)


ess/v20/imaging/helper_funcs.py
```diff
@@ -6,6 +6,7 @@
 import numpy as np
 import scipp as sc
 from tifffile import imsave
```
nvaytet: i wonder why you need both `fabio` and `tifffile`? Can't you read the tiffs also with `tifffile`?

ess/v20/imaging/helper_funcs.py
```diff
@@ -81,7 +97,20 @@ def tiffs_to_variable(tiff_dir, dtype=np.float64, with_variances=True):
     if with_variances:
         return sc.Variable(["t", "spectrum"], values=data, variances=data)
     else:
+        return sc.Variable(["t", "spectrum"], values=data)
+
+
+def fits_to_variable(fits_dir, dtype=np.float64, with_variances=True):
```
nvaytet, line 110: Remove duplicate code between `tiffs_to_variable` and `fits_to_variable` by making a generic function called by both?

nvaytet (changes requested):

ess/v20/imaging/helper_funcs.py
OwenArnold: I think so. There is some history here I don't recall, but I would definitely be in favour of dropping one.

ess/v20/imaging/helper_funcs.py
OwenArnold: I had thought to do that, I think we should do this.

nvaytet (approved):

PR #6, opened 2021-02-03
Changes for imaging notebooks

nvaytet:
- Add v20 beamline chopper parameters
- Add wfm module for analytical frame calculation and tof stitching
- Add peak finding frame calculation from `dress` repository but this has not been tested yet.
- Add groupby2d function used in bragg edge imaging notebook.


src/ess/v20/beamline.py
```diff
@@ -0,0 +1,72 @@
...
+    Convert degrees to radians.
+    """
+    return x * (np.pi * sc.units.rad / 180.0)
+
+
+def _to_angular_frequency(f):
+    """
+    Convert frequency in Hz to angular frequency.
+    """
+    return (2.0 * np.pi * sc.units.rad) * (f / (1.0 * sc.units.s))
```
OwenArnold: if `f` is Hz, should be 1/s already?

src/ess/v20/beamline.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def beamline():
+    """
+    Create V20 chopper cascade and component positions.
```
OwenArnold: beamline method only contains choppers, no other component positions? I see in the notebooks you append additional components to it based on the experiment (a good idea). Should this maybe just be renamed to imply that you are not going to get the full beamline?

src/ess/v20/beamline.py
```diff
@@ -0,0 +1,72 @@
...
+            values=np.concatenate([
+                np.array([94.7, 155.79, 212.56, 265.33, 314.37, 360.0]) + 15.0,
+                np.array([76.03, 141.4, 202.18, 254.97, 307.74, 360.0]) + 15.0,
+                np.array([84.99, 148.29, 205.22, 254.27, 302.8, 360.0]) + 15.0,
+                np.array([116.38, 172.47, 221.94, 267.69, 311.69, 360.0]) +
+                15.0
+            ]).reshape(4, 6)))
+
+    # Length of pulse
+    ds["pulse_length"] = 2.86e+03 * sc.units.us
```
OwenArnold: Isn't this also chopper derived? Or maybe not worth performing, in which case we can comment on why

src/ess/v20/beamline.py
```diff
@@ -0,0 +1,72 @@
...
+        dims=["chopper"],
+        values=["WFM1", "WFM2", "frame-overlap-1", "frame-overlap-2"])
+
+    ds["angular_frequency"] = _to_angular_frequency(
+        sc.array(dims=["chopper"], values=[70.0, 70.0, 56.0, 28.0]))
+
+    ds["phase"] = _deg_to_rad(
+        sc.array(dims=["chopper"], values=[47.10, 76.76, 62.40, 12.27]))
+
+    ds["distance"] = sc.array(dims=["chopper"],
```
OwenArnold: Chopper cutout information is pretty well baked into the beamline, but frequency is an experiment level configuration (should be injected). I believe the WFM choppers are on motors so they can be brought together (or moved further apart) to change the wavelength resolution, so those should be parametrised too. TDC should be fixed for the choppers, but as we've seen TDC is not always correctly known, so maybe that should be parameterised too.

src/ess/v20/beamline.py
```diff
@@ -0,0 +1,72 @@
...
+                              values=[[0, 0, 6.6], [0, 0, 7.1], [0, 0, 8.8],
```
OwenArnold: Not defined what 0, represents - this is not for example a nexus instrument frame, so maybe should define what the origin is. Also assumed that users will know that z is the beam direction. Could we maybe set an attribute (scalar vector_3) to define?

src/ess/v20/imaging/helper_funcs.py
```diff
@@ -109,31 +109,50 @@ def export_tiff_stack(dataset, key, base_name, output_dir, x_len, y_len,
...
+                       reshape=False):
     """
     Loads all images from the directory into a scipp Variable.
     """
     stack = loader(image_dir)
-    data = stack.astype(dtype).reshape(stack.shape[0],
-                                       stack.shape[1] * stack.shape[2])
-    if with_variances:
-        return sc.Variable(["t", "spectrum"], values=data, variances=data)
+    data = stack.astype(dtype)
```
OwenArnold, line 118: I think astype will elide the conversion if target and original are the same?

src/ess/v20/imaging/operations.py
```diff
@@ -139,3 +141,76 @@ def median_from_adj_pixels(data):
...
+                                        data.coords['x']['x',
+                                                         -1].value, nbins + 1),
+                                    unit=data.coords['x'].unit)
+        item.coords["y"] = sc.array(dims=["y"],
+                                    values=np.linspace(
+                                        data.coords['y']['y', 0].value,
+                                        data.coords['y']['y',
+                                                         -1].value, nbins + 1),
+                                    unit=data.coords['y'].unit)
+        for c in ["wavelength", "source-position"]:
```
OwenArnold, line 203: Wouldn't this be more general if we were using any coord not in `x`, `y`. And maybe those two coords could be input args?

src/ess/wfm/__init__.py
```diff
@@ -0,0 +1,2 @@
+from .stitch import stitch
```
OwenArnold: I'm not 100% on having things in the ess package that are outside the `facitlity/instrument/experiment-type` structure. I feel this would quickly decay into all kinds of `misc`, `utility`, `tools` package at every level. While it seems like a regression - I suggest putting all the WFM code alongside v20. The advantages of this I see are:

* We are not overpromising on it's general functionality. If another instrument area attempts to use it, it's clearler that we have only tested this for working with V20. Moderator neutron velocities (constraining) might be critical to avoid frame overlaps for other instruments but we don't need to introduce them for v20. Maybe lots of other assumptions.
* In the former case, it's a trigger to promte it to the `scipp-neturon` package.

src/ess/wfm/frames_analytical.py
```diff
@@ -0,0 +1,151 @@
...
+
+def frames_analytical(instrument=None, plot=False, offset=None):
+    """
+    Compute analytical frame boundaries and shifts based on chopper
+    parameters and detector pixel positions.
+    A set of frame boundaries is returned for each pixel.
+    The frame shifts are the same for all pixels.
+
+    TODO: This currently ignores scattering paths, only the distance from
+    source to pixel. We should use l1 + l2 in the future.
```
OwenArnold: Depends on the technique. For imaging we should not do l1 + l2.

src/ess/wfm/frames_analytical.py
```diff
@@ -0,0 +1,151 @@
...
+    # Now find frame boundaries and draw frames
+    frames = sc.Dataset()
+    frames["left_edges"] = sc.zeros(
+        dims=["frame"] + instrument["position"].dims,
+        shape=[instrument.sizes["frame"]] + instrument["position"].shape,
+        unit=sc.units.us)
+    frames["right_edges"] = frames["left_edges"].copy()
+    frames["shifts"] = sc.zeros(dims=["frame"],
+                                shape=[instrument.sizes["frame"]],
+                                unit=sc.units.us)
```
OwenArnold: This function has two roles (1) compute frame boundaries and shifts (2) plot the results. Can we better seperate the two concerns in the code. I think that having the `plot` flag is not a bad thing, but maybe we should call one sub function to perform the calculations and have a second function to do the plotting (based on the results) provided `plot=True`

src/ess/wfm/frames_analytical.py
```diff
@@ -0,0 +1,151 @@
+import numpy as np
+import matplotlib.pyplot as plt
+from matplotlib.patches import Rectangle
+import scipp as sc
+
+
+def frames_analytical(instrument=None, plot=False, offset=None):
```
OwenArnold: Why default initialise instrument rather than making it a required parameter?

src/ess/wfm/frames_peakfinding.py
```diff
@@ -0,0 +1,237 @@
+"""
+  Find WFM frames using peak-finding.
```
OwenArnold: For people not familiar with the idea - maybe explain the premise that frames should not overlap in time and we can therefore identify minima using fitting to find the centres of the gaps?

src/ess/v20/imaging/operations.py
```diff
@@ -89,7 +90,8 @@ def mean_from_adj_pixels(data):
     container['neighbor', 7:9] = _shift(container['neighbor', 1:3], "y", False,
                                         fill)
 
-    edges_mask = sc.less_equal(container, sc.Variable(value=fill))
+    edges_mask = sc.less_equal(container,
```
OwenArnold: minor point: `<=` rather than named function now we support it?

Not sure why this didn't also require an update to the operations tests.

OwenArnold (changes requested):
Good work thus far - we are very much lacking testing and documentation, but maybe we should immediately work on that together after this has been merged rather than trying to do that all here and otherwise holding the notebook publication up?

src/ess/v20/beamline.py
nvaytet: I would say it's not worth deriving this from choppers, as they are meant to simulate the ESS pulse anyway. I'll add a note as a comment.

src/ess/v20/imaging/helper_funcs.py
nvaytet: i think `astype` always does the conversion.

src/ess/v20/imaging/operations.py
nvaytet: There were hard-coded dimension label everywhere in that file so i left it as is.

src/ess/v20/imaging/helper_funcs.py
OwenArnold: It has to return a `Variable`, so a copy is made even if dtype is the same.

src/ess/wfm/frames_analytical.py
nvaytet: I would say it's not really 2 completely separate things. The plotting is just showing you in a graphical way what was done to compute the frames, a bit like a sanity check.
I can move all the plotting stuff into a separate function but I then have to pass many quantities as arguments, or repeat several calculations to find, e.g. the lengths of the chopper openings in microseconds.
I could move these small calculations to separate functions i guess...

OwenArnold: Yes, separate functions for the calculations sounds good. I think refactoring this is worthwhile given its importance. At present, there are multiple checks for the plotting flag.

src/ess/v20/imaging/helper_funcs.py
nvaytet: Yes, a copy is always made. This is the same as the original code.
I'm not sure I understand the issue here

OwenArnold: my point is whether this is performing unnecessary work in that the dtype does not need to change. It looks like it is and can be avoided. Agree that this was the same in the original code, but fix should be simple.

src/ess/wfm/frames_peakfinding.py
nvaytet: This part of the code has not been tested for at least 6+ months, I've just included here so that all code is in the same place. I don't really want to spend time documenting something that will change again.
I think it's beyond the scope of this PR.
Maybe I can just disable the peak-finding branch for now, to prevent anyone from trying to use it?

nvaytet:
> we are very much lacking testing and documentation

I totally agree about tests and docs. I'll open an issue.

src/ess/v20/beamline.py
nvaytet: Added possibility to pass chopper parameters (frequency, phase, etc..) as a dict to the chopper cascade constructor.

src/ess/v20/imaging/helper_funcs.py
nvaytet: Or maybe this should be a fix that could be implemented in scipp directly, so that we don't have to have `if` statements in all our notebooks?

src/ess/v20/beamline.py
nvaytet: Not sure what you mean here exactly. You mean set an attribute that would contain the origin?
Since this is just for `v20` i don't every see the positions of the choppers moving, so I don't think it's that important. I added a comment saying that the convention was assuming that [0, 0, 0] is the source chopper.

OwenArnold: Yes, that's what I meant, that the something called origin would return the source chopper. Documenting that aspect is fine.

Regarding the beam direction, we have the beam running parallel to z `[0, 0, 1]` my idea was to encode that to make downstream calculations easier (dot product etc).

OwenArnold: I'm ok with both being documented for the time being.

src/ess/v20/imaging/helper_funcs.py
OwenArnold: Without the function returning a view, i'm not sure we can put the fix on the scipp side.

src/ess/v20/imaging/operations.py
OwenArnold: It would be nice to fix this, and move the dimension info to the args - but maybe it can wait for some follow up.

src/ess/v20/imaging/helper_funcs.py
nvaytet: Not sure how you want to implement a quick fix here because
```Py
np.float64 == sc.dtype.float64
```
returns `False`. So I don't really know how to check that the types differ because
```Py
if dtype != stack.dtype:
    stack = stack.astype(dtype)
```
won't work.

OwenArnold: What i've had to do in the past is something like.
`np.Variable(value=1, dtype=np.float64).dtype == sc.dtype.float64` but clearly that's not nice. Just leave it as it is for now.

conda/meta.yaml
```diff
@@ -26,7 +26,7 @@ test:
   requires:
     - pytest
   source_files:
-    - ess/
+    - src/ess/
```
OwenArnold:
```suggestion
    - tests/
```

OwenArnold: Without this pytest won't run anything and the `conda build` step will fail.

OwenArnold (approved):

PR #7, opened 2021-02-04
Set scipp (stable) as upstream dependency

OwenArnold:
part of https://github.com/scipp/scipp/issues/1494

- Change for (stable) scipp dependency
- Set upstream resource (scipp) as trigger for builds.

Note we should still only produce `ess` packages for the `dev` channel (no main) as I don't think this is ready for general consumption yet.

**Tester**

After these changes are merged, kicking off (manually) the `Release` pipeline in `scipp` azure project should result in the `master` pipeline in this project being executed

Note these changes are very similar to those for the `ess-notebooks` https://github.com/scipp/ess-notebooks/pull/26 - note that in that PR, there is a demo run. This is slightly easier to demostrate for the `ess` -> `ess-notebooks` linkage.


.azure-pipelines/master.yml
```diff
@@ -30,7 +37,7 @@ stages:
              conda config --set always_yes yes --set changeps1 no
             displayName: 'create anaconda environment'
           - bash: |
-             conda build conda/ --channel scipp/label/dev --channel conda-forge --label dev --user scipp --token "$ANACONDA_TOKEN" 
+             conda build conda/ --channel scipp/label/main --channel conda-forge --label dev --user scipp --token "$ANACONDA_TOKEN"
```
nvaytet, line 40: Is this the same as just the plain `scipp` channel?

.azure-pipelines/master.yml
```diff
@@ -63,7 +70,7 @@ stages:
              conda config --set always_yes yes --set changeps1 no
```
nvaytet, line 70: Unrelated to this PR, but it would be nice here to do the same as you did for the conda recipe, i.e. remove the duplication between the linux and osx jobs

.azure-pipelines/master.yml
OwenArnold: Yes, I'm just being verbose here.

.azure-pipelines/master.yml
OwenArnold: Agree

PR #9, opened 2021-02-16
Add support for windows package

Matthew-Andrew:


.azure-pipelines/master.yml
```diff
@@ -79,6 +85,35 @@ stages:
...
+              Write-Host "##vso[task.prependpath]$env:CONDA\Scripts"
+              Write-Host "##vso[task.setvariable variable=conda_dir]$env:CONDA"
+            displayName: 'add conda to PATH'
+          - script: |
+              conda --version
+              conda install --yes anaconda-client conda-build
+              conda config --set always_yes yes --set changeps1 no
+            displayName: 'Conda configuration'
+          - script: |
+              conda build conda/ --channel scipp/label/main --channel conda-forge --label dev --user scipp --token "$ANACONDA_TOKEN"
```
OwenArnold: Changes look OK. It's an internal but that prevents us knowing about failure to at the token level (expired) rather than something we can extract via additional flags (verbose or otherwise)? i.e. looks like this problem will reappear in future?

.azure-pipelines/master.yml
```diff
@@ -1,7 +1,13 @@
 trigger:
-  - master
+  branches:
+    exclude:
+      - '*'
 
-pr: none
+pr:
```
OwenArnold: This must be removed before merging and reset to what it was previously.

OwenArnold (changes requested):
Once pipeline trigger changes are reverted this can be merged.

OwenArnold (approved):
Checked them out and they work.

PR #11, opened 2021-02-19
Resolved file path issues

Matthew-Andrew:
This resolves the issues described in https://github.com/scipp/scipp/issues/1709 and https://github.com/scipp/scipp/issues/1708

### DIFF
--- a/src/ess/v20/imaging/helper_funcs.py
+++ b/src/ess/v20/imaging/helper_funcs.py
@@ -38,6 +38,7 @@ def read_x_values(tof_file, **kwargs):
 
 
 def _load_images(image_dir, extension, loader):
+    image_dir = os.path.abspath(image_dir)
     if not os.path.isdir(image_dir):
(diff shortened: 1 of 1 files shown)


src/ess/v20/imaging/helper_funcs.py
```diff
@@ -115,6 +116,11 @@ def _image_to_variable(image_dir,
     Loads all images from the directory into a scipp Variable.
     """
     stack = loader(image_dir)
+
+    if stack.size == 0:
+        print(f'No image files found in {image_dir}')
+        return
```
OwenArnold: This is certainly better. What informed return none, rather than throwing or returning empty numpy array etc?

Matthew-Andrew: Good point probably should do one of those instead. I think throwing is probably better as returning an empty variable is somewhat pointless?

OwenArnold (approved):

PR #13, opened 2021-04-07
Refactor groupby2d to use fold/flatten approach

OwenArnold:
This is looking reasonably generic (maybe apart from actual groupby op, which could be passed as an arg). We might consider moving this to scipp neutron later.

### DIFF
--- a/src/ess/v20/imaging/operations.py
+++ b/src/ess/v20/imaging/operations.py
@@ -141,39 +141,41 @@ def median_from_adj_pixels(data):
     return _median(container, edges_mask, dim='neighbor')
 
 
-def groupby2D(data, nbins):
+def groupby2D(data,
+              nx_target,
+              ny_target,
+              x='x',
+              y='y',
+              z='wavelength',
+              preserve=['source_position']):
 
-    nx_target = nbins
-    ny_target = nbins
+    element_width_x = data.sizes[x] // nx_target
+    element_width_y = data.sizes[y] // ny_target
 
-    element_width_x = data.sizes['x'] // nx_target
-    element_width_y = data.sizes['y'] // ny_target
-
-    x = sc.Variable(dims=['x'],
-                    values=np.arange(data.sizes['x']) // element_width_x)
-    y = sc.Variable(dims=['y'],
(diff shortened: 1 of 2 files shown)


src/ess/v20/imaging/operations.py
```diff
@@ -141,73 +141,38 @@ def median_from_adj_pixels(data):
...
+              preserve=['source_position']):
+
+    element_width_x = data.sizes[x] // nx_target
+    element_width_y = data.sizes[y] // ny_target
+
+    xx = sc.Variable(dims=[x],
+                     values=np.arange(data.sizes[x]) // element_width_x)
+    yy = sc.Variable(dims=[y],
+                     values=np.arange(data.sizes[y]) // element_width_y)
+    grid = xx + nx_target * yy
```
nvaytet, line 153: Are we always sure that this is going to be the spectrum numbering order?
Did you check that the image wasn't flipped in the imaging notebook?
I guess it's fine for this PR, but maybe if we make it part of `scippneutron` there should be a parameter to control the order?

OwenArnold: Agreed. I'll check the notebooks, but I think this is another assumption we'd want to generalise if we "promoted" such functionality to scippneutron.

src/ess/v20/imaging/operations.py
```diff
@@ -141,73 +141,38 @@ def median_from_adj_pixels(data):
...
-    y = sc.Variable(dims=['y'],
-                    values=np.arange(data.sizes['y']) // element_width_y)
-    grid = x + nx_target * y
+def groupby2D(data,
+              nx_target,
+              ny_target,
+              x='x',
+              y='y',
+              z='wavelength',
+              preserve=['source_position']):
```
nvaytet: I don't see `preserve` used anywhere in the function.

nvaytet: Shouldn't this be picked up by `flake8`?

OwenArnold: You're right. We no longer need preserve. It was part of my intermediate refactoring. I'll remove.

nvaytet (approved):

PR #15, opened 2021-04-15
Amor reflectometry

arm61:
This is the _first draft_ of the Amor/reflectometry reduction code. Somethings to note:
- it is not feature complete
- test coverage for the reflectometry modules is 100 % (not including the plotting stuff)
- there is currently no testing for the amor package (this is a todo once the code is feature complete)


arm61:
The CI is failing cause of line length (in comments and docstrings not in code) and one convenience import in a `__init__.py`. Any flexibility of `flake8`?

arm61:
I am not familiar enough with yapf to know what it doesn't like now...seems to be all of the code 🤣

arm61:
🎉

src/ess/reflectometry/resolution.py
```diff
@@ -0,0 +1,46 @@
...
+
+def detector_resolution(spatial_resolution, z_pixel_position,
+                        z_sample_position):
+    """
+    Calculation the resolution function due to the spatial resolution of the detector.
+
+    Args:
+        spatial_resolution (`sc.Variable`): Detector spatial resolution.
+        z_pixel_position (`sc.Variable`): The position of each pixel in the z-dimension.
+        z_sample_position (`sc.Variable`): The position of the sample in the z-dimension.
```
SimonHeybrock: This assumes that the sample-normal vector is in `z` direction? Is there a reasonable way to make this independent, e.g., by requiring to either specify it, or storing it as a coord in the data?

src/ess/reflectometry/plot.py
```diff
@@ -0,0 +1,69 @@
...
+    """
+    Plot 2d histogram of wavelength against theta.
+
+    Args:
+        data (`ReflData`): ReflData object for plotting.
+
+    Returns:
+        (`scipp.plotting.plot2d.SciPlot2d`): 2d histogram of wavelength against theta.
+    """
+    return sc.plot(data.wavelength_theta_bin().bins.sum())
```
SimonHeybrock: By calling `bins.sum()` we limit how useful these functions can be. Is the reason for this https://github.com/scipp/scipp/issues/1564?

src/ess/reflectometry/__init__.py
```diff
@@ -0,0 +1,7 @@
+import scipp as sc
+from scipy.constants import neutron_mass, h, g
+
+h = (h * 1e20 / 1e6 *
+     (sc.units.kg * sc.units.angstrom * sc.units.angstrom / sc.units.us))
+HDM = h / (neutron_mass * sc.units.kg)
+G_ACC = -g * (sc.units.m / (sc.units.s * sc.units.s))
```
SimonHeybrock: Maybe we should make this a vector?

src/ess/amor/amor.py
```diff
@@ -0,0 +1,112 @@
...
+            wavelength_max (`sc.Variable`, optional): Maximum wavelength to be used. Optional, default to `wavelength_min + tau * (HDM / chopper_detector_distance)`.
+        """
+        if wavelength_min is None:
+            wavelength_min = self.wavelength_cut
+        if wavelength_max is None:
+            wavelength_max = wavelength_min + self.tau * (
+                HDM / self.chopper_detector_distance)
+        self.data.bins.masks["wavelength"] = (
+            self.data.bins.coords["wavelength"] < wavelength_min) | (
+                self.data.bins.coords["wavelength"] > wavelength_max)
```
SimonHeybrock: Masking individual event will become very costly very quickly as we get more data. My intention for masking binned data was to use it as follows (assuming you don't have other wavelength binning yet):

```python
wavelength = sc.array(dims=['wavelength'],
                      unit=sc.units.angstrom,
                      values=[wavelength_global_min, wavelength_min, wavelength_max, wavelength_global_max])
binned = sc.bin(binned, edges=[wavelength])
binned.masks['mymask'] = sc.array(dims=['wavelength'], values=[False, True, False])
```

src/ess/amor/amor.py
```diff
@@ -0,0 +1,112 @@
...
+        """
+        Here we correct for the presence of the chopper with respect to the "true" ToF.
+        """
+        self.data.coords["position"].unit = sc.units.m
+        buf = self.data.bins.constituents["data"]
+        tof = buf.coords["tof"].astype(sc.dtype.float64) * 1e-3
+        tof.unit = sc.units.us
+        del buf.coords["tof"]
+        buf.coords["tof"] = tof
+        tof_offset = self.tau * self.chopper_phase / 180.0
```
SimonHeybrock: Assumes phase in in `deg`? Consider using units to avoid bugs?

src/ess/amor/amor.py
```diff
@@ -0,0 +1,112 @@
...
+        tof_e = self.data.bins.constituents["data"].coords["tof"]
```
SimonHeybrock: Same as `tof` above?

src/ess/amor/amor.py
```diff
@@ -0,0 +1,112 @@
...
+        tof_e = (sc.Variable(
+            values=np.remainder(
+                (tof_e - tof_cut + self.tau).values, self.tau.values),
+            unit=sc.units.us,
+            dims=["event"],
+        ) + tof_cut + tof_offset)
+        buf = self.data.bins.constituents["data"]
+        tof = tof_e.astype(sc.dtype.float64)
+        del buf.coords["tof"]
+        buf.coords["tof"] = tof
```
SimonHeybrock: Not clear to me what is done here. What is the remaining being used for?

src/ess/amor/amor.py
```diff
@@ -0,0 +1,112 @@
...
+        sample_angle_offset=0 * sc.units.deg,
+        gravity=True,
+        beam_size=0.001 * sc.units.m,
+        sample_size=0.01 * sc.units.m,
+        detector_spatial_resolution=0.0025 * sc.units.m,
+        chopper_sample_distance=15.0 * sc.units.m,
+        chopper_speed=20 / 3 * 1e-6 / sc.units.us,
+        chopper_detector_distance=1.9e11 * sc.units.angstrom,
+        chopper_chopper_distance=0.49 * sc.units.m,
+        chopper_phase=-8.0 * sc.units.dimensionless,
```
SimonHeybrock: Should this have a `rad` or `deg` unit?

src/ess/amor/amor.py
```diff
@@ -0,0 +1,112 @@
...
+        wavelength_cut=2.4 * sc.units.angstrom,
+    ):
+        """
+        Args:
+            data (`scipp._scipp.core.DataArray`): The data to be reduced.
+            sample_angle_offset (`scipp.Variable`, optional): Correction for omega or possibly misalignment of sample. Optional, default `0 degrees of arc`.
+            gravity (`bool`, optional): Should gravity be accounted for. Optional, default `True`.
+            beam_size (`sc.Variable`, optional): Size of the beam perpendicular to the scattering surface. Optional, default `0.001 m`.
+            sample_size (`sc.Variable`, optional): Size of the sample in direction of the beam. Optional, default `0.01 m`.
```
SimonHeybrock: Is this a variant of Sphinx/rst I am not aware of? If not, can use use Sphinx syntax for the docstrings?

src/ess/reflectometry/corrections.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+def angle_with_gravity(data, pixel_position, sample_position):
+    """
+    Find the angle of reflection when accounting for the presence of gravity.
+
+    Args:
+        data (`scipp._scipp.core.DataArray`): Reduction data array.
+        pixel_position (`scipp._scipp.core.VariableView`): Detector pixel positions, should be a `vector_3_float64`-type object.
+        sample_position (`scipp._scipp.core.VariableView`): Scattered neutron origin position.
```
SimonHeybrock, line 18: Have you considered using the same mechanism as used by `scippneutron.convert`? See https://scipp.github.io/scippneutron/user-guide/unit-conversions.html#Beamline-geometry-parameters-used-in-unit-conversions and https://scipp.github.io/scippneutron/user-guide/overview.html#beamline-geometry? I would generally suggest:

- Store beamline parameters as coords or attrs with data.
- Use [beamline geometry function](https://scipp.github.io/scippneutron/user-guide/overview.html#beamline-geometry) to obtain what is actually needed.

In this case you seem to ignore X (is the beam in Y direction?), but can we just use, e.g., `scn.scattered_beam()` and a `g` vector for the computation?

Even if we store the parameters as coords or attrs, it can still make sense to have this as a function that directly uses the parameters instead of extracting them from a data array. In that case, I would still recommend to work with the "derived" parameters (scattered_beam) instead of underlying positions, since the latter do not actually matter?

src/ess/reflectometry/corrections.py
```diff
@@ -0,0 +1,105 @@
...
+    y_dash = y_dash0(velocity, z_origin, y_origin, z_measured, y_measured)
+    intercept = y_origin - y_dash * z_origin
+    y_true = z_measured * y_dash + intercept
+    angle = sc.Variable(
+        values=np.degrees(
+            sc.atan(y_true / z_measured).bins.constituents["data"].values),
+        unit=sc.units.deg,
+        dims=["event"],
+    )
+    return angle
```
SimonHeybrock: - Working in `deg` seems odd. What is wrong with `rad`? Is this user-facing?
- Is `to_unit` not working for `deg`/`rad` conversion?

src/ess/reflectometry/corrections.py
```diff
@@ -0,0 +1,105 @@
...
+        y_origin (`scipp._scipp.core.Variable`): The y-origin position for the reflected neutron.
+        z_origin (`scipp._scipp.core.Variable`): The z-measured position for the reflected neutron.
+        y_origin (`scipp._scipp.core.Variable`): The y-measured position for the reflected neutron.
+
+    Returns:
+        (`scipp._scipp.core.VariableView` or `array_like`): The gradient of the trajectory of the neutron at the origin position.
+    """
+    return ((-G_ACC * (z_measured - z_origin) * (z_measured - z_origin) /
+             (2 * velocity * velocity)) - y_origin +
+            y_measured) / (z_measured - z_origin)
```
SimonHeybrock: Since this includes `velocity` I assume this contains the `event` dimension (or is a binned variable)? In that case, performance matters, so we should avoid repeated work and memory allocations. Most important is probably to ensure that we do the minimum number of operations with it. As written above, I count 5 operations involving `v**2`.

If I parsed the parenthesis correctly, we can rewrite as `(-0.5 g dz) / (v**2) + dy/dz` which is just 2?

Furthermore, once we did the first operation, we should do the second one in-place, to avoid further memory allocations:

```python
out = (-0.5 * g * dz) / (v**2)
out += dy / dz
return out
```
or something like that.

src/ess/reflectometry/corrections.py
```diff
@@ -0,0 +1,105 @@
...
+    Args:
+        beam_size (:py:attr:`sc.Variable`): Width of incident beam.
+        sample_size (:py:attr:`sc.Variable`): Width of sample in the dimension of the beam.
+        theta (:py:attr:`sc.Variable`): Incident angle.
+
+    Returns:
+        (:py:attr:`array_like`): Correction factor.
+    """
+    beam_on_sample = beam_size / sc.sin(theta)
+    scale_factor = erf((sample_size / beam_on_sample * 2.35482).values)
```
SimonHeybrock: - What is the magic number?
- Return a scipp variable instead of a numpy array?

src/ess/reflectometry/corrections.py
```diff
@@ -0,0 +1,105 @@
...
+    Args:
+        beam_size (:py:attr:`sc.Variable`): Width of incident beam, in metres.
+        sample_size (:py:attr:`sc.Variable`): Width of sample in the dimension of the beam, in metres.
+        theta (:py:attr:`sc.Variable`): Incident angle.
+
+    Returns:
+        (`sc.Variable`): The size of the beam, for each theta, on the sample.
+    """
+    beam_on_sample = beam_size / sc.sin(theta)
+    if ((sc.mean(beam_on_sample)) > sample_size).values:
```
SimonHeybrock:
```suggestion
    if ((sc.mean(beam_on_sample)) > sample_size).value:
```
if this is expected to have only a single value?

src/ess/reflectometry/corrections.py
```diff
@@ -0,0 +1,105 @@
...
+        beam_on_sample = sc.ones(shape=theta.shape,
+                                 unit=sc.units.dimensionless,
+                                 dims=theta.dims) * sample_size
```
SimonHeybrock: Is this doing a broadcast? Can you use https://scipp.github.io/generated/scipp.broadcast.html#scipp.broadcast?

SimonHeybrock (commented):
Part 1 (have not looked at all files yet):

arm61:
> Part 1 (have not looked at all files yet):

Why I opened it now and not when feature complete!

src/ess/reflectometry/resolution.py
arm61: de7b8de

src/ess/reflectometry/plot.py
arm61: What should be used instead?

src/ess/reflectometry/__init__.py
arm61: What benefit would a vector bring?

src/ess/amor/amor.py
arm61: I don't really understand the above code... what is `binned`, it is created from itself?

src/ess/reflectometry/plot.py
SimonHeybrock: If this is 2D data, you can just `sc.plot(data.wavelength_theta_bin())`, it will then dynamically resample the events (and will also work if data is histogrammed). If it is 1D you have to wait for the issue to be fixed.

src/ess/reflectometry/__init__.py
SimonHeybrock: Making us independent of different definitions of where "up" is in the lab, potentially avoiding bad bugs if at any point you make an assumption about this (and someone changes the coord system definition).

src/ess/amor/amor.py
arm61: 6508389

src/ess/amor/amor.py
SimonHeybrock: It is you binned data, `self.data` in this case, I think.

src/ess/amor/amor.py
arm61: This is also not exactly clear to me. The source of neutrons at AMOR is defined by the chopper (they have a continuous source), to the `event_time_offset` in the `.nxs` file is not correct. This is taken from the original code that AMOR provided, but I am speaking to the scientists today to check what is means.

src/ess/amor/amor.py
arm61: 23ccd0c

src/ess/amor/amor.py
arm61: 23ccd0c

src/ess/amor/amor.py
arm61: This is google style sphinx syntax (https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html)

SimonHeybrock: Does this require plugins to Sphinx?

src/ess/reflectometry/corrections.py
arm61: - plotting of theta as a function of reflectivity is a common user operation, and they expect theta to be deg.
- cc55089

src/ess/reflectometry/corrections.py
arm61: Not sure your parsing is correct (see equation 9 https://www.overleaf.com/9574295584nyybsdzmpbxd)

SimonHeybrock: I cannot access that without account.

src/ess/amor/amor.py
arm61: https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html

I think it needs this. I have always used google or numpy style so having `'sphinx.ext.napoleon'` is a given for me. If you are opposed I can change the style.

src/ess/reflectometry/corrections.py
arm61: [AMOR_Reduction-3.pdf](https://github.com/scipp/ess/files/6323345/AMOR_Reduction-3.pdf)

src/ess/reflectometry/__init__.py
arm61: One thing this has made me realise is that I had negative gravity. Nice, accidental, catch (adding a test for this now)

SimonHeybrock: Yet another reason for making it a vector?

src/ess/reflectometry/corrections.py
SimonHeybrock: Your code does not seem to include the final term?

.flake8
```diff
@@ -0,0 +1,11 @@
+[flake8]
```
nvaytet: You can also add `flake8` flags to ignore certain lines or files in the files directly.
I don't think we have agreed on a standard way of doing things, but my opinion would be that it might be safer to not use a separate `.flake8` file, because someone might change the name of a file, or move it into a different subfolder, and then forget to change this file also?

.flake8
```diff
@@ -0,0 +1,11 @@
...
+per-file-ignores =
+    # line too long
```
nvaytet: For the long lines, I would say we should keep lines short, as enforced by `flake8`.
(one simple example is that it makes code much easier to review when lines don't go beyond the edge of the screen)
But we can discuss this. @SimonHeybrock what do you think?

src/ess/amor/__init__.py
```diff
@@ -0,0 +1 @@
+from ess.amor.amor import AmorData
```
nvaytet: We had something like `scipp.plot.plot` in the past, and we decided it was sometimes a little confusing, especially for novice python users. Maybe consider a different name for the `amor/amor.py` file?

src/ess/reflectometry/corrections.py
```diff
@@ -59,7 +59,7 @@ def y_dash0(velocity, z_origin, y_origin, z_measured, y_measured):
     """
     velocity2 = velocity * velocity
     z_diff = z_measured - z_origin
-    return ((-G_ACC * (z_diff) * (z_diff) / (2 * velocity2)) - y_origin + y_measured) / (z_diff)
+    return ((sc.geometry.y(G_ACC) * (z_diff) * (z_diff) / (2 * velocity2)) - y_origin + y_measured) / (z_diff)
```
SimonHeybrock: `-sc.norm(G_ACC)`, in case someone changes the direction?

src/ess/amor/amor.py
```diff
@@ -0,0 +1,111 @@
...
+        self.find_qz()
+        self.illumination()
+
+    def tof_correction(self):
+        """
+        Here we correct for the presence of the chopper with respect to the "true" ToF.
+        """
+        self.data.coords["position"].unit = sc.units.m
+        buf = self.data.bins.constituents["data"]
+        tof = buf.coords["tof"].astype(sc.dtype.float64) * 1e-3
```
nvaytet: This appears to assume some unit for the `"tof"` coord. Should there be a check, or maybe use a `to_unit` here?

src/ess/reflectometry/__init__.py
```diff
@@ -0,0 +1,8 @@
+import scipp as sc
+from scipy.constants import neutron_mass, h, g
```
nvaytet, line 2: Don't we also have a neutron mass in scipp? Is it the same as is `scipy`?

src/ess/reflectometry/corrections.py
arm61: The final term is 0, when z = z_0 (I haven't added this comment to the document just yet).

src/ess/reflectometry/data.py
```diff
@@ -0,0 +1,389 @@
...
+            bins (`array_like`): q-bin edges.
+            unit (`scipp._scipp.core.Unit`): Unit for q. Defaults to 1/Å.
+
+        Returns:
+            (`scipp._scipp.core.DataArray`): Data array binned into qz with resolution.
+        """
+        if "qz" in self.event.coords and "tof" in self.event.coords:
+            if bins is None:
+                q_z_vector = self.event.coords["qz"].values
+                bins = np.linspace(q_z_vector.min(), q_z_vector.max(), 200)
```
nvaytet: I see `np.linspace` a number of times, and I also use it a lot. Sounds like adding a `sc.linspace` could be useful

src/ess/reflectometry/corrections.py
SimonHeybrock: Ok, but then I can't spot the difference between my equation and the one in the document (assuming `dz = z - z0`, etc.)?

nvaytet (commented):
Some small comments

src/ess/amor/amor.py
arm61: Sorry this still isn't clear to me...

src/ess/reflectometry/corrections.py
arm61: Grumble...it took me 2 hours to derive that equation from first principles, my brain was refusing to have it be so simply written 🤣

src/ess/reflectometry/corrections.py
arm61: Yeah, these functions were something that I was not really aware of until yesterday! Will try and put something together that uses them

.flake8
arm61: the long lines are exclusively in docstrings. I have found that line breaks in docstrings can have unexpected results in sphnix (I have always assumed this is why black and yapf don't touch docstrings).

src/ess/reflectometry/__init__.py
arm61: Personally, think that HDM should be a `scippneutron` object that I can import.

.flake8
arm61: I agree about the `.flake8` file (I partly put it in cause the CI won't complete until flake8 is passed), however, I could only work out how to do this on a per line bases and I worry that the `# noqa: E501` would be considered part of the docstring by sphinx (see my point below)

src/ess/reflectometry/corrections.py
arm61: Issue with the `scattered_beam()` approach is that the function doesn't only use the vector from origin to measured but also uses both individually so the function would need all three arguments.

arm61: Something like 
```
def angle_with_gravity(data, sample_to_pixel, sample_position, pixel_position):
    """
    Find the angle of reflection when accounting for the presence of gravity.

    Args:
        data (`scipp._scipp.core.DataArray`): Reduction data array.
        sample_to_pixel (`scipp._scipp.core.VariableView`): Vector of distance from sample to pixel, should be a `vector_3_float64`-type object.
        sample_position (`scipp._scipp.core.VariableView`): Scattered neutron origin position.
        pixel_position (`scipp._scipp.core.VariableView`): Detector pixel positions, should be a `vector_3_float64`-type object.

    Returns:
        (`scipp._scipp.core.Variable`): Gravity corrected angle values.
    """
    # This is a workaround until scipp #1819 is resolved, at which time the following should be used instead
    # velocity = sc.to_unit(HDM / wavelength, 'm/s')
    # At which point the args can be changed to wavelength
    # (where this is data.bins.constituents['data'].coords['wavelength'].astype(sc.dtype.float64) or similar)
    # instead of data
    velocity = sc.to_unit(
        HDM / data.bins.constituents["data"].coords["wavelength"].astype(
            sc.dtype.float64),
        "m/s",
    )
    data.bins.constituents["data"].coords["velocity"] = velocity
    velocity = data.bins.coords["velocity"]
    velocity.unit = sc.units.m / sc.units.s
    z_diff = sc.geometry.z(sample_to_pixel)
    y_diff = sc.geometry.y(sample_to_pixel)
    z_origin = sc.geometry.z(sample_position)
    y_origin = sc.geometry.y(sample_position)
    y_dash = y_dash0(velocity, z_diff, y_diff)
    intercept = y_origin - y_dash * z_origin
    y_true = sc.geometry.z(pixel_position) * y_dash + intercept
    angle = sc.to_unit(
        sc.atan(y_true / z_measured).bins.constituents["data"], 'deg')
    return angle
```
I am probably missing something that fresh eyes can help though

arm61: My aim was to write this function (and all the functions in `corrections.py` and `resolution.py`) as that they can operate independently from the `DataArray` object.

src/ess/amor/amor.py
SimonHeybrock: @OwenArnold @nvaytet Opinions?

src/ess/reflectometry/corrections.py
SimonHeybrock: But shouldn't everything be translation invariant and not depend on actual positions? Why do you need absolute positions?

src/ess/reflectometry/__init__.py
SimonHeybrock: Yes, I have been thinking about adding physical constant for a while, we definitely should.

src/ess/amor/amor.py
arm61: A possible problem: if I use the same approach to mask theta and wavelength the previous mask is lost...(see line 332 onwards in `src/ess/reflectometry/data.py`)

SimonHeybrock: Hmm, what are the dimensions of your binned data? Can you show the html view before and after?

SimonHeybrock: There are cases where we need to apply masks when rebinning, I am not sure this is the case here, it might just be that the logic for that is too zealous.

SimonHeybrock: Can you make sure to rebin the outer dimension first? So if you dims are `['theta', 'wavelength']` and you rebin first `wavelength` then `theta`, I think the `wavelength` mask cannot be preserved. But if you first rebin `theta` it should work (I hope :D).

SimonHeybrock: You can also do both dims at the same time. Or set the mask later. But in any case I would like to see exactly what you have, so I can double check that the logic does what it should.

arm61: opps, I made a mistake actually. Different error now that I will deal with then call this resolved.

arm61: This is now resolved, but the `q_bin()` method is a lot slower now (cause new dimensions have been added to the DataArray, I think).

src/ess/reflectometry/corrections.py
arm61: I need probably rewrite my maths to make it vector-based to get this to work in that framework.

src/ess/amor/amor.py
SimonHeybrock: I need to think about what is happening there. Can you post screenshots of the input to `q_bin` and its output?

src/ess/reflectometry/data.py
```diff
@@ -0,0 +1,405 @@
...
+            edges = sc.array(dims=["qz"], unit=unit, values=bins)
+            binned = sc.bin(self.data, erase=erase, edges=[edges])
+            if "sigma_qz_by_qz" in self.event.coords:
+                qzr = []
+                for i in binned.data.values:
+                    try:
+                        qzr.append(i.coords["sigma_qz_by_qz"].values.max())
+                    except ValueError:
+                        qzr.append(0)
+                qzr = np.array(qzr)
```
SimonHeybrock: - You are including masked events or bins here, is that intentional?
- What is the `ValueError` for? Empty bins?
- Is this the slow part you mention? If you have 3x as many bins I can well imagine a slowdown. I suppose this would be fixed if we add support for `binned.bins.coords['sigma_qz_by_qz'].bins.max()`?

src/ess/amor/amor.py
arm61: It also seems to break my `instrument_view`

arm61: Screenshot 2021-04-16 at 14 05 58

This is a screenshot of the `DataArray`, the result is all 0s in the instrument view

SimonHeybrock: The first odd thing I notice is that you have `detector_id` as well as `theta` as dimensions? Shouldn't it typically be one or the other?

SimonHeybrock: Did you bin in `theta` just to mask? If so, don't do that, instead compute a `detector_id`-dependent mask based on the theta of the detector.

SimonHeybrock: > This is a screenshot of the DataArray, the result is all 0s in the instrument view

Do you get some extra sliders for the extra dimensions? It may be that you are just viewing the first slice which contains no events.

src/ess/amor/amor.py
arm61: Spoke to the AMOR scientist, this is related to the `event_time_offset` being from the pulse generator at SINQ and therefore the AMOR `tof` is not the same. He is confident that this is the correct maths to normalise it.

src/ess/amor/amor.py
arm61: theta is a event object cause of the gravity correction

arm61: There are lots more sliders than there should be but all slices are either nan or 0

src/ess/reflectometry/data.py
arm61: Not intentional, just not clear how best to to this. Ideally a `binned.bins.coords['sigma_qz_by_qz'].bins.max()` function that did if for me 🙂

src/ess/amor/amor.py
SimonHeybrock: So you are saying: While events from the same detector are detected at the same theta, their actual theta is something else? Do we still need the `detector_id` dimension then? It feels slightly confusing (at least when thinking about it initially), but maybe you do need it?

src/ess/reflectometry/data.py
SimonHeybrock: Ok, will put it on the TODO list.

.flake8
nvaytet: See e.g. here to apply it to the entire file: https://github.com/scipp/scipp/blob/main/python/src/scipp/__init__.py#L6
I think you can also add an error code so that it only ignores selected types of errors

arm61: I tried this for selected error types yesterday but couldn't get it to work.

src/ess/amor/amor.py
arm61: Once I find theta, I don't think I need `detector_id`, but I imagine it is necessary for things like `instrument_view`?

SimonHeybrock: The question is: do you still need the instrument view at that point? As far as I understand it is mainly used for visualizing the raw data, but the more we proceed with data reduction the more we would use other visualization options.

arm61: That would involve having >1 `DataArray` which is additional complexity.

SimonHeybrock: Not sure I understand. Maybe we should have a call next week to chat about this and similar things?

arm61: Yeah there is a disconnect somewhere I think 🤣. I will send you a meeting thing.

src/ess/reflectometry/data.py
```diff
@@ -0,0 +1,405 @@
...
+                self.data.bins.attrs['offset_positive'],
+            )
+            angle_min = corrections.angle_with_gravity(
+                self.data,
+                self.data.coords["position"],
+                self.data.bins.attrs['offset_negative'],
+            )
+            del self.data.bins.constituents['data'].attrs['offset_positive']
+            del self.data.bins.constituents['data'].attrs['offset_negative']
+            sigma_theta_position = (angle_max - angle_min) / 2.354820045
```
OwenArnold: What is this hard-coded value used in the division?

src/ess/reflectometry/data.py
```diff
@@ -0,0 +1,405 @@
...
+
+import numpy as np
+import scipp as sc
+import scippneutron as scn
+from ess.reflectometry import corrections, resolution
+
+
+class ReflData:
+    """
+    The general reflectometry data class.
```
OwenArnold, line 17: Until we know it's generic, and have multiple reflectometers I think it probably makes sense to keep everything under `reflectometry` in the `amor` directory. That way we can enforce a simple structure in ess `src/ess/{instrument}`. We have followed this approach with the wavelength-frame-multiplication, which is for the  `v20` instrument only at present, but will in principle be fully generic.

tests/reflectometry/corrections_test.py
```diff
@@ -0,0 +1,138 @@
...
+        z_sample = 0.0 * sc.units.m
+        y_pixel = 1.0 * sc.units.m
+        y_sample = 0.0 * sc.units.m
+        x_pixel = 0.0 * sc.units.m
+        x_sample = 0.0 * sc.units.m
+        pixel_position = sc.geometry.position(x_pixel, y_pixel, z_pixel)
+        sample_position = sc.geometry.position(x_sample, y_sample, z_sample)
+        actual_result = corrections.angle_with_gravity(BINNED, pixel_position,
+                                                       sample_position)
+        assert_almost_equal(
```
OwenArnold, line 76: No major problem with this, but scipp also provides `isclose` `identical` and a range of other comparison functions you can use without having to decompose your variables.

OwenArnold (changes requested):
Nice work! great to see how you're making use of scipp.

src/ess/reflectometry/data.py
arm61: Personally (in part, cause I don't wanna refactor now to put it all in `amor` just to refactor back in X years time when ESTIA and then FREIA are using the code), I think that having this structure "defined" as the general class makes it more likely that this class will be used in future.

tests/reflectometry/corrections_test.py
arm61: I thought this may be the case, but I wasn't aware of them/am comfortable in NumPy.

src/ess/reflectometry/data.py
SimonHeybrock: I think I agree with Andrew here (since this is an ESS library): Putting this into `reflectometry` makes it *more* likely that we will end up with common code between instrument, since no one will want to dig through some instrument folders to find functionality. If this was something going into `scippneutron.reflectometry` I would agree with @OwenArnold, but I think this would be a much later step.

src/ess/amor/amor.py
arm61: I have reverted to the original masking for now. I reckon we should still chat (my plan from this morning didn't pan out when I thought harder about it).

src/ess/reflectometry/corrections.py
arm61: The absolute positions are needed for the resolution function stuff (cause the sample position is not necessarily where the beam was scattered from). The only way to achieve this code with the`scn.scattered_beam()` would involve overwriting the `sample_position` attribute.

.flake8
arm61: I got this working in the end

src/ess/amor/amor.py
arm61: This has been fixed now I think

tests/reflectometry/corrections_test.py
arm61: I will gradually adapt the tests to scipp tests over time.

arm61:
In my opinion, the code is now feature complete. Just getting the last tests in place today then I think I am ready.

tests/reflectometry/corrections_test.py
OwenArnold: Yes, I'm ok with that.

src/ess/reflectometry/data.py
OwenArnold: The new contributing guidelines for the repo should make things a bit clearer after we had an internal discussion. But we are in agreement that having the `reflectometry` directory is OK.

arm61:
That's test coverage back at 100 %.

tests/reflectometry/resolution_test.py
```diff
@@ -0,0 +1,34 @@
...
+
+# author: Andrew R. McCluskey (arm61)
+
+import unittest
+import numpy as np
+from numpy.testing import assert_almost_equal
+import scipp as sc
+from ess.reflectometry import resolution
+
+np.random.seed(1)
```
SimonHeybrock: Hmm, maybe not a good idea at file scope? Not sure what this affects if `pytest` runs multiple files? Where is it even used below?

src/ess/amor/amor_data.py
```diff
@@ -0,0 +1,350 @@
...
+            chopper_sample_distance=chopper_sample_distance,
+            chopper_speed=chopper_speed,
+            chopper_detector_distance=chopper_detector_distance,
+            chopper_chopper_distance=chopper_chopper_distance,
+            chopper_phase=chopper_phase,
+            wavelength_cut=wavelength_cut,
+            data_file=data_file,
+        )
+        self.m_value = m_value
+        supermirror_min_q = ((0.022) * self.event.coords['qz'].unit)
```
SimonHeybrock: Magic number?

src/ess/amor/amor_data.py
```diff
@@ -0,0 +1,350 @@
...
+        supermirror_max_q = self.m_value * supermirror_min_q
+        self.event.coords['normalisation'] = sc.ones(
+            dims=['event'], shape=self.event.data.shape)
+        self.event.masks[
+            'normalisation'] = self.event.coords['qz'] >= supermirror_max_q
+        self.event.coords['normalisation'].values[
+            self.event.coords['qz'].values < supermirror_max_q.
+            value] = self.event.coords['normalisation'].values[
+                self.event.coords['qz'].values < supermirror_max_q.value] / (
+                    1.0 - (0.25 / 0.088) * (self.event.coords['qz'].values[
```
SimonHeybrock: Magic number?

src/ess/amor/amor_data.py
```diff
@@ -0,0 +1,350 @@
...
+        self.sample = sample
+        self.reference = reference
+        if self.reference.data_file is None:
+            self.sample.orso.reduction.comment = 'Live reduction'
+        else:
+            self.sample.orso.reduction.input_files.reference_files = [
+                orso.File(self.reference.data_file)
+            ]
+
+    def q_bin(self, bins=None, unit=(1 / sc.units.angstrom).unit):
```
SimonHeybrock:
```suggestion
    def q_bin(self, bins=None, unit=sc.Unit('1/angstrom')):
```
may be nicer

tests/reflectometry/resolution_test.py
arm61: Lazy copy-paste

src/ess/reflectometry/corrections.py
```diff
@@ -0,0 +1,107 @@
...
+        beam_size (:py:attr:`sc.Variable`): Width of incident beam.
+        sample_size (:py:attr:`sc.Variable`): Width of sample in the dimension of the beam.
+        theta (:py:attr:`sc.Variable`): Incident angle.
+
+    Returns:
+        (:py:attr:`array_like`): Correction factor.
+    """
+    beam_on_sample = beam_size / sc.sin(theta)
+    fwhm_to_std = 2 * np.sqrt(2 * np.log(2))
+    scale_factor = erf((sample_size / beam_on_sample * fwhm_to_std).values)
```
SimonHeybrock, line 83: Just to double check, is this doing what you want to? As it is written it computes
```python
(sample_size / beam_on_sample) * fwhm_to_std
```
is that what you meant?

src/ess/reflectometry/corrections.py
```diff
@@ -0,0 +1,107 @@
...
+        (`sc.Variable`): The size of the beam, for each theta, on the sample.
+    """
+    beam_on_sample = beam_size / sc.sin(theta)
+    if ((sc.mean(beam_on_sample)) > sample_size).value:
+        beam_on_sample = sc.broadcast(sample_size,
+                                      shape=theta.shape,
+                                      dims=theta.dims)
+        # beam_on_sample = sc.ones(shape=theta.shape,
+        #                          unit=sc.units.dimensionless,
+        #                          dims=theta.dims) * sample_size
```
SimonHeybrock: commented code?

src/ess/reflectometry/corrections.py
```diff
@@ -0,0 +1,107 @@
...
+        theta (:py:attr:`sc.Variable`): Incident angle.
+
+    Returns:
+        (`sc.Variable`): The size of the beam, for each theta, on the sample.
+    """
+    beam_on_sample = beam_size / sc.sin(theta)
+    if ((sc.mean(beam_on_sample)) > sample_size).value:
+        beam_on_sample = sc.broadcast(sample_size,
+                                      shape=theta.shape,
+                                      dims=theta.dims)
```
SimonHeybrock, line 103: Is the explicit broadcast required? Since scipp operation broadcast automatically, does it work with just `beaom_on_sample = sample_size`?

src/ess/reflectometry/corrections.py
arm61: Yeah that is correct.

src/ess/reflectometry/data.py
```diff
@@ -0,0 +1,375 @@
...
+            units) / (self.event.shape[0] * sc.units.dimensionless)
+
+    def find_wavelength(self):
+        """
+        From the time-of-flight data, find the wavelength for each neutron event.
+        """
+        self.data.bins.constituents["data"].coords["wavelength"] = (
+            scn.convert(
+                self.data, "tof", "wavelength",
+                scatter=True).bins.constituents["data"].coords["wavelength"])
```
SimonHeybrock, line 142: Is there a reason why this is done by hand, rather than as:
```python
self.data = scn.convert(self.data, "tof", "wavelength", scatter=True)
```
Is it simply to also keep the `tof` coord?

arm61: real answer: I am daft and didn't realise it would work in that way. 

That said, keeping `tof` is useful.

SimonHeybrock: We can consider providing an option for preserving the input coord in `scn.convert()`.

src/ess/reflectometry/data.py
```diff
@@ -0,0 +1,375 @@
...
+                (self.data.attrs["sigma_gamma"] /
+                 self.data.bins.coords["theta"]) *
+                (self.data.attrs["sigma_gamma"] /
+                 self.data.bins.coords["theta"]) +
+                (self.data.bins.attrs["sigma_theta_position"] /
+                 self.data.bins.coords["theta"]) *
+                (self.data.bins.attrs["sigma_theta_position"] /
+                 self.data.bins.coords["theta"]))
+            self.data.bins.constituents["data"].coords[
+                "sigma_theta_by_theta"] = sigma_theta.bins.constituents["data"]
```
SimonHeybrock, line 207: Not now, but we should look into whether we can handle some of these as coord/attrs with variances.

arm61: I will make a note of this.

src/ess/reflectometry/data.py
```diff
@@ -0,0 +1,375 @@
...
+            y_min = sc.min(y_position)
+        if y_max is None:
+            y_max = sc.max(y_position)
+        if z_min is None:
+            z_min = sc.min(z_position)
+        if z_max is None:
+            z_max = sc.max(z_position)
+        self.data.masks["x_mask"] = (x_position < x_min) | (x_position > x_max)
+        self.data.masks["y_mask"] = (y_position < y_min) | (y_position > y_max)
+        self.data.masks["z_mask"] = (z_position < z_min) | (z_position > z_max)
```
SimonHeybrock, line 289:
```python
no_mask = sc.scalar(True)
mask = x_position == x_position
self.data.masks['x'] = mask
self.data.masks['y'] = mask
self.data.masks['z'] = mask
a['x', x_min:x_max].masks['x'] &= no_mask
a['y', y_min:y_max].masks['y'] &= no_mask
a['z', z_min:z_max].masks['z'] &= no_mask
```
Nicer? Yes, this should work even if you have `None` for some of the inputs.

SimonHeybrock: No, sorry, scratch that, this would only work if you had separate `x`, `y`, and `z` coords. Leave it as it is.

SimonHeybrock (approved):
Nice work for a "first draft" 👍 

I assume there will be many more changes in the future, but I don't see anything wrong with merging this now.

arm61:
🎉 🎉 🎉

PR #19, opened 2021-04-22
Improving Amor documentation at function level

arm61:
Also adding some documentation things to autodoc the reflectometry/amor methods.


SimonHeybrock (approved):
Docs is better than no docs, so I see no reason for holding this up, even if we change the structure later.

PR #20, opened 2021-04-23
Add documentation deploy script

nvaytet:
This will only show up on github pages once it is merge into main


.azure-pipelines/documentation_deploy.yml
```diff
@@ -0,0 +1,79 @@
+parameters:
+  - name: release
```
OwenArnold, line 2: Shouldn't this be wired up? This isn't set from anywhere yet.

nvaytet: I guess it would be wired up once we have a proper `release` pipeline.
Right now, i'm about to delete the `release.yml` file. But once we have one, it will just be set as
```
- stage: 'deploy_documentation'
    displayName: 'Deploy documentation'
    jobs:
      - template: documentation_deploy.yml
        parameters:
          release: true
```

OwenArnold (approved):

PR #21, opened 2021-04-23
runtime __version__ info

OwenArnold:
Expose runtime version information. This must be consistent with the package version information.

### DIFF
--- a/src/ess/_version.py
+++ b/src/ess/_version.py
@@ -3,10 +3,9 @@ import os
 
 def _version():
     dir_path = os.path.dirname(os.path.realpath(__file__))
-    git_root = os.path.join(dir_path, '..', '..')
     try:
         import git
-        g = git.cmd.Git(git_root)
+        g = git.cmd.Git(dir_path)
         g.fetch()
         return g.describe('--tags')
     except (ImportError, git.exc.GitError):
--- a/src/ess/_version.py
+++ b/src/ess/_version.py
@@ -1,0 +1,20 @@
+import os
+
+
+def _version():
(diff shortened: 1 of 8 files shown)


src/ess/_version.py
```diff
@@ -0,0 +1,20 @@
+import os
+
+
+def _version():
+    dir_path = os.path.dirname(os.path.realpath(__file__))
+    git_root = os.path.join(dir_path, '..', '..')
```
nvaytet: I'm not sure i understand why you need to go back to the root?
Don't `git` commands work in any directory? Having this hardcoded `../..` path seems like it will break in the future..

OwenArnold: I think you're right. If I understand `gitpython` correctly it can also handle this correctly. I'll remove.

nvaytet (approved):

nvaytet:
@OwenArnold apparently the failed import raises a `ModuleNotFoundError` which is not caught by the `try/except`
https://dev.azure.com/scipp/ess/_build/results?buildId=4786&view=logs&j=463cc57e-0c35-55d4-fcef-59d014c9c24e&t=4c5a4896-430a-59f6-9855-eba00dfc8271

OwenArnold:
> @OwenArnold apparently the failed import raises a `ModuleNotFoundError` which is not caught by the `try/except`
> https://dev.azure.com/scipp/ess/_build/results?buildId=4786&view=logs&j=463cc57e-0c35-55d4-fcef-59d014c9c24e&t=4c5a4896-430a-59f6-9855-eba00dfc8271

I didn't see this when I was using conda-build locally, but should be easy to fix.

Issue #22, opened 2021-04-27
Reflectometry binning code should use `scipp` objects


arm61:
Currently all of the binning code in `src/ess/reflectometry` and `src/ess/amor` uses NumPy arrays for binning of data. This is less optimal and more error prone as they lack units and dimensions. 

Once `scipp` has functions similar to `np.linspace` (https://github.com/scipp/scipp/issues/1865), everything should be converted to these `scipp` objects.

nvaytet:
https://github.com/scipp/scipp/pull/1871 has now been merged

arm61:
Hmmm...sounds like I have work to do 😬

PR #23, opened 2021-04-27
add custom amor q grid function

arm61:
Following some feedback from that Amor instrument scientist, I have added a custom grid definition function as a helper function. 

Currently this returns a NumPy array, but in future this will be converted to a `scipp` object (see #22).


src/ess/amor/amor_data.py
```diff
@@ -365,3 +365,39 @@ def write_wavelength_theta(self, filename, bins):
             bins (:py:attr:`tuple` of :py:attr:`array_like`): wavelength and theta edges.
         """
         write.wavelength_theta(self, filename, bins, self.sample.orso)
+
+
+def q_grid(q_min=0.008, q_fix=0.005, q_max=0.08, d_q=None):
```
nvaytet: It seems like this should maybe move to a different place than `amor_data.py`, which until now seems to contain classes that hold data.
This is more like a tool or utility?
And the amor_data file is starting to get quite long.

arm61: This is a very good comment. Split it out a bit now.

arm61:
I think it is best to merge this before dealing with #22. So that this is available to the scientist soon while the stuff in https://github.com/scipp/scipp/pull/1871 makes its way to a release. Thoughts?

arm61:
This has been rebased and added a bug fix that effected the testing on the DMSC infrastructure.

arm61:
@nvaytet if there are no issues can this be merged in. Then I will start on #22

nvaytet:
The last thing I wanted to check with the rest of the scipp team was the formatting rules.
I see you are just disabling some flake8 checks at the top of your files.
Our feeling is that we shouldn't disable checks if it is not necessary, but we also see that the 80 character limit is a bit painful to work with.
But we never came to a firm conclusion. I will ask again and see if we can come up with something fast.

nvaytet:
For now, could you check what the docs would look like if you made the lines fit inside the 79char limit?
I would rather only disable flake8 checks in `__init__.py` files, because I think the more people start contributing, the more messy it's going to get if people just realise "oh i can just disable flake8" and push all sorts of ugliness in the repo.

arm61:
So the issue arises if you need the line break in the type definition of the doctoring, i.e. below I put a line break before the word `optional` (which is column like 78 or somthing) in https://github.com/scipp/ess/blob/d0da21ff22989f6f65c4fd67aa912923cdecd163/src/ess/reflectometry/data.py#L28

Screenshot 2021-05-07 at 08 09 53

arm61:
If the issue of ugliness not dealt with by `yapf` anyway?

nvaytet:
Hmm so why don't i see the issue in e.g. the `plot` function's docstring? https://github.com/scipp/scipp/blob/main/python/src/scipp/plotting/__init__.py#L79
Screenshot at 2021-05-07 22-47-37

nvaytet:
Maybe because of the different syntax we are using for the docstring?

arm61:
You might be right, but changing all of the docstrings is a big task. I have opened an issue for it so I don't forget (https://github.com/scipp/ess/issues/26).

nvaytet:
> You might be right, but changing all of the docstrings is a big task. I have opened an issue for it so I don't forget (#26).

Ok, just remember to add as a task on that PR to remove the lines that disable some `flake8` checks.

nvaytet:
Also remember to add the license identifier (BSD3) at the top of your files.
Don't know if you want to do that here or when you change the docstrings.

arm61:
I have added the licence id this time so that I don't forget!

tests/amor/tools_test.py
```diff
@@ -0,0 +1,50 @@
+# flake8: noqa: E501
```
nvaytet: I'm really sorry for being a pain, 🙏 but we also have license identifiers/headers in the test files.

nvaytet (changes requested):

tests/amor/tools_test.py
arm61: My bad!

nvaytet (approved):

PR #24, opened 2021-04-27
Change license to BSD3

nvaytet:


OwenArnold (commented):
Looks good. Should we increment the version number at this point so we can distinguish old-new licensed versions?

nvaytet:
> Looks good. Should we increment the version number at this point so we can distinguish old-new licensed versions?

Yes, good idea. How do we do this? Do we have to make a new tag on github before we merge this PR?

OwenArnold:
> > Looks good. Should we increment the version number at this point so we can distinguish old-new licensed versions?
> 
> Yes, good idea. How do we do this? Do we have to make a new tag on github before we merge this PR?

Yes. The configuration in meta.yml is already set up to do the rest. But the tag must exist and have been pushed first.

OwenArnold (approved):

PR #25, opened 2021-05-03
use __version__ in orso

arm61:
Taking advantage of the version availability

### DIFF
--- a/tests/reflectometry/orso_test.py
+++ b/tests/reflectometry/orso_test.py
@@ -13,6 +13,7 @@ import numpy as np
 from numpy.testing import assert_almost_equal, assert_equal
 import scipp as sc
 from ess.reflectometry import orso
+from ess import __version__
 import getpass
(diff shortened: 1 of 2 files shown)


tests/reflectometry/orso_test.py
```diff
@@ -13,6 +13,7 @@
 from numpy.testing import assert_almost_equal, assert_equal
 import scipp as sc
 from ess.reflectometry import orso
+from ess import __version__
```
nvaytet: Maybe not very important but is there any risk here of potentially overriding a `__version__` that is already defined?
As opposed to doing `from ess import __version__ as VERSION` like you did above?

nvaytet (approved):

arm61:
Not sure if you are waiting for me to merge (I can't) @nvaytet

nvaytet:
Ok I will merge 🙂

PR #28, opened 2021-05-28
Pin last stable scippneutron

OwenArnold:
Done in same way and for same reasons as this https://github.com/scipp/scippneutron/pull/84

Also see https://github.com/scipp/scipp/issues/1938 for context


OwenArnold:
Going ahead and merging.

PR #31, opened 2021-06-16
Load Component Info

### LINKED ISSUE #30: load_component_info needed for imaging workflows
load_component_info needed.

See https://github.com/scipp/scippneutron/issues/93

OwenArnold:
Fixes #30 

* Loads component information and returns as a `dict`
* Allows reshaping to any 2D (including differing extents) dims (volume preserving must apply)
* Sanity checks inputs including volume

### DIFF
--- a/src/ess/imaging/mantid.py
+++ b/src/ess/imaging/mantid.py
@@ -1,0 +1,43 @@
+import scippneutron as scn
+import scipp as sc
+import numpy as np
+
+
+def load_component_info_to_2d(geometry_file, advanced_geometry=False):
+    """Load geometry information from a mantid Instrument Definition File
+        or a NeXuS file containing instrument geometry. and reshape into 2D
+        physical dimensions.
+
+        This function requires mantid-framework to be installed
+
+        :param geometry_file: IDF or NeXus file
+        :return dictionary of component names to positions,
+            rotations and shapes
+        :raises ImportError if mantid cannot be imported
+        """
+    from mantid.simpleapi import LoadEmptyInstrument
+    ws = LoadEmptyInstrument(Filename=geometry_file, StoreInADS=False)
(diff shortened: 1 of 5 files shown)


src/ess/imaging/mantid.py
```diff
@@ -0,0 +1,57 @@
...
+    """Load geometry information from a mantid Instrument Definition File
+        or a NeXuS file containing instrument geometry. and reshape into 2D
+        physical dimensions.
+
+        This function requires mantid-framework to be installed
+
+        :param geometry_file: IDF or NeXus file
+        :param sizes: Dict of dim to size for output
+        :return dictionary of component names to positions,
+            rotations and shapes
```
SimonHeybrock: Why not `Dataset`?

OwenArnold: Yes information could be made to work as a Dataset. However we are assembling coord information. We can't add coordinates before having data loaded for the dataset (unless storing scalars) so we would have to add our coord information as data variables then when we unpack (on the client side) set as coordinates. I didn't see the benefit of these additional steps. We could have an out arg for dataset for which we want to add the coords, but that's what we had originally in `scippneutron` and it didn't seem like a nice API.

SimonHeybrock: > However we are assembling coord information.

Not sure I agree. Yes, this may be turned into coord information by adding it as coords for data. But considered on its own, most of if is data, isn't it? For example, positions can be data and we could use `groupby().mean()` to compute average positions, etc.

Probably the detector IDs could be considered coordinates, but most other bits should be data (until added as coords to other data themselves)?

OwenArnold: Ok applying functions might be a good argument to having this returned as a Dataset. I'll change.

PR #34, opened 2021-06-30
SANS basics

SimonHeybrock:
Moved and improved from `ess-legacy` repo.

- Combine reduction with and without wavelength bands.
- Avoid hard-coding reduction and grouping dimensions (previously `spectrum` and `layer`).
- Configurable reducer, to support summing all spectra, or summing in groups.
- Use label-based indexing instead of `make_slices` helper.
- Rename `reduce` to `reduce_to_q` since `reduce` shadows a builtin name. Open to better suggestions.

For now this is untried, this will be done as follow-ups when other components get moved.

### DIFF
--- a/src/ess/sans/reduction.py
+++ b/src/ess/sans/reduction.py
@@ -1,0 +1,20 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2021 Scipp contributors (https://github.com/scipp)
+
+
+def reduce_by_wavelength(data, q_bins, groupby, wavelength_bands):
+    # TODO Backup of the coord is necessary until `convert` can keep original
+    wavelength = data.coords['wavelength']
+    data = sc.neutron.convert(data, 'wavelength', 'Q',
+                              out=data)  # TODO no gravity yet
+    data.coords['wavelength'] = wavelength
+    bands = None
+    for i in range(wavelength_bands.sizes['wavelength'] - 1):
+        low = wavelength_bands['wavelength', i]
+        high = wavelength_bands['wavelength', i + 1]
+        band = sc.histogram(data['wavelength', low:high], q_bins)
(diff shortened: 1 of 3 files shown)


wpotrzebowski (changes requested):
1. convert function is from `scipp`not `scippneutron` (old API)
2. `reducer `is not defined here and I presume it is not built-in function
3. I don't quite understand line 30 in reduction.py. Is this, what you meant?

Otherwise, code looks good to me.

src/ess/sans/reduction.py
```diff
@@ -0,0 +1,31 @@
...
+    data.coords['wavelength'] = wavelength
+    bands = None
+    for i in range(wavelength_bands.sizes['wavelength'] - 1):
+        low = wavelength_bands['wavelength', i]
+        high = wavelength_bands['wavelength', i + 1]
+        band = sc.histogram(data['wavelength', low:high], q_bins)
+        band = reducer(band)
+        bands = sc.concatenate(bands, band,
+                               'wavelength') if bands is not None else band
+    bands.coords['wavelength'] = wavelength_bands
```
SimonHeybrock, line 35: @wpotrzebowski 
> I don't quite understand line 30 in reduction.py. Is this, what you meant?

If you process with wavelength bands, your output will have an additional dimension (`dims=['wavelength', 'Q']`). Here we store the band information as a coord for this extra dim.

SimonHeybrock:
>     1. convert function is from `scipp`not `scippneutron` (old API)
> 
>     2. `reducer `is not defined here and I presume it is not built-in function

See update, hope that explains.

wpotrzebowski (commented):
Ok, thanks for explaining. Looks good.  I don't remember this but does `scatter` arg in `convert`  have default True or does it need to be set up explicitly?

wpotrzebowski (approved):

Issue #8, opened 2021-02-08
Tests and docs for WFM code


nvaytet:
#6 added code for stitching WFM data, but lacks documentation and tests.

OwenArnold:
The lack of tests proved problematic around the time of the June 2021 ecosystem release. I had assumed everything was working correctly (all tests passed) and only found out later that the `ess-notebooks` were broken because of some migrations that needed to be made for WFM related code.

nvaytet:
Fixed by #42 and #43

PR #38, opened 2021-07-01
Unit testing wfm

OwenArnold:
* Support easier refactoring for more instruments ODIN imminent 
* Catch issues as scipp/scippneutron evolves
* Share more knowledge about WFM
* Catch bugs and opportunities for improvements
part of #8


src/ess/v20/wfm/frames_analytical.py
```diff
@@ -94,15 +94,20 @@ def frames_analytical(instrument, plot=False, offset=None):
         intercept_min = y0 - (slopes_min[imin] * x0.value * y0.unit)
         intercept_max = y0 - (slopes_max[imax] * x1.value * y0.unit)
 
+        def make_edge(dims, x):
```
nvaytet:
```suggestion
        def _make_edge(dims, x):
```

src/ess/v20/wfm/frames_analytical.py
```diff
@@ -94,15 +94,20 @@ def frames_analytical(instrument, plot=False, offset=None):
...
-        frames["right_edges"]["frame", i] = sc.array(
-            dims=pos_norm.dims,
-            values=((pos_norm - intercept_min).values / slopes_min[imin]),
-            unit=sc.units.us)
-        frames["left_edges"]["frame", i] = sc.array(
-            dims=pos_norm.dims,
-            values=((pos_norm - intercept_max).values / slopes_max[imax]),
-            unit=sc.units.us)
+        frames["right_edges"]["frame", i] = make_edge(
+            pos_norm.dims,
```
nvaytet: use keyword args?

src/ess/v20/wfm/frames_analytical.py
```diff
@@ -94,15 +94,20 @@ def frames_analytical(instrument, plot=False, offset=None):
...
+            (pos_norm - intercept_min).values / slopes_min[imin])
+        frames["left_edges"]["frame", i] = make_edge(
+            pos_norm.dims,
```
nvaytet: use keyword args?

tests/v20/wfm/frames_analytical_test.py
```diff
@@ -0,0 +1,231 @@
...
+                             phase=None):
+    instrument = sc.Dataset()
+    # single pixel set 10m down the beam
+    instrument['position'] = sc.vector(value=[0, 0, 10], unit=sc.units.m)
+    # single chopper set 5m down the beam
+    z_offset = sc.array(dims=['chopper'], values=[5.0], unit=sc.units.m)
+    no_offset = sc.zeros(dims=['chopper'], shape=[1], unit=sc.units.m)
+    instrument['distance'] = sc.geometry.position(no_offset, no_offset,
+                                                  z_offset)
+    # Arbitrary pulse length of 10us in duration
```
nvaytet: it seems there is no default 10us set in the function def.
where does the number come from?

tests/v20/wfm/frames_analytical_test.py
```diff
@@ -0,0 +1,231 @@
...
+            dims=['chopper'], shape=[1])
+    # window closing angle
+    instrument['frame_end'] = sc.ones(dims=['frame'], shape=[1]) * sc.ones(
+        dims=['chopper'], shape=[1]) * window_size
+
+    instrument['choppers'] = sc.array(dims=['chopper'], values=['chopper'])
+    return instrument
+
+
+def _single_chopper_double_window(window_opening_t,
```
nvaytet: Can some of the code from the single and double windows be made common to reduce duplication?

src/ess/v20/wfm/frames_analytical.py
```diff
@@ -94,15 +94,20 @@ def frames_analytical(instrument, plot=False, offset=None):
         intercept_min = y0 - (slopes_min[imin] * x0.value * y0.unit)
         intercept_max = y0 - (slopes_max[imax] * x1.value * y0.unit)
 
+        def make_edge(dims, x):
+            kwargs = {'unit': sc.units.us}
```
SimonHeybrock: Hard-coded unit?

tests/v20/wfm/frames_analytical_test.py
```diff
@@ -0,0 +1,231 @@
...
+    return instrument
+
+
+def test_frames_analytical_one_chopper_one_cutout():
+    offset = 5.0 * sc.units.us  # start of pulse t offset from 0
+    window_opening_t = 5.0 * sc.units.us
+    pulse_length = 10.0 * sc.units.us
+    instrument = _single_chopper_beamline(window_opening_t, pulse_length)
+    frames = frames_analytical(instrument, offset=offset)
+    # Following results best understood sketching simple time-distance
```
nvaytet, line 130: I think this is a really weird setup to have
Screenshot at 2021-07-07 10-05-14
you have neutrons that are traveling back in time...
I would say we should change the phase, and maybe the pulse length, so we can get something that is realistic.

tests/v20/wfm/frames_analytical_test.py
```diff
@@ -0,0 +1,231 @@
...
+        frames['shifts'].data,
+        sc.array(dims=['frame'], values=[5.0], unit=sc.units.us) + offset)
+
+
+def test_frames_analytical_one_chopper_one_cutout_different_pulse_offset():
+    # Offsetting the pulse start time should impact directly on the
+    # frame edges and shifts
+    window_opening_t = 5.0 * sc.units.us
+    pulse_length = 10.0 * sc.units.us
+    for offset in [0.0 * sc.units.us, 5.0 * sc.units.us]:
```
nvaytet: 5us is the same offset as the previous test, so we are not testing anything new here?

tests/v20/wfm/frames_analytical_test.py
```diff
@@ -0,0 +1,231 @@
...
+    # windows are opposite (pi apart)
+    time_between_frames = np.pi * sc.units.rad / freq
+    instrument = _single_chopper_double_window(window_opening_t, pulse_length,
+                                               window_size)
+    frames = frames_analytical(instrument)
+    np.isclose(frames['left_edges'].data.values[0] + time_between_frames.value,
+               frames['left_edges'].data.values[1])
+    np.isclose(
+        frames['right_edges'].data.values[0] + time_between_frames.value,
+        frames['right_edges'].data.values[1])
```
nvaytet: Ideally, it would also be good to have a test with 2 WFM choppers, since this will probably be the norm, and make sure the shifts are computed at the mid-point between the choppers.

nvaytet (changes requested):

nvaytet:
Good to see you didn't have to do many changes for a new beamline!

tests/v20/wfm/frames_analytical_test.py
OwenArnold: Yes, I did this deliberately as the symmetry seems to be the easiest way to reason what is happening without having full time-distance diagrams generated for the unit tests. This is also the reason i've focused on the single-chopper, single cut-out cases. I think this simplicity is reasonable given that this is at the unit-test level, but we would never want to expose/describe at the top-level. In fact, we should probably sanitise inputs at the `get_frames` level to ensure that calculated times cannot be negative and that the velocities cannot be negative. I'm open to suggestions on improving this, provided that maintainers can easily figure things out without schematics.

tests/v20/wfm/frames_analytical_test.py
OwenArnold: Comment left over after refactoring

src/ess/v20/wfm/stitch.py
```diff
@@ -31,18 +31,31 @@ def _stitch_item(item=None, dim=None, frames=None, target=None, plot=True):
     return target
 
 
+def items(x):
```
nvaytet: I don't see where this is being used.

tests/reflectometry/test2.txt
```diff
@@ -0,0 +1,4 @@
+# hello
```
nvaytet: I don't see this file being used anywhere?

tests/v20/wfm/frames_analytical_test.py
```diff
@@ -0,0 +1,226 @@
...
+    window_size = (np.pi / 4.0) * sc.units.rad  # 45 degree opening
+
+    padding = 1.0 * sc.units.us
+    a = _test_single_chopper_single_cutout(offset, padding, pulse_length,
+                                           window_opening_t, window_size)
+    # Now with new larger offset
+    offset += 1.0 * sc.units.us
+    b = _test_single_chopper_single_cutout(offset, padding, pulse_length,
+                                           window_opening_t, window_size)
+    assert sc.all(b['left_edges'].data > a['left_edges'].data).value
```
nvaytet:
```suggestion
    assert allclose(b['left_edges'].data, a['left_edges'].data + 1.0 * sc.units.us).value
```
and the same for the lines below?

tests/v20/wfm/frames_analytical_test.py
```diff
@@ -0,0 +1,226 @@
...
+    return frames
+
+
+def test_frames_analytical_one_chopper_one_cutout():
+    offset = 0.0 * sc.units.us  # start of pulse t offset from 0
+    window_opening_t = 0.5 * sc.units.us
+    pulse_length = 10.0 * sc.units.us
+    window_size = (np.pi / 4.0) * sc.units.rad  # 45 degree opening
+
+    padding = 1.0 * sc.units.us
```
nvaytet, line 149: Could you explain what you mean by padding here?

nvaytet (changes requested):

tests/v20/wfm/frames_analytical_test.py
OwenArnold: absolute time between the pulse finishing and the chopper opening in example. Otherwise the gradients result in division by zero. I have tried to minimise changes to wfm code itself.

tests/reflectometry/test2.txt
OwenArnold: Added by mistake. I need to merge master. I've already made the fix to stop Andrew's code from producing these outputs in the source directory.

src/ess/v20/wfm/stitch.py
OwenArnold: 👍 I still need to wire this in. Will do.

tests/v20/wfm/frames_analytical_test.py
nvaytet: Add this as a comment?

OwenArnold: The function that uses it now has a renamed argument so it should be clearer.

tests/v20/wfm/frames_analytical_test.py
```diff
@@ -0,0 +1,234 @@
...
+    a = _test_single_chopper_single_cutout(offset, padding, pulse_length,
+                                           window_opening_t, window_size)
+    # Now with new larger offset
+    offset += 1.0 * sc.units.us
+    b = _test_single_chopper_single_cutout(offset, padding, pulse_length,
+                                           window_opening_t, window_size)
+    assert allclose(b['left_edges'].data,
+                    a['left_edges'].data + 1.0 * sc.units.us)
+    assert sc.all(b['right_edges'].data > a['right_edges'].data).value
```
nvaytet: `allclose` here also?

nvaytet (changes requested):
Also, I still don't see the `items` being used.

nvaytet (approved):

PR #40, opened 2021-07-26
No next

OwenArnold:
https://github.com/scipp/scipp/issues/2074


nvaytet (changes requested):
We should also remove the `ess-developer-next.yml` and `ess-developer-dev.yml` files to only have one env file left?

nvaytet:
@OwenArnold I made some additional changes. Can you double-check? Thanks.

SimonHeybrock:
Should we try to get this merged? There is also a conflict.

nvaytet (approved):
@OwenArnold , can you have a second look since I also made changes in this PR?

PR #41, opened 2021-08-03
Python line length 88

nvaytet:
Following https://github.com/scipp/scipp/pull/2071


arm61 (commented):
I can't see anything wrong and I prefer longer lines (we have evolved beyond punch cards)

arm61 (approved):

PR #42, opened 2021-08-04
Generalize and cleanup WFM stitching to allow for ODIN

nvaytet:
- New method to determine frame edges based on https://www.sciencedirect.com/science/article/pii/S0168900220308640
- Store the chopper information in the dataset as a `choppers` coordinate, which contains a Dataset
- Separate plotting of time-distance diagram into a new function which accepts the dataset as input (this now calls `get_frames` internally)
- Allow `stitch` to return either a single DataArray where all frames are rebinned into the same tof bins (original behaviour) or return a dict of DataArrays which contain the individual frames (use `merge_frames=False`).


nvaytet:
@OwenArnold I think this is ready for review now.
I've added some release notes (there wasn't actually a file for them).
I will add a docs notebook on how to use the WFM stuff in a second step.

docs/about/release-notes.rst
```diff
@@ -0,0 +1,44 @@
+.. _release-notes:
+
+Release Notes
+=============
+
+Since v0.4
```
OwenArnold: since v0.0.4

FYI I think this next release should correspond to `v0.1`

src/ess/odin/beamline.py
```diff
@@ -0,0 +1,115 @@
...
+
+    Note that there is a typo in the paper for the FOC1 opening angles.
+    These have been corrected here.
+    """
+
+    dim = 'frame'
+    hz = sc.units.one / sc.units.s
+
+    choppers = {
+        "WFMC1":
```
OwenArnold, line 21: One of our discussion items was whether the naming here is good I think. Didn't we discuss having some indirection to access particular sets of choppers that would not rely on the naming? Maybe an inner dictionary/list containing the name of corresponding keys would be helpful, or maybe we nest the choppers under "WFM", "FOC", "BPC" keys?

src/ess/odin/beamline.py
```diff
@@ -0,0 +1,115 @@
...
+                opening_angles_center=sc.array(
+                    dims=[dim],
+                    values=[91.93, 142.23, 189.40, 233.63, 275.10, 313.99],
+                    unit='deg'),
+                opening_angles_width=sc.array(
+                    dims=[dim],
+                    values=[5.74, 8.98, 12.01, 14.85, 17.52, 20.02],
+                    unit='deg')),
+        "WFMC2":
+        Chopper(frequency=sc.scalar(56.0, unit=hz),
```
OwenArnold, line 35: 👍  this looks unambiguous. Do we need to make additional TDC considerations, or is this all wrapped in the phase?

src/ess/odin/beamline.py
```diff
@@ -0,0 +1,115 @@
...
+                    unit='deg')),
+        "FOC1":
+        Chopper(
+            frequency=sc.scalar(42.0, unit=hz),
+            phase=sc.scalar(0.0, unit='deg'),
+            position=sc.vector(value=[0, 0, 8.4], unit='m'),
+            opening_angles_center=sc.array(
+                dims=[dim],
+                values=[81.12, 127.82, 171.60, 212.66, 251.16, 288.85],
+                # values=[81.12, 127.82, 171.60, 212.66, 251.16, 287.50],
```
OwenArnold: Remove dead code

src/ess/odin/beamline.py
```diff
@@ -0,0 +1,115 @@
...
+from ..wfm.choppers import Chopper
+
+
+def make_beamline() -> dict:
+    """
+    ODIN chopper cascade and component positions.
+    Chopper opening angles taken from Schmakat et al. (2020)
+    https://www.sciencedirect.com/science/article/pii/S0168900220308640
+
+    Note that there is a typo in the paper for the FOC1 opening angles.
```
OwenArnold: Might be useful to say a bit more about what the typo is exactly so easier to track down in future

src/ess/v20/imaging/beamline.py
```diff
@@ -0,0 +1,80 @@
...
+
+    choppers = {
+        "WFMC1":
+        Chopper(
+            frequency=sc.scalar(70.0, unit=hz),
+            phase=sc.scalar(47.10, unit='deg'),
+            position=sc.vector(value=[0, 0, 6.6], unit='m'),
+            opening_angles_open=sc.array(
+                dims=[dim],
+                values=np.array([83.71, 140.49, 193.26, 242.32, 287.91, 330.3]) + 15.0,
```
OwenArnold, line 30: Where does +15 come from? maybe add comment

src/ess/wfm/beamline.py
```diff
@@ -0,0 +1,125 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2021 Scipp contributors (https://github.com/scipp)
+import scipp as sc
+import numpy as np
+from .choppers import Chopper, make_chopper_cascade
+
+# class Beamline:
```
OwenArnold: Dead code remove

src/ess/wfm/beamline.py
```diff
@@ -0,0 +1,125 @@
...
+    """
+    Fake chopper cascade with 2 optically blind WFM choppers.
+    Based on mathematical description in Schmakat et al. (2020);
+    https://www.sciencedirect.com/science/article/pii/S0168900220308640
+    """
+
+    dim = 'frame'
+    # Neutron mass to Planck constant ratio
+    # TODO: would be nice to use physical constants in scipp or scippneutron
+    alpha = 2.5278e-4 * (sc.Unit('s') / sc.Unit('angstrom') / sc.Unit('m'))
```
OwenArnold, line 27: Should say that this is neutron mass divided by planks constant (as a comment) or better yet define alpha from those.

src/ess/wfm/beamline.py
```diff
@@ -0,0 +1,125 @@
...
+import numpy as np
+from .choppers import Chopper, make_chopper_cascade
+
+# class Beamline:
+#     def __init__(self, choppers, source):
+#         self.choppers = choppers
+#         self.source = source
+
+
+def make_fake_beamline(
```
OwenArnold, line 8: Why does this live in the working code rather than being part of the tests?

src/ess/wfm/beamline.py
```diff
@@ -0,0 +1,125 @@
...
+        chopper_positions={
+            "WFMC1": sc.vector(value=[0.0, 0.0, 6.775], unit='m'),
+            "WFMC2": sc.vector(value=[0.0, 0.0, 7.225], unit='m')
+        },
+        frequency=sc.scalar(56.0, unit=sc.units.one / sc.units.s),
+        lambda_min=sc.scalar(1.0, unit='angstrom'),
+        pulse_length=sc.scalar(2.86e-03, unit='s'),
```
OwenArnold, line 15: source pulse length. Would help with the mental model below if this was named as such.

src/ess/wfm/beamline.py
```diff
@@ -0,0 +1,125 @@
...
+        pulse_t_0=sc.scalar(1.3e-4, unit='s'),
```
OwenArnold, line 16: source pulse t0 (start of source pulse)

src/ess/wfm/beamline.py
```diff
@@ -0,0 +1,125 @@
...
+#         self.choppers = choppers
+#         self.source = source
+
+
+def make_fake_beamline(
+        chopper_positions={
+            "WFMC1": sc.vector(value=[0.0, 0.0, 6.775], unit='m'),
+            "WFMC2": sc.vector(value=[0.0, 0.0, 7.225], unit='m')
+        },
+        frequency=sc.scalar(56.0, unit=sc.units.one / sc.units.s),
```
OwenArnold, line 13: WFM chopper(s) frequency

src/ess/wfm/beamline.py
```diff
@@ -0,0 +1,125 @@
...
+from .choppers import Chopper, make_chopper_cascade
+
+# class Beamline:
+#     def __init__(self, choppers, source):
+#         self.choppers = choppers
+#         self.source = source
+
+
+def make_fake_beamline(
+        chopper_positions={
```
OwenArnold, line 9: Assumes source chopper at origin (quite important in following calculations)

src/ess/wfm/beamline.py
```diff
@@ -0,0 +1,125 @@
...
+
+    for i in range(nframes):
+        lambda_max = (pulse_length +
+                      alpha * lambda_min * sc.norm(chopper_positions["WFMC1"])) / (
+                          alpha * sc.norm(chopper_positions["WFMC2"]))
+        theta = omega * (
+            pulse_length + alpha *
+            (lambda_min - lambda_max) * sc.norm(chopper_positions["WFMC1"]))
+
+        phi_wfm_1 = omega * (
```
OwenArnold, line 46: Could we have better names or just comments for these so that one does not keep having to cross reference to paper. Alternatively you could refer to the equation in the paper for each of these.

src/ess/wfm/beamline.py
```diff
@@ -0,0 +1,125 @@
...
+        pulse_length=sc.scalar(2.86e-03, unit='s'),
+        pulse_t_0=sc.scalar(1.3e-4, unit='s'),
+        nframes=2):
+    """
+    Fake chopper cascade with 2 optically blind WFM choppers.
+    Based on mathematical description in Schmakat et al. (2020);
+    https://www.sciencedirect.com/science/article/pii/S0168900220308640
+    """
+
+    dim = 'frame'
```
OwenArnold, line 24:
```suggestion
    dim_frame = 'frame'
```
Helps with readability further down

src/ess/wfm/beamline.py
```diff
@@ -0,0 +1,125 @@
...
+    coords = {
+        'choppers': sc.scalar(chopper_cascade),
+        'position': sc.vector(value=[0., 0., 60.], unit='m')
+    }
+    for key, value in source.items():
+        coords[key] = value
+    return coords
+
+
+# def make_coords(**kwargs):
```
OwenArnold: Remove dead code

src/ess/wfm/choppers.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def _extract_and_concatenate(container: dict, key: str, dim: str) -> sc.Variable:
+    array = None
+    for item in container.values():
+        scalar = getattr(item, key)
+        if array is None:
+            array = scalar
+        else:
+            array = sc.concatenate(array, scalar, dim)
```
OwenArnold: Is this ever reached? `array=None` on line 33.

src/ess/wfm/frames_analytical.py
```diff
@@ -0,0 +1,118 @@
...
+    """
+    Compute analytical frame boundaries and shifts based on chopper
+    parameters and detector pixel positions.
+    A set of frame boundaries is returned for each pixel.
+    The frame shifts are the same for all pixels.
+    See Schmakat et al. (2020);
+    https://www.sciencedirect.com/science/article/pii/S0168900220308640
+    for a description of the procedure.
+
+    TODO: This currently ignores scattering paths, only the distance from
```
OwenArnold, line 18: Good point.

src/ess/wfm/frames_analytical.py
```diff
@@ -0,0 +1,118 @@
...
+    use l1 + l2 in the future.
+    """
+
+    # Compute distances for each pixel
+    pos_norm = sc.norm(data.meta["position"])
+
+    # Get the number of WFM frames
+    nframes = data.meta["choppers"].value["opening_angles_open"].sizes["frame"]
+
+    # Now find frame boundaries
```
OwenArnold: Tweak comment - Initialise entries for the frame edges. Finding happens below.

src/ess/wfm/frames_analytical.py
```diff
@@ -0,0 +1,118 @@
...
+    frames = sc.Dataset()
+    frames["left_edges"] = sc.zeros(dims=["frame"] + data.meta["position"].dims,
```
OwenArnold: Maybe have the names for this more verbose. Could easily be thought of as the angle of the left edge of a cutout in radians.

src/ess/wfm/frames_analytical.py
```diff
@@ -0,0 +1,118 @@
...
+    for a description of the procedure.
+
+    TODO: This currently ignores scattering paths, only the distance from
+    source to pixel.
+    For imaging, this is what we want, but for scattering techniques, we should
+    use l1 + l2 in the future.
+    """
+
+    # Compute distances for each pixel
+    pos_norm = sc.norm(data.meta["position"])
```
OwenArnold: Could you include pixel in the name for this variable. Helps further down.

src/ess/wfm/plot.py
```diff
@@ -0,0 +1,131 @@
...
+from .wfm import get_frames
+from .tools import frame_opening_and_closing_times
+
+
+def time_distance_diagram(data: sc.DataArray, **kwargs) -> plt.Figure:
+    """
+    Plot the time-distance diagram for a WFM beamline.
+    This internally calls the `get_frames` method which is used to compute the
+    frame properties for stitching.
+    """
```
OwenArnold, line 17: Maybe say something about what the expected `data` argument is. i.e. this i the beamline + chopper information.

src/ess/wfm/plot.py
```diff
@@ -0,0 +1,131 @@
...
+                va="top")
+
+    ax.plot([0, sc.max(frames["right_edges"].data).value], [det_last.value] * 2,
+            lw=3,
+            color='grey')
+    ax.text(0.0, det_last.value, "Detector", va="bottom", ha="left")
+    ax.set_xlabel("Time [microseconds]")
+    ax.set_ylabel("Distance [m]")
+
+    return fig
```
OwenArnold, line 144: An orthogonal comment - we don't handle the BPC at all either in the frame calculations or the diagrams. Maybe that's something we should do in future as the paper describes their use for skipping pulses (7Hz operation)

src/ess/wfm/tools.py
```diff
@@ -0,0 +1,49 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2021 Scipp contributors (https://github.com/scipp)
+
+import scipp as sc
+
+
```
OwenArnold, line 9: I think we said we were going to use the plot utilities (from scipp) implementation of these when we get a new scipp release - maybe mark as a todo.

OwenArnold (changes requested):
There are some other parts I still need to take a detailed look at.

* Looks much cleaner than what we had before and easier to understand
* Good use of reference for the calculations
* Good structure with instruments providing the bespoke parts and a common WFM framework
* Good use of scipp in the updates

src/ess/wfm/beamline.py
nvaytet: This is exactly the comments that are above?

src/ess/wfm/beamline.py
nvaytet: I wanted to use it in notebooks for the docs, and not sure how to import it if it is in the tests.
I also thought this might have been what the Odin people were after: a tool to create a chopper cascade from user-defined requirements?

src/ess/wfm/beamline.py
nvaytet: I'll reference the equations

src/ess/wfm/tools.py
nvaytet: I don't know how to make a good time-distance diagram with the scipp plotting.
I think it needs to be done manually with matplotlib.

src/ess/wfm/beamline.py
OwenArnold: Yes sorry you're right.

src/ess/wfm/tools.py
OwenArnold: I'm referring to the `to_bin_centers`, `to_bin_edges` general functions below. Not the plotting with is clearly bespoke to the time-distance-diagrams.

src/ess/wfm/beamline.py
OwenArnold: I think the desire would be to output the chopper cascade positions rather than take it as part of the argument set. i.e. given the wavelength resolution.

Good point about the notebooks. Can we at least put this into another examples modules? It seems out of place where it is.

src/ess/wfm/tools.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+    Get coordinates of a chopper frame opening in time and distance.
+    """
+    t_open = _angular_frame_edge_to_time(frame["angular_frequency"].data,
+                                         frame["opening_angles_open"].data,
+                                         frame["phase"].data)
+    t_close = _angular_frame_edge_to_time(frame["angular_frequency"].data,
+                                          frame["opening_angles_close"].data,
+                                          frame["phase"].data)
+    return t_open, t_close
```
OwenArnold: Again probably orthogonal to this, but I wonder if some of the common chopper functionality should be separated from WFM itself. Candiates might include.

* the two functions above relating to chopper times and angles
* calculations that determine neutron velocity/time from wavelenth
* time_distance_diagram

I don't see a need for this at present, but it could be something we want when we start working with direct-inelastic instruments (among others).

docs/about/release-notes.rst
nvaytet: yes i can see that we never made a release. However, ~git describe gives `0.4`, not sure i understand why.~ you're right it's `0.0.4`

src/ess/odin/beamline.py
nvaytet: It's all in the phase and the opening angles.

src/ess/odin/beamline.py
nvaytet: Yes, I had tried to address this by adding an argument to the `frames_analytical` function (https://github.com/scipp/ess/blob/06b36a32db309294a4d5723e378889a810ac4c9a/src/ess/wfm/frames_analytical.py#L10) but I don't really like it.

Now I am thinking of adding a `kind` field to the `Chopper` class, which would just be an `Enum`, and using that to identify the WFM choppers. What do you think?

OwenArnold: Sounds good. Presumably once we know that two choppers are WFM we don't need to know which is the second or the first, either because we are averaging the position, or because we can just use distance from the source to determine which is which?

nvaytet: Yes, I am already using the distance to do this: https://github.com/scipp/ess/blob/06b36a32db309294a4d5723e378889a810ac4c9a/src/ess/wfm/frames_analytical.py#L54

src/ess/wfm/beamline.py
nvaytet: I don't think this assumes anything about the source. Differences between source position and chopper positions are taken further along the workflow.

src/ess/wfm/beamline.py
nvaytet: I don't see any other `dim` in that function or file, so i left it as is.

src/ess/wfm/plot.py
nvaytet: Yes, but we are only ever plotting a single pulse, so I think it's fine for now.

src/ess/wfm/tools.py
nvaytet: I think maybe your concerns here were partially solved by refactoring the `Chopper` to do the angle+frequency -> opening times conversion etc.. ?

src/ess/v20/imaging/beamline.py
nvaytet: Done in the functions's docstring

src/ess/wfm/beamline.py
nvaytet: Not sure where else to put it.... any suggestions?

src/ess/wfm/frames_analytical.py
nvaytet: yes i renamed everything. `left_edges` and `right_edges` are now `time_min` and `time_max`

nvaytet:
@OwenArnold I managed to speed up the time-consuming step by using a mask to discard the invalid neutrons.
Can you have another look at this? I think I addressed all your comments.
Thanks!

src/ess/wfm/choppers.py
```diff
@@ -12,63 +19,140 @@ def __init__(self,
...
-    for item in container.values():
-        scalar = getattr(item, key)
-        if array is None:
-            array = scalar
+                 opening_angles_close: sc.Variable = None,
+                 kind: ChopperKind = None):
+        self._frequency = frequency
+        self._position = position
+        self._phase = phase
+        self._opening_angles_center = opening_angles_center
```
OwenArnold, line 27: We don't have a docstring or otherwise to explain that not all of these inputs are mandatory and that they can generate conflict if they do not match.

tests/wfm/chopper_test.py
```diff
@@ -0,0 +1,202 @@
...
+    assert str(e_info.value) == "Chopper opening angles are not monotonic."
+
+
+def test_chopper_constructor_bad_open_angles(params):
+    opening_angles_open = params[
+        'opening_angles_center'] - 0.5 * params['opening_angles_width']
+    opening_angles_close = params[
+        'opening_angles_center'] + 0.5 * params['opening_angles_width']
+    opening_angles_open.values = opening_angles_open.values[[1, 0, 2, 3, 4, 5]]
+    with pytest.raises(ValueError) as e_info:
```
OwenArnold, line 104: Minor. You might consider adding custom exceptions for some of the failure modes in the `Chopper` this is probably as many lines of code as we have already, and then we don't rely on fragile text based matching.

OwenArnold (approved):
Some minor comments but those could be addressed in a follow up if required.

PR #43, opened 2021-08-14
Add wfm intro notebook

nvaytet:
Add notebook to docs as intro to WFM.

There are a number of figures in this notebook that are used to describe the WFM method, they use matplotlib directly, not the scipp plotting. They are mostly copied from the paper by Schmakat et al. (2020); https://www.sciencedirect.com/science/article/pii/S0168900220308640

I wanted the notebook to be self-contained, and also to have interactive figures one could zoom in, instead of having static images. I also did not want to scrape the images from the paper, as it is owned by the journal and there is no open-access copy available.

It ended up being quite a lot of code in the notebook, and I felt it was removing focus from the discussion.
I searched for a good way to have a toggle button to hide/show the code (and hide it by default), but I could not find something that would also work in the sphinx docs.

I thought of adding a small python file that contains the plotting functions as helpers, that would be imported locally by the notebook, but the MPL code was not very clean and I didn't want to pollute the `ess` codebase.

In the end, I settled with creating a new repo in the Scipp organisation, with a folder structure that follows the `docs` folder structure `ess/techniques/wfm`.
I am then importing this by downloading the file from the notebook (using `requests`) into the same folder:
```Py
# Fetch helper functions to create the figures from an external repo
import requests
remote_url = 'https://raw.githubusercontent.com/scipp/documentation-helpers/main/ess/techniques/wfm/introduction-to-wfm.py'
local_file = 'dochelpers.py'
data = requests.get(remote_url)
with open(local_file, 'w') as f:
    f.write(str(data.content.decode()))
from dochelpers import *
```


docs/techniques/wfm/introduction-to-wfm.ipynb
```diff
@@ -0,0 +1,707 @@
...
+    "$$\\frac{\\Delta \\lambda}{\\lambda} = \\frac{t_{P}}{t} = \\frac{t_{P}}{\\alpha \\lambda z_{\\rm det}} ~, ~~~~~(1)$$\n",
+    "where $\\lambda$ is the neutron wavelength, $t$ is time,\n",
+    "$\\alpha = m_{\\rm n}/h = 2.5278 \\times 10^{-4}~{\\rm s}\\unicode{x212B}^{-1}{\\rm m}^{-1}$\n",
+    "is the ratio of the neutron mass and the Planck constant,\n",
+    "and $z_{\\rm det}$ is the distance from the source to the detector.\n",
+    "\n",
+    "A natural consequence of this is that the wavelength resolution $\\Delta \\lambda / \\lambda$\n",
+    "becomes finer with increasing wavelength.\n",
+    "\n",
+    "This also means that the resolution can be pretty poor for a long-pulsed source such as the ESS.\n",
```
OwenArnold:
```suggestion
    "This also means that the resolution is poor for a long-pulsed source such as the ESS compared to their short-pulse counterparts, such as ISIS.\n",
```

docs/techniques/wfm/introduction-to-wfm.ipynb
```diff
@@ -0,0 +1,707 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "f21a6a9e",
+   "metadata": {},
+   "source": [
+    "As illustrated in Fig. 1, with a long pulse, two neutrons with very different wavelengths\n",
+    "($\\lambda_{1} < \\lambda_{2}$) can reach the detector at the exact same time,\n",
+    "if they originated from a different part of the pulse.\n",
+    "\n",
+    "This has the effect of merging many different wavelengths together on the same detector pixel,\n",
```
OwenArnold: Can we reword this. Maybe something along the lines of:

The problem cannot be resolved at the detector. According to the difference between time recorded at the detector and the measured start of the source pulse, both neutrons have the same wavelength. The detector recording system has no way of knowing that this is not the reality or adjusting for this. We instead look at the pulse generation to find ways to better measure the wavelength of our neutrons.

docs/techniques/wfm/introduction-to-wfm.ipynb
```diff
@@ -0,0 +1,707 @@
...
+   "id": "6f3026e9",
+   "metadata": {},
+   "source": [
+    "## So what is WFM anyway?\n",
+    "\n",
+    "Within the concept of wavelength frame multiplication (WFM),\n",
+    "each source pulse is chopped into a number of sub pulses referred to as wavelength frames,\n",
+    "where each wavelength frame $N$ contains a subsequent part of the spectrum of the source pulse.\n",
+    "\n",
+    "The main reason for using the WFM concept is to redefine the burst time $t_{P} = \\Delta t$ as implied by Eq. (1), in order to match the required wavelength resolution of the experiment."
```
OwenArnold: A secondary objective, (or constraint of the first) is to utilise as much of the source pulse as possible.

Side discussion. As you improve reduce dLam/Lam you trade off flux. I'm not sure if you go on to derive or discuss those considerations. - My initial guess is that you can derive this as a function of the intersection of t_opening for both choppers as a ratio of the inverse chopper frequency.

docs/techniques/wfm/introduction-to-wfm.ipynb
```diff
@@ -0,0 +1,707 @@
...
+   "source": [
+    "Figure3"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "8e6c9c22",
+   "metadata": {},
+   "source": [
+    "Here we define several important quantities:\n",
```
OwenArnold, line 163: You might want to say that FOCs (not yet introduced) control  introduce filtering to limit the neutrons to fall between `t0`, `tA` after they pass through the WFM chopper system.

docs/techniques/wfm/introduction-to-wfm.ipynb
```diff
@@ -0,0 +1,707 @@
...
+    "\n",
+    "- The pulse $t_0$ is defined as the point in time at which the pulse is bright enough for the purposes of the experiment.\n",
+    "- $t_{\\rm A}$ is the time when the flux has fallen down to a level below the required brightness.\n",
+    "- $t_{\\rm P} = t_{\\rm A} - t_{0}$ is the portion of the pulse that is used for the measurements, the analog of the pulse length for the ideal rectangular pulse above.\n",
+    "- $t_{\\rm B}$ marks the end of the pulse; i.e. the time when the flux is considered to be effectively zero.\n",
+    "\n",
+    "### Using a single WFM chopper\n",
```
OwenArnold, line 170: 👍  for logically working through the chopper system like this. It can get confusing when we consider the whole cascade rather than the choppers one by one.

docs/techniques/wfm/introduction-to-wfm.ipynb
```diff
@@ -0,0 +1,707 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "7dcbd2b3",
+   "metadata": {},
+   "source": [
+    "The relative spectral resolution $\\Delta \\lambda / \\lambda$ at the detector position $z_{\\rm det}$\n",
+    "is defined by the burst time $\\Delta t$ and the time-of-flight $t(\\lambda)$ of the neutrons (see eq. (1)).\n",
+    "Because $\\Delta t$t is independent of the wavelength for the case of a single WFM chopper disc,\n",
```
OwenArnold: `$\\Delta t$t` ?

docs/techniques/wfm/introduction-to-wfm.ipynb
```diff
@@ -0,0 +1,707 @@
...
+    "$\\Delta \\lambda / \\lambda$ depends on $\\lambda$,\n",
+    "on the distance $z_{\\rm WFM}$ of the WFM chopper from the source,\n",
+    "and on the detector position $z_{\\rm det}$ as depicted in Fig. 4.\n",
+    "\n",
+    "The WFM chopper acts as a virtual source,\n",
+    "reducing the effective burst time $\\Delta t$,\n",
+    "while at the same time also reducing the effective time-of-flight to the detector from\n",
+    "$t(\\lambda) = \\alpha \\lambda z_{\\rm det}$ to $t(\\lambda) = \\alpha \\lambda (z_{\\rm det} - z_{\\rm WFM})$.\n",
```
OwenArnold, line 209: Assuming straight-line WFM-Detector via source. I think introducing the source into this is too complex, but the reader might wonder why we omit it.

docs/techniques/wfm/introduction-to-wfm.ipynb
```diff
@@ -0,0 +1,707 @@
...
+    "\n",
+    "A resolution that depends on $\\lambda$ is not suited for some applications (such as imaging, reflectometry, ...)\n",
+    "where a constant $\\Delta \\lambda / \\lambda$ is much more desirable.\n",
```
OwenArnold, line 215: Because, of limitations of the detector and timing systems, or some other physical features? Maybe Soren could add something to explain this.

docs/techniques/wfm/introduction-to-wfm.ipynb
```diff
@@ -0,0 +1,707 @@
...
+    "\n",
+    "### Using a pair of optically blind choppers\n",
+    "\n",
+    "A constant wavelength resolution $\\Delta \\lambda / \\lambda$ can be enforced by using a pair\n",
+    "of optically blind WFM chopper discs,\n",
+    "positioned at the positions $z_{\\rm WFM1}$ and $z_{\\rm WFM2} = z_{\\rm WFM1} + \\Delta z_{\\rm WFM}$,\n",
+    "as shown in Fig. 5.\n",
+    "\n",
+    "In this context, 'optically blind' indicates that the choppers have the same opening angles,\n",
```
OwenArnold, line 224: Is this true. Do they have to have the same opening angles to be optically blind. I seem to recall the V20 choppers not being duplicate choppers with a phase offset (in design)

docs/techniques/wfm/introduction-to-wfm.ipynb
```diff
@@ -0,0 +1,707 @@
...
+    "more than one frame can be realized by equipping the chopper with additional windows such that the full\n",
+    "band width that fits in the time period between two subsequent source pulses at the detector position is used.\n",
+    "This is illustrated in Fig. 6.\n",
+    "\n",
+    "The frame overlap chopper defines the bandwidth of the frame $N$ in order to suppress cross-talk\n",
+    "from neighboring wavelength frames,\n",
+    "as illustrated in Fig. 6 for two subsequent wavelength frames $N = 1$ and $N = 2$.\n",
+    "They are designed such that $\\lambda_{N, {\\rm max}} = \\lambda_{N+1, {\\rm min}}$ yield a continuous spectrum\n",
+    "at the detector.\n",
+    "The concept of wavelength frame multiplication implies that no data can be recorded in the\n",
```
OwenArnold, line 269: Fundamental requirement by design rather than something we imply. Do you want to mention that this is prior to stitching to distinguish the fact that once stitching occurs we do indeed expect overlap?

docs/techniques/wfm/introduction-to-wfm.ipynb
```diff
@@ -0,0 +1,707 @@
...
+   "source": [
+    "In a real WFM chopper system,\n",
+    "more than one frame can be realized by equipping the chopper with additional windows such that the full\n",
+    "band width that fits in the time period between two subsequent source pulses at the detector position is used.\n",
+    "This is illustrated in Fig. 6.\n",
+    "\n",
+    "The frame overlap chopper defines the bandwidth of the frame $N$ in order to suppress cross-talk\n",
+    "from neighboring wavelength frames,\n",
+    "as illustrated in Fig. 6 for two subsequent wavelength frames $N = 1$ and $N = 2$.\n",
+    "They are designed such that $\\lambda_{N, {\\rm max}} = \\lambda_{N+1, {\\rm min}}$ yield a continuous spectrum\n",
```
OwenArnold, line 267:
```suggestion
    "They are designed such that $\\lambda_{N, {\\rm max}} = \\lambda_{N+1, {\\rm min}}$ yield a continuous spectrum when measured in arrival time\n",
```

docs/techniques/wfm/introduction-to-wfm.ipynb
```diff
@@ -0,0 +1,707 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "2d8754af",
+   "metadata": {},
+   "source": [
+    "An additional feature of the WFM concept is the possibility to adjust the wavelength resolution by simply\n",
+    "changing the distance $\\Delta z_{\\rm WFM}$ between the WFM chopper discs (e.g. by using a motorized linear stage).\n",
+    "According to Eq. (2),\n",
+    "reducing $\\Delta z_{\\rm WFM}$ leads to a finer wavelength resolution at the cost of\n",
+    "intensity by effectively reducing the time window in which the neutrons can pass.\n",
```
OwenArnold, line 292: See comment above. Maybe this is sufficient if it's not easier to derive loss of intensity.

docs/techniques/wfm/introduction-to-wfm.ipynb
```diff
@@ -0,0 +1,707 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "23da3459",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "coords = wfm.make_fake_beamline(nframes=6)\n",
+    "ds = sc.Dataset(coords=coords)\n",
```
OwenArnold, line 313: Until now we haven't used scipp I think. Depending on audience, is it worth linking to other documentation for scipp?

docs/techniques/wfm/introduction-to-wfm.ipynb
```diff
@@ -0,0 +1,707 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "a94afdc5",
+   "metadata": {},
+   "source": [
+    "### Wrap the neutron counts and the beamline into a DataArray\n",
+    "\n",
+    "We make a data array that contains the beamlineinformation and a histogram of the neutrons over the time dimension."
```
OwenArnold:
```suggestion
    "We make a data array that contains the beamline information and a histogram of the neutrons over the time dimension."
```

docs/techniques/wfm/introduction-to-wfm.ipynb
```diff
@@ -0,0 +1,707 @@
...
+    "(1.5*sc.units.angstrom) * dlambda_over_lambda"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "ff3d05ba",
+   "metadata": {},
+   "source": [
+    "At 1.5 &#8491;, the resolution is 0.0127 &#8491;, which is smaller than the offset we observe above.\n",
+    "In fact, we can perform a quick check using Scipp's value-based slicing, to verify that the sum of the counts\n",
```
OwenArnold:
```suggestion
    "In fact, we can perform a quick check using Scipp's label-based slicing, to verify that the sum of the counts\n",
```

docs/techniques/wfm/introduction-to-wfm.ipynb
```diff
@@ -0,0 +1,707 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "a4720739",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "for i in range(len(wavelengths)):\n",
+    "    lam = wavelengths[\"wavelength\", i]\n",
+    "    dlam = 0.5 * dlambda_over_lambda * lam\n",
+    "    print(\"Lambda:\", lam, \", count in range:\", sc.sum(wav['wavelength', lam - dlam:lam + dlam]).value)"
```
OwenArnold: 👍  I think this is readable, but we could alternatively provide the `start`, `end` as variables above.

docs/techniques/wfm/introduction-to-wfm.ipynb
```diff
@@ -0,0 +1,707 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "3c25e1bf",
+   "metadata": {},
+   "source": [
+    "### Stitch the frames\n",
+    "\n",
+    "By using the information on the `frames`,\n",
```
OwenArnold: This seems a little brief.

Now that we have found the edges for each sub-frame we can apply a correction to shift it from a TOF time based on source t0 to a TOF based on t0n for each sub-pulse. Once our neutrons are adjusted to be in a consistent TOF frame, they can be directly merged between the sub-frames. At this point the WFM procedure is complete and further analysis (expect for filtering maybe) does not need to consider the WFM mode again.

docs/techniques/wfm/reducing-wfm-data.ipynb
```diff
@@ -0,0 +1,596 @@
...
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "218c876d",
+   "metadata": {},
+   "source": [
+    "# Reducing WFM data\n",
+    "\n",
+    "This notebook aims to illustrate how to work with the wavelength frame multiplication submodule `wfm`.\n",
+    "We will create a beamline that resembles the ODIN instrument beamline,\n",
```
OwenArnold:
```suggestion
    "We will create a beamline that resembles the ODIN instrument beamline, and\n",
```

docs/techniques/wfm/reducing-wfm-data.ipynb
```diff
@@ -0,0 +1,596 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "5b1b01a6",
+   "metadata": {},
+   "source": [
+    "## Generate some fake data\n",
+    "\n",
+    "Next, we will generate some fake data that is supposed to mimic a spectrum with a Bragg edge located at $4\\unicode{x212B}$.\n",
```
OwenArnold:
```suggestion
    "Next, we will generate some fake imaging data that is supposed to mimic a spectrum with a Bragg edge located at $4\\unicode{x212B}$.\n",
```

docs/techniques/wfm/reducing-wfm-data.ipynb
```diff
@@ -0,0 +1,596 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "8db72c1b",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def has_valid_path(wavelength, birth_time):\n",
```
OwenArnold: Do you still need this?

OwenArnold (changes requested):
I've focused on the text rather than the graphics. Between the new notebooks, I think this is really good.

I can see that the basics notebook would be great to onboard developers, along those lines, for the WFM basics I wonder if it's a good idea to include an appendix of some basic concepts/calculations/proofs such as the de Broglie wavelength. Might be overkill though.

docs/techniques/wfm/introduction-to-wfm.ipynb
nvaytet: I think it's more that some experiments just need more resolution to properly resolve the features (e.g. Bragg peaks/edges) they are looking at?

docs/techniques/wfm/introduction-to-wfm.ipynb
nvaytet: ~Yes, they have the same opening angles, also for v20.~
Actually you're right, they are different. I am suspecting a typo in the angles... I've asked Robin on Slack.

docs/techniques/wfm/introduction-to-wfm.ipynb
nvaytet: You are trading off flux, yes. I didn't want to go into too much detail on this because it also depends on the wavelengths of the neutrons, not just the opening times.

docs/techniques/wfm/introduction-to-wfm.ipynb
nvaytet: I mention the FOC just after, so I think there is no need to repeat it here?

docs/techniques/wfm/introduction-to-wfm.ipynb
OwenArnold: True, In the case of the bragg edge, it's just a factor of 2 t switch to d-spacing.

docs/techniques/wfm/introduction-to-wfm.ipynb
nvaytet: Did you mean `sample`?

docs/techniques/wfm/introduction-to-wfm.ipynb
nvaytet: We don't actually expect overlap, apart from the fact that the region Delta_t-wide on the right hand side of frame 1 will overlap with the region Delta_t-wide on the left hand side of frame 2, because by design they are the same wavelength.
This ensures no gaps between the frames.

I think the reason why you would see spikes at the overlap region might be slightly different (non-straight line) flight paths at a real beamline.

docs/techniques/wfm/introduction-to-wfm.ipynb
nvaytet: I think makes the point not quite as clear, and would rather leave it out.

docs/techniques/wfm/introduction-to-wfm.ipynb
nvaytet: This will be hosted on the scipp.github.io domain, so I think it's fine

docs/techniques/wfm/introduction-to-wfm.ipynb
nvaytet: oh, that's what we call it now?
i somehow missed that. I think i prefered the old name...

docs/techniques/wfm/introduction-to-wfm.ipynb
nvaytet: I made it longer, trying to make it more clear.
Please check

docs/techniques/wfm/reducing-wfm-data.ipynb
nvaytet: i think there is an `and` just after, so i would say no `and` here?

docs/techniques/wfm/reducing-wfm-data.ipynb
nvaytet: I addressed this in the other PR, which should be merged first.

docs/techniques/wfm/reducing-wfm-data.ipynb
nvaytet: No, it's been replaced by using a mask., which is a million times faster...
See #42

docs/techniques/wfm/introduction-to-wfm.ipynb
```diff
@@ -0,0 +1,741 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "id": "46d9de2e",
+   "metadata": {},
+   "source": [
+    "### Create some neutrons\n",
+    "\n",
+    "We create 6 neutrons with known wavelengths, one for each frame.\n",
+    "We choose the values for the wavelengths based on the limits given in the `frames` information above."
```
OwenArnold, line 344: You could do this programatically, why do this by stating the numbers again?

docs/techniques/wfm/introduction-to-wfm.ipynb
```diff
@@ -0,0 +1,741 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "b7d85ff4",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "for i in range(len(wavelengths)):\n",
+    "    lam = wavelengths[\"wavelength\", i]\n",
+    "    dlam = 0.5 * dlambda_over_lambda * lam\n",
+    "    print(\"Lambda:\", lam, \", count in range:\", sc.sum(wav['wavelength', lam - dlam:lam + dlam]).value)"
```
OwenArnold: dlam is might better be described as a half width assuming symmetry between + and - values. might want to choose a different name `half_dlam`?

Also there is a repetition of code here between a number of cells above for the stitching and the wavelength resolution tests.

OwenArnold (commented):
Overall, this is really good. I think this is actually the best description/demonstration of the technique I have found.

docs/techniques/wfm/introduction-to-wfm.ipynb
nvaytet: > dlam is might better be described as a half width assuming symmetry between + and - values. might want to choose a different name half_dlam?

ok, i'll do that

> Also there is a repetition of code here between a number of cells above for the stitching and the wavelength resolution tests.

Yes, I think the repetition is meant to illustrate the process I went through when figuring out what was going on. I think it is a small repetition that someone would be doing as part of a natural way of working with the notebook by simply re-executing the cells. So I think it's fine not to write functions for this.

docs/techniques/wfm/introduction-to-wfm.ipynb
nvaytet: I think because if i programmatically selected the dead center of each frame (which is probably what you would do), then I would go through the openings and I would not see the bad results for frame 1 and 2.

This was meant to illustrate the thought process I went through when writing this notebook, I could not understand why the wavelengths didn't match, until I tried to apply the DLambda/Lambda condition, and thought it was a nice way to understand the way WFM work and what it guarantees.

OwenArnold (approved):

PR #44, opened 2021-08-18
Fix wfm frames calcuation when source position is not (0,0,0)

nvaytet:
In the WFM time-distance diagram, the source position is assumed to be at zero.
So we need to take the difference between chopper position and source position everywhere.

### DIFF
--- a/src/ess/wfm/frames_analytical.py
+++ b/src/ess/wfm/frames_analytical.py
@@ -44,7 +44,7 @@ def frames_analytical(data: Union[sc.DataArray, sc.Dataset]) -> sc.Dataset:
     far_wfm_chopper = wfm_choppers[wfm_chopper_names[far_index]]
 
     # Compute distances for each detector pixel
-    detector_positions = data.meta["position"]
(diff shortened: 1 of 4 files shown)


src/ess/wfm/frames_analytical.py
```diff
@@ -82,7 +83,8 @@ def frames_analytical(data: Union[sc.DataArray, sc.Dataset]) -> sc.Dataset:
     # t_lambda_min is found from the relation between lambda_N and lambda_N+1,
     # equation (3) in Schmakat et al. (2020).
     t_lambda_min = t_lambda_max * z_ratio_wfm - data.meta["source_pulse_length"] * (
-        zdet_minus_zwfm / sc.norm(near_wfm_chopper.position))
+        zdet_minus_zwfm /
+        sc.norm(near_wfm_chopper.position - data.meta["source_position"]))
```
OwenArnold, line 87: This relative to source norm calc is repeated a lot. Could be extracted as a function.

OwenArnold (approved):
Fix makes sense.

PR #45, opened 2021-08-19
Add image on software stack

SimonHeybrock:
- Add image on software stack to title page, with links to related projects
- Fix theme

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -3,6 +3,11 @@ ess - Data reduction for ESS instrumentation
 
 `European Spallation Source`_ (ESS) toolkit for neutron scattering data reduction powered by `scipp`_ and `scippneutron`_.
 Provides ESS facility and instrument bespoke tools.
+The ``ess`` module is part of the software stack for data reduction:
+
+.. raw:: html
+
+   <object data="images/software-stack.svg" type="image/svg+xml"></object>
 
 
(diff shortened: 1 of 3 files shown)


docs/index.rst
```diff
@@ -3,7 +3,19 @@ ess - Data reduction for ESS instrumentation
 
 `European Spallation Source`_ (ESS) toolkit for neutron scattering data reduction powered by `scipp`_ and `scippneutron`_.
 Provides ESS facility and instrument bespoke tools.
+The ``ess`` module is part of the software stack for data reduction:
 
+.. raw:: html
+    :file: images/software-stack.svg
+
+- ``scipp``, ``scippneutron``, and ``ess`` are Python packages that can be installed using ``conda``.
+  Each package is released independently.
```
OwenArnold, line 12: We may get people wondering (in fact people do ask) why their latest changes for package Y don't work automatically against latest developments on X - Should we say in addition and is made to work against a target upstream package version - Or is this too much?

OwenArnold (approved):
Basically looks good. No need to hold this up.

docs/index.rst
SimonHeybrock: I think that is too complicated for the front page? Could be explained elsewhere (FAQ?), if required.

PR #48, opened 2021-09-10
Event mode wfm

nvaytet:
- Add a section in the WFM notebook on how to use the WFM stitching in event mode
- Also add a comparison to a reduction at a beamline without WFM choppers.

### DIFF
--- a/docs/techniques/wfm/reducing-wfm-data.ipynb
+++ b/docs/techniques/wfm/reducing-wfm-data.ipynb
@@ -430,7 +430,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "3561bab0",
+   "id": "5539ed9c",
    "metadata": {},
    "outputs": [],
    "source": [
@@ -440,7 +440,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "c68fce71",
+   "id": "b1a6d98e",
    "metadata": {},
    "outputs": [],
    "source": [
@@ -818,7 +818,7 @@
    "id": "b62325c3",
    "metadata": {},
    "source": [
-    "As written above, the `stitch` routine accepts both histogrammed and binned (event) data.\n",
+    "As explained above, the `stitch` routine accepts both histogrammed and binned (event) data.\n",
     "So stitching the binned data works in the exact same way as above, namely"
    ]
   },
@@ -837,7 +837,7 @@
   },
   {
    "cell_type": "markdown",
-   "id": "57c2cc6d",
+   "id": "63dd9a7c",
    "metadata": {},
    "source": [
     "The `stitch` function will return a data structure with a single bin in the `'tof'` dimension.\n",
@@ -848,7 +848,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "5e04c224",
+   "id": "d2088bd5",
    "metadata": {},
    "outputs": [],
    "source": [
--- a/docs/techniques/wfm/reducing-wfm-data.ipynb
+++ b/docs/techniques/wfm/reducing-wfm-data.ipynb
@@ -98,7 +98,7 @@
    "id": "e39e75dd",
    "metadata": {},
    "source": [
-    "We then proceed to generate two sets of 100,000 events:\n",
+    "We then proceed to generate two sets of 1,000,000 events:\n",
     "- one for the `sample` using the distribution defined above\n",
     "- and one for the `vanadium` which will be just a flat random distribution\n",
     "\n",
@@ -335,7 +335,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "9bb0a7e1",
+   "id": "95ae4c2b",
    "metadata": {},
    "outputs": [],
    "source": [
@@ -544,7 +544,7 @@
   },
   {
    "cell_type": "markdown",
-   "id": "99acb7de",
+   "id": "1bae47c9",
    "metadata": {},
    "source": [
     "We can see that the counts in the `stitched` data agree very well with the original data.\n",
@@ -562,7 +562,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "c4062491",
+   "id": "f53ed106",
    "metadata": {},
    "outputs": [],
    "source": [
@@ -596,7 +596,7 @@
   },
   {
    "cell_type": "markdown",
-   "id": "24a5346b",
+   "id": "2163950f",
    "metadata": {},
    "source": [
     "We then histogram these events to create a new Dataset.\n",
@@ -607,7 +607,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "412960b6",
+   "id": "3502d595",
    "metadata": {},
    "outputs": [],
    "source": [
@@ -638,7 +638,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "090ec557",
+   "id": "5aa17b52",
    "metadata": {},
    "outputs": [],
    "source": [
@@ -647,7 +647,7 @@
   },
   {
    "cell_type": "markdown",
-   "id": "af89eb54",
+   "id": "f214b67d",
    "metadata": {},
    "source": [
     "We then perform the standard unit conversion and normalization"
@@ -656,7 +656,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "df783215",
+   "id": "7baa7f73",
    "metadata": {},
    "outputs": [],
    "source": [
@@ -667,7 +667,7 @@
   },
   {
    "cell_type": "markdown",
-   "id": "a2e74378",
+   "id": "503846a5",
    "metadata": {},
    "source": [
     "In the same manner and in the previous section, we compare to the real neutron wavelengths"
@@ -676,7 +676,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "f756948c",
+   "id": "70d0efd0",
    "metadata": {},
    "outputs": [],
    "source": [
@@ -689,7 +689,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "96a57292",
+   "id": "b087e458",
    "metadata": {},
    "outputs": [],
    "source": [
@@ -707,7 +707,7 @@
   },
   {
    "cell_type": "markdown",
-   "id": "6effb32f",
+   "id": "573e400f",
    "metadata": {},
    "source": [
     "We can see that there is a significant shift between the calculated wavelength of the Bragg edge around $4\\unicode{x212B}$\n",
@@ -718,7 +718,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "59732126",
+   "id": "b0103324",
    "metadata": {},
    "outputs": [],
    "source": [
@@ -736,7 +736,11 @@
     "## Working in event mode\n",
     "\n",
     "It is also possible to work with WFM data in event mode.\n",
-    "The `stitch` utility will accept both histogrammed and binned (event) data. "
+    "The `stitch` utility will accept both histogrammed and binned (event) data.\n",
+    "\n",
+    "We first create a new dataset, with the same events as in the first example,\n",
+    "but this time we bin the data with `sc.bin` instead of using `sc.histogram`,\n",
+    "so we can retain the raw events."
    ]
   },
   {
@@ -773,13 +777,29 @@
    ]
   },
   {
+   "cell_type": "markdown",
+   "id": "5c02a0b1",
+   "metadata": {},
+   "source": [
+    "The underlying events can be inspected by using the `.events` property of our objects:"
+   ]
+  },
+  {
    "cell_type": "code",
    "execution_count": null,
-   "id": "8d648139",
+   "id": "b4a94774",
    "metadata": {},
    "outputs": [],
    "source": [
     "ds_event[\"sample\"].events"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "a41e2625",
+   "metadata": {},
+   "source": [
+    "We can visualize this to make sure it looks the same as the histogrammed case above:"
    ]
   },
   {
@@ -793,17 +813,34 @@
    ]
   },
   {
+   "cell_type": "markdown",
+   "id": "2b98d882",
+   "metadata": {},
+   "source": [
+    "As written above, the `stitch` routine accepts both histogrammed and binned (event) data.\n",
+    "So stitching the binned data works in the exact same way as above, namely"
+   ]
+  },
+  {
    "cell_type": "code",
    "execution_count": null,
-   "id": "869a8d95",
+   "id": "b05a0d2e",
    "metadata": {},
    "outputs": [],
    "source": [
     "stitched_event = wfm.stitch(frames=frames,\n",
-    "                      data=ds_event,\n",
-    "                      dim='time',\n",
-    "                      merge_frames=True)\n",
+    "                            data=ds_event,\n",
+    "                            dim='time',\n",
+    "                            merge_frames=True)\n",
     "stitched_event"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "eb64ed68",
+   "metadata": {},
+   "source": [
+    "At this point, it may be useful to compare the results of the two different stitching operations."
    ]
   },
   {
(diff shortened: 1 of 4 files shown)


src/ess/wfm/stitch.py
```diff
@@ -4,8 +4,27 @@
...
+
+def _update_source_position(item, frames):
+    coords_or_attrs = None
+    for meta in ["coords", "attrs"]:
+        if "source_position" in getattr(item, meta):
+            coords_or_attrs = meta
+    if coords_or_attrs is None:
+        raise KeyError("'source_position' was not found in metadata.")
+    getattr(item,
+            coords_or_attrs)['source_position'] = frames["wfm_chopper_mid_point"].data
```
SimonHeybrock: Is the reason for removing the source position that `scn.convert` would otherwise use it to compute `L1`?

src/ess/wfm/stitch.py
```diff
@@ -32,26 +51,18 @@ def _stitch_item(item: sc.DataArray, dim: str, frames: sc.Dataset, merge_frames:
...
-        out = sc.DataArray(data=sc.zeros(dims=dims,
-                                         shape=shape,
-                                         variances=item.variances is not None,
-                                         unit=item.unit),
-                           coords={"tof": tof_coord})
-        for group in ["coords", "attrs"]:
-            for key in getattr(item, group):
-                if key != dim:
-                    getattr(out, group)[key] = getattr(item, group)[key].copy()
+        if item.bins is None:
```
SimonHeybrock: Isn't this checked by the top-level function?

src/ess/wfm/stitch.py
```diff
@@ -60,27 +71,79 @@ def _stitch_item(item: sc.DataArray, dim: str, frames: sc.Dataset, merge_frames:
         section.coords['tof'] = section.meta[dim] - frames["time_correction"].data[
             "frame", i]
         del section.meta[dim]
-        # TODO: when scipp 0.8 is released, rename_dims will create a new object.
-        # section = section.rename_dims({dim: 'tof'})
         section.rename_dims({dim: 'tof'})
 
         if merge_frames:
             out += sc.rebin(section, 'tof', out.meta["tof"])
```
SimonHeybrock: Is the `rebin` actually necessary to be included here? Could we just copy into the correct slice of `out` (and `rebin` the whole thing later on, if nicely aligned bins are desired)? Or is the issue that there may be different numbers of bins for different pixels? Or do the frames overlap after applying shifts (and that is why the `+=` is needed, rather than an assignment?)?

src/ess/wfm/stitch.py
```diff
@@ -60,27 +71,79 @@ def _stitch_item(item: sc.DataArray, dim: str, frames: sc.Dataset, merge_frames:
         section.coords['tof'] = section.meta[dim] - frames["time_correction"].data[
             "frame", i]
```
SimonHeybrock: Related to my question about the use of `rebin`: Have you ever considered working with bin centers rather than histograms for the entire stitching procedure, and use `histogram` as the final step? Basically something like:

```python
item.coords['tof'] += time_offset  # also update pulse_time?
stitched = sc.histogram(item, bins=tof_coord)  # assume regions between frames have been offset to be out of range
```

src/ess/wfm/stitch.py
```diff
@@ -60,27 +71,79 @@ def _stitch_item(item: sc.DataArray, dim: str, frames: sc.Dataset, merge_frames:
...
+                           sc.concatenate(frames["time_min"].data["frame", i],
+                                          frames["time_max"].data["frame", i], 'time')
+                       ])
+        piece.bins.coords[
+            'tof'] = piece.bins.meta[dim] - frames["time_correction"].data["frame", i]
+        piece.coords['tof'] = piece.meta['time'] - frames["time_correction"].data[
+            "frame", i]
+        del piece.meta['time']
+        del piece.bins.meta['time']
+        section = piece.events.copy()
```
SimonHeybrock: Why the copy?

src/ess/wfm/stitch.py
```diff
@@ -60,27 +71,79 @@ def _stitch_item(item: sc.DataArray, dim: str, frames: sc.Dataset, merge_frames:
...
+                       edges=[
+                           sc.concatenate(frames["time_min"].data["frame", i],
+                                          frames["time_max"].data["frame", i], 'time')
+                       ])
+        piece.bins.coords[
+            'tof'] = piece.bins.meta[dim] - frames["time_correction"].data["frame", i]
+        piece.coords['tof'] = piece.meta['time'] - frames["time_correction"].data[
+            "frame", i]
+        del piece.meta['time']
+        del piece.bins.meta['time']
```
SimonHeybrock: Can this be done in-place (modifying `time`), such that we avoid the large extra memory allocation for `tof`?

src/ess/wfm/stitch.py
```diff
@@ -60,27 +71,79 @@ def _stitch_item(item: sc.DataArray, dim: str, frames: sc.Dataset, merge_frames:
...
+        section = piece.events.copy()
+        section.rename_dims({'time': 'tof'})
+
+        if merge_frames:
+            if out is None:
+                out = section
+            else:
+                out = sc.concatenate(out, section, 'tof')
```
SimonHeybrock: This is expensive. Can we modify the events in-place instead?

- Instead of extracting individual frames using `bin` (in line 100), do not use a `for` loop, but just use `bin` to cut into frames (but keep everything in the same data array).
- Use in-place coord modification to apply offsets.
- Use (a single) `bin` to map into new `tof` bins and drop events between frames.

SimonHeybrock (commented):
Didn't look at the notebook yet, just the code. Some questions and potential ideas for improvement:

src/ess/wfm/stitch.py
nvaytet: Yes, that is the reason. I guess you are suggesting to just directly set `L1` instead?

SimonHeybrock: Yes.

docs/techniques/wfm/reducing-wfm-data.ipynb
```diff
@@ -98,11 +98,11 @@
...
    "metadata": {},
    "source": [
-    "We then proceed to generate two sets of 100,000 events:\n",
+    "We then proceed to generate two sets of 1,000,000 events:\n",
     "- one for the `sample` using the distribution defined above\n",
-    "- and one for the `background` which will be just a flat random distribution\n",
+    "- and one for the `vanadium` which will be just a flat random distribution\n",
     "\n",
-    "For the events in both `sample` and `background`,\n",
+    "For the events in both `sample` and `vanadium`,\n",
```
OwenArnold, line 105: In the view/plot step below. Are we sufficiently sure that users would have encountered the binned data and histogram concepts before? If not you could either:
* point them at the corresponding introductions in `scipp` in the introduction markdown
* briefly introduce the concepts.

docs/techniques/wfm/reducing-wfm-data.ipynb
```diff
@@ -114,7 +114,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "nevents = 100_000\n",
+    "nevents = 1_000_000\n",
```
OwenArnold, line 117: Not directly related to these changes, but you refer to `slopes` in some of the calculations. These are just `velocities` on a time-distance diagram. I think these would be better terms.

docs/techniques/wfm/reducing-wfm-data.ipynb
```diff
@@ -532,21 +536,397 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "3502d595",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "tmin = min([item[\"arrival_times\"].values.min() for item in events_no_wfm.values()])\n",
+    "tmax = max([item[\"arrival_times\"].values.max() for item in events_no_wfm.values()])\n",
+    "\n",
+    "dt = 0.1 * (tmax - tmin)\n",
```
OwenArnold, line 618: What is 0.1 in this context? How did you select it?

docs/techniques/wfm/reducing-wfm-data.ipynb
```diff
@@ -532,21 +536,397 @@
...
+    "time_coord_no_wfm = sc.linspace(dim='tof',\n",
+    "                         start=tmin - dt,\n",
+    "                         stop=tmax + dt,\n",
+    "                         num=257,\n",
```
OwenArnold, line 622: We have 257 bins, where did this come from?

docs/techniques/wfm/reducing-wfm-data.ipynb
```diff
@@ -532,21 +536,397 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "for item in events.values():\n",
+    "    item[\"valid_times\"] = item[\"arrival_times\"].values[item[\"valid_indices\"]]\n",
+    "\n",
+    "tmin = min([item[\"valid_times\"].min() for item in events.values()])\n",
+    "tmax = max([item[\"valid_times\"].max() for item in events.values()])\n",
+    "\n",
+    "dt = 0.1 * (tmax - tmin)\n",
```
OwenArnold, line 760: Same question as above

docs/techniques/wfm/reducing-wfm-data.ipynb
```diff
@@ -532,21 +536,397 @@
...
+   "source": [
+    "sc.plot(ds_event.bins.sum())"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "2b98d882",
+   "metadata": {},
+   "source": [
+    "As written above, the `stitch` routine accepts both histogrammed and binned (event) data.\n",
```
OwenArnold:
```suggestion
    "As explained above, the `stitch` routine accepts both histogrammed and binned (event) data.\n",
```

src/ess/wfm/stitch.py
```diff
@@ -32,26 +51,18 @@ def _stitch_item(item: sc.DataArray, dim: str, frames: sc.Dataset, merge_frames:
...
-            for key in getattr(item, group):
-                if key != dim:
-                    getattr(out, group)[key] = getattr(item, group)[key].copy()
+        if item.bins is None:
+            out = sc.DataArray(data=sc.zeros(dims=dims,
+                                             shape=shape,
+                                             variances=item.variances is not None,
+                                             unit=item.unit),
+                               coords={"tof": tof_coord})
+            _populate_coords(item, out, dim)
```
OwenArnold: keyword args rather than positional would be useful esp given in-place operation.

src/ess/wfm/stitch.py
```diff
@@ -60,27 +71,79 @@ def _stitch_item(item: sc.DataArray, dim: str, frames: sc.Dataset, merge_frames:
...
+
+        if merge_frames:
+            if out is None:
+                out = section
+            else:
+                out = sc.concatenate(out, section, 'tof')
+        else:
+            _populate_coords(item, section, dim)
+            _update_source_position(section, frames)
+            out[f"frame{i}"] = section
```
OwenArnold, line 122: just return here, and then you can avoid subsequent `if merge_frames` condition below.

OwenArnold (changes requested):
Overall looks good. Most of my comments are minor.

src/ess/wfm/stitch.py
```diff
@@ -60,27 +71,79 @@ def _stitch_item(item: sc.DataArray, dim: str, frames: sc.Dataset, merge_frames:
...
     if merge_frames:
-        getattr(
-            out,
-            coords_or_attrs)['source_position'] = frames["wfm_chopper_mid_point"].data
+        _update_source_position(out, frames)
 
     return out
 
 
+def _stitch_event_data(item: sc.DataArray, dim: str, frames: sc.Dataset,
```
OwenArnold: There seem to be no corresponding unit tests to cover the event stitching cases.

src/ess/wfm/stitch.py
nvaytet: @SimonHeybrock  How would you handle, in a nice way, the cases where you need to set `L1` (because you will use `scatter=True` in the conversion later), and when you need to set `Ltotal` (because you need `scatter=False` for imaging)?
Any good ideas?

src/ess/wfm/stitch.py
nvaytet: yes, rebin is need because there is a small amount of overlap.

docs/techniques/wfm/reducing-wfm-data.ipynb
nvaytet: Just to add a bit of padding. It's already used further up the notebook.

docs/techniques/wfm/reducing-wfm-data.ipynb
nvaytet: The same number as further up the notebook, when WFM choppers are in use. No particular reason for tha number and notebook should work if you modify it.

src/ess/wfm/stitch.py
nvaytet: yes, no unit tests for event data yet. Need to add them.

src/ess/wfm/stitch.py
SimonHeybrock: Not really. I think it might be better to use a special WFM unit conversion graph (instead of adding coords to hit the correct branch inn the standard graph), which would not look for `source_position` but something like `wfm_chopper_position`.

src/ess/wfm/stitch.py
```diff
@@ -4,94 +4,130 @@
...
-            coords_or_attrs)['source_position'] = frames["wfm_chopper_mid_point"].data
+    _update_source_position(out, frames)
 
     return out
 
 
+def _stitch_event_data(item: sc.DataArray, dim: str, frames: sc.Dataset, new_dim: str,
+                       bins: Union[int, sc.Variable]) -> Union[sc.DataArray, dict]:
+
+    # Note: it's better to have frame as the inner dim and pixels as the outer dims
```
SimonHeybrock: This comment seems unrelated to the line after?

src/ess/wfm/stitch.py
```diff
@@ -4,94 +4,130 @@
...
+                       to=dim)
+
+    binned = sc.bin(item, edges=[edges])
+
+    for i in range(frames.sizes["frame"]):
+        binned[dim, i * 2].bins.coords[dim] -= frames["time_correction"].data["frame",
+                                                                              i]
+
+    binned.bins.coords[new_dim] = binned.bins.coords[dim]
+    del binned.bins.coords[dim]
```
SimonHeybrock: This would go wrong if `new_dim == dim`, which does not seem too unlikely, I suppose if we just load raw data using, say, `load_nexus` we get `dim='tof'`, which is also what we want as a target dim.

src/ess/wfm/stitch.py
```diff
@@ -4,94 +4,130 @@
...
+
+    binned.masks['frame_gaps'] = (sc.arange(dim, 2 * frames.sizes["frame"] - 1) %
+                                  2).astype(sc.dtype.bool)
+
+    new_edges = sc.concatenate(
+        (frames["time_min"]["frame", 0] - frames["time_correction"]["frame", 0]).data,
+        (frames["time_max"]["frame", -1] - frames["time_correction"]["frame", -1]).data,
+        new_dim)
+    return sc.bin(binned, edges=[new_edges], erase=[dim])
```
SimonHeybrock: Similar to above comment we should consider support `new_dim == dim`?

src/ess/wfm/stitch.py
nvaytet: Ok, so for now I will just keep the modification of the `source_position` and change it once we have the transform coord graphs in the unit conversion.

docs/techniques/wfm/reducing-wfm-data.ipynb
nvaytet: I'm not really sure where to point them to? It could be a large portion of the scipp docs pages.

docs/techniques/wfm/reducing-wfm-data.ipynb
nvaytet: I had `slopes` because they go along `intercepts`.
I think `velocities` and `intercepts` is a little weird?

docs/techniques/wfm/reducing-wfm-data.ipynb
nvaytet: done

docs/techniques/wfm/reducing-wfm-data.ipynb
OwenArnold: I think your intercepts correspond to distances at which the paths would cross. So the term intercept, as in intercept point would still work. Nevertheless this is too minor to worry much about, it can go in as is.

tests/wfm/stitch_test.py
```diff
@@ -93,42 +109,62 @@ def _do_stitching_on_beamline(wavelengths):
...
-        _check_lambda_inside_resolution(wavelengths['wavelength', i],
-                                        dlambda_over_lambda, rebinned)
+        _check_lambda_inside_resolution(wavelengths['event', i],
+                                        dlambda_over_lambda,
+                                        stitched,
+                                        event_mode=event_mode)
 
 
-def test_stitching_on_beamline_bad_wavelength():
+@pytest.mark.parametrize("event_mode", [False, True])
```
OwenArnold, line 151: 👍

SimonHeybrock (approved):

OwenArnold (approved):

PR #49, opened 2021-09-16
With scipp 0.8

nvaytet:
- Had to remove some comparisons in the relfectometry data tests because when using grouping, all events in the same bin have the same ID, and they are no longer stored redundantly
- Some small fixes to function arguments, e.g. `variances -> with_variances`

**Note:** builds will fail until `scippneutron=0.3` is released.


arm61 (approved):
I am happy with the changes to the reflectometry code

src/ess/wfm/stitch.py
```diff
@@ -67,8 +65,10 @@ def _stitch_event_data(item: sc.DataArray, frames: sc.Dataset, dim: str, new_dim
     binned = sc.bin(item, edges=[edges])
 
     for i in range(frames.sizes["frame"]):
-        binned[dim, i * 2].bins.coords[dim] -= frames["time_correction"].data["frame",
-                                                                              i]
+        # TODO: temporary fix working on the .values because read-only flag is set
+        binned[dim, i *
+               2].value.coords[dim].values -= frames["time_correction"].data["frame",
+                                                                             i].values
```
SimonHeybrock: Do you actually need the `.values`? The crucial bit of the workaround is the first `.value` (singular).

PR #50, opened 2021-09-24
transform_coords building blocks for gravity correction in theta vs wavelength transformations

OwenArnold:
Aim is to put generic building blocks in place rather than wire it in. `transform_coords` became available in scipp 0.8 and is the new generic mechanism.

Effectively does this 
image

See non `transform_coords` implementation https://scipp.github.io/ess-notebooks/reflectometry/amor_reduction_detailed.html#Theta-calculation-and-gravity-correction 

This implementation produces identical for input data that @arm61  uses:
image


src/ess/reflectometry/transform_coords.py
```diff
@@ -0,0 +1,28 @@
+import scipp as sc
+from scipp.constants import neutron_mass, h, g
+
+
+def to_velocity(wavelength):
+    return sc.to_unit(h / (wavelength * neutron_mass), sc.units.m / sc.units.s,
+                      copy=False)
```
SimonHeybrock: Rewrite this to do large operations last. As written here we first create multiple large intermediates, consuming a lot of memory and compute. Rewrite to multiply with `wavelength` as the very last step. Also `to_unit` could be done before.

src/ess/reflectometry/transform_coords.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+# Derivative of y with respect to z
+def to_y_dash(wavelength, sample_position, detector_position):
```
SimonHeybrock: Could this use `scattered_beam` instead of the two positions?

src/ess/reflectometry/transform_coords.py
```diff
@@ -0,0 +1,28 @@
...
+    diff = (detector_position - sample_position)
+    velocity_sq = to_velocity(wavelength)
+    velocity_sq *= velocity_sq
+    gy = sc.vector(value=[0, -1, 0]) * g
+    # dy due to gravity = -0.5gt^2 = -0.5g(dz/dv)^2
+    # therefore y'(z) = dy/dz - 0.5g.dz/dv^2 / dz
+    return (-0.5 * sc.norm(gy)
+            * diff.fields.z / velocity_sq) + (diff.fields.y / diff.fields.z)
```
SimonHeybrock: This is bound to break since it hard-codes the direction of gravity. I would recommend to require a `g` coordinate that is defined as a *vector* with every data array. Essentially this vector would replace (parts of) what Mantid defines as "reference frame".

src/ess/reflectometry/transform_coords.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+def to_scattering_angle(w, wavelength, detector_id, position, sample_position):
+    z_origin = sample_position.fields.z
+    y_origin = sample_position.fields.y
+    z_measured = position.fields.z
+    y_dash = to_y_dash(wavelength, sample_position, position)
+    height = y_dash * (z_measured - z_origin) + y_origin
+    return sc.atan2(y=height, x=z_measured) - w
```
SimonHeybrock: Hard coding some directions as well? Can we define/require, e.g., a sample-normal vector instead?

SimonHeybrock (commented):
Some comments below. Should we drop the `to_` prefixes, we don't use prefixes much elsewhere?

src/ess/reflectometry/transform_coords.py
OwenArnold: Yes, fully agree. Would be much happier if we did this with configurable beam and sample normal (up) directions.

SimonHeybrock: Would not call it `up` though, since there are reflectometry beamlines where "up" is not the sample normal, since the beamline "lies" on its side.

src/ess/reflectometry/transform_coords.py
```diff
@@ -8,22 +9,59 @@ def to_velocity(wavelength):
...
+def L1(incident_beam):
+    return sc.norm(incident_beam)
+
+
+def L2(scattered_beam):
+    return sc.norm(scattered_beam)
+
+
+def _angle(a, b):
+    return sc.acos(sc.dot(a, b) / (sc.norm(a) * sc.norm(b)))
```
jl-wynen, line 41: Those functions will be provided by scippneutron. I think we should prefer finding a good API there and using that here. ess should only provide specialised building blocks that can be added to or replace components of graphs in scippneutron.

nvaytet:
The use of `transform_coords` has been implemented in #61 , but I feel like the unit tests here should still make it into the codebase.

SimonHeybrock:
> The use of `transform_coords` has been implemented in #61 , but I feel like the unit tests here should still make it into the codebase.

Yes, but I think we should also avoid testing `transform_coords`, and rather focus on testing the graph building blocks directly. Otherwise tests get unnecessarily complex.

jl-wynen:
This is provided by ScippNeutron now.

PR #57, opened 2021-10-07
Diffraction event filtering functions

Pasarus:
This is the culmination of work on producing functions that are similar enough to the mantid algorithms that scientists are used to using for Neutron data.

The conversion/functionality is comparable for these:
FilterBadPulses -> filter_bad_pulses
FilterByLogValue -> filter_data_array_on_attribute
FilterByTime -> filter_by_time
FilterByXValue -> (This can be done with normal slicing, bin, groupby etc already)
FilterEvents -> (functions very similar to bin or groupby)
FilterLogByTime -> filter_attribute_by_time
(You could filter the sample log by value easily in mantid, not so easy in scipp so I made a helper function) -> filter_attribute_by_value


tests/diffraction/filter_test.py
```diff
@@ -0,0 +1,264 @@
...
+    np.testing.assert_almost_equal(good_da_sum, good_sum)
+    np.testing.assert_almost_equal(bad_da_sum, bad_sum)
+
+
+def test_results_of_bad_pulse_defaults():
+    da = scn.data.tutorial_event_data()
+
+    filtered_da = filter_bad_pulses(da)
+
+    _test_filter_sums_for_bad_pulses(filtered_da, 51799.703125, 6282525.0)
```
jl-wynen: Where do those numbers come from?

Maybe it would be better to test smaller data arrays where you can compute the expected results by hand. This would also speed up the tests because we would not load large data files all the time.

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,237 @@
...
+import random
+
+import numpy as np
+
+
+def filter_bad_pulses(da: sc.DataArray,
+                      centre: bool = False,
+                      minimum_threshold: float = .95):
+    """
+    Filters out any bad pulses based on an attribute named "proton_charge" if the
```
jl-wynen, line 14: Please add and argument to allow users to specify the name of the proton charge attribute. And the same would probably make sense for `"pulse_time"`. They can default to the values used here if those are expected to be used by ESS but we should not force users to use a specific naming scheme.

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,237 @@
...
+
+import scipp as sc
+import string
+import random
+
+import numpy as np
+
+
+def filter_bad_pulses(da: sc.DataArray,
+                      centre: bool = False,
```
jl-wynen: Please change to `center`, we use American spelling in scipp.

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,237 @@
...
+    Args:
+        da: sc.DataArray, the data array that is to be filtered.
+
+        centre: bool, whether or not to centre the data of the "proton_charge" as if it
+         were binned, this is done with SNS data.
+
+        minimum_threshold: float, the minimum threshold is a percentage used calculate
+         at what point is determined as a bad proton charge, this value is multiplied by
+         the mean value of the proton_charge data, and that is the minimum by which
+         filtering will occur.
```
jl-wynen: Is there a typo in the 2nd line? I cannot quite understand what you mean. Also, an equation would be nice, regardless of how simple it is.

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,237 @@
...
+    proton_charge = da.attrs['proton_charge'].value
+    proton_charge = proton_charge.rename_dims({'time': 'pulse_time'})
+    proton_charge.coords['pulse_time'] = proton_charge.coords.pop('time')
+
+    if centre:
+        proton_charge.data.values[:-1] = 0.5 * (proton_charge.data.values[1:] +
+                                                proton_charge.data.values[:-1])
+
+    max_charge = sc.max(proton_charge.data)
+    max_charge *= 1.1  # determine true max (max_value * 1.1)
```
jl-wynen: What does 'true' mean here? I don't understand why you multiply the max by 1.1

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,237 @@
...
+    """
+    good_data_coord = good_data_lut.coords[mapping_coord]
+    edge = good_data_lut.data
+    edge = edge[mapping_coord, :-1] ^ edge[mapping_coord, 1:]
+    good_edges = sc.Dataset(data={
+        mapping_coord: good_data_coord[mapping_coord, 1:],
+        'good': good_data_lut.data[mapping_coord, :-1]
+    },
+                            coords={'edge': edge})
+    good_edges = good_edges.groupby(group='edge').copy(1)
```
jl-wynen: Please use a bool as argument to `copy`.

jl-wynen (commented):
I have a conceptional question since I don't know how this is supposed to be used. Would it make sense to add a mask that is `True` for bin 0 and `False` for bin 1 to explicitly 'filter' the bad data? Or are users expected to always split the two bins into separate arrays?

src/ess/diffraction/filter.py
Pasarus: These were written with PG3 data in mind, and as defaults, my thoughts was this would be updated as the ESS defaults were determined.

tests/diffraction/filter_test.py
Pasarus: this is a case of I didn't comment "magic numbers", truthfully I was unsure what to use for testing this function, the data looks right and inputs will just be adjusted by users as needed to make it "look" right, so besides the fact that's the output and these wills stop regression these numbers mean very little.

src/ess/diffraction/filter.py
Pasarus: Believe it or not auto-correct did this, I will correct, good spot.

src/ess/diffraction/filter.py
Pasarus: Left over bad comments, thanks for spotting this I will remove them as the code speaks for itself. The max being used is explained the in the function comment.

"if a proton charge is 10% higher
     than the maximum (This won't actually make any difference in comparison to just the
     max value) based on how Mantid did filtering of bad pulses."

src/ess/diffraction/filter.py
Pasarus: Can do!

src/ess/diffraction/filter.py
Pasarus: Added your suggested arguments and an argument for defining the time dimension's name in the attribute.

src/ess/diffraction/filter.py
Pasarus: Updated and improved, thanks!

Pasarus:
For FilterBadPulses/filter_bad_pulses the data is either useful or it isn't, so my assumption is that users will just want rid of the "bad" data, I then applied this logic to each of the other functions. Hence the use of slicing instead of a mask, it allows you to just have the useful data and dump "bad" data via the gc. In large datasets this will save not small amounts of memory if PG3 data is anything to go by.

src/ess/diffraction/filter.py
jl-wynen: Ok, so the reason is 'because Mantid does it'? I am unsure how much we should care about that

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,251 @@
...
+    edges = sc.concatenate(start, end, time_dim_name)
+    return sc.bin(da, edges=[edges])
+
+
+def filter_attribute_by_time(da: sc.DataArray,
+                             attribute_name: str,
+                             start,
+                             end,
+                             time_name: str = "time",
+                             datetime_unit: str = "ns"):
```
SimonHeybrock: I don't think this is an interface that is in-line with scipp standards. We should only support variables (since units are builtin). I am note sure this entire function is a good idea? I looks like all it provides is a convenience to convert ranges to the correct unit, but in turn requires memorizing a new function name and function signature.

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,251 @@
...
+         not already constructed start or end i.e. a string instead of a Scalar variable
+         containing datetime64.
+
+    Returns:
+        The output of the sc.bin after filtering out everything between the start and
+         end args from the input da
+    """
+    start, end = _input_processing_time_conversion(start, end, datetime_unit)
+    edges = sc.concatenate(start, end, time_dim_name)
+    return sc.bin(da, edges=[edges])
```
SimonHeybrock: Why do we need this long and complicated function (mainly the docstring)? Apart from violating the scipp approach of always using quantities (values with unit, i.e., scalar variables) as function parameters, it is also restrictive for now reason. Why do we need a special function just for filtering datetime? Can we not do this filtering explicitly? If there is something inconvenient about have start and end are specified we may consider improving the generic interfaces, or adding a generic helper, but as it is I feel all it does is replicating the unfortunate Mantid situation where we have 1000 different algorithms, of which many do the same thing.

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,251 @@
...
+import string
+import random
+
+import numpy as np
+
+
+def filter_bad_pulses(da: sc.DataArray,
+                      center: bool = False,
+                      minimum_threshold: float = .95,
+                      proton_charge_attribute_name: str = "proton_charge",
```
SimonHeybrock: Just use a data array as input here? Then the user can specify any time-series log, not just those that are stored as attributes.

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,251 @@
...
+                      proton_charge_time_dim_name: str = "time",
```
SimonHeybrock: Is this required? The proton charge log has to be 1D anyway, so dimension should always be automatic?

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,251 @@
...
+        .rename_dims({proton_charge_time_dim_name: data_time_coord})
+    proton_charge.coords[data_time_coord] = \
+        proton_charge.coords.pop(proton_charge_time_dim_name)
+
+    if center:
+        proton_charge.data.values[:-1] = 0.5 * (proton_charge.data.values[1:] +
+                                                proton_charge.data.values[:-1])
+
+    max_charge = sc.max(proton_charge.data)
+    max_charge *= 1.1
```
SimonHeybrock: Hard coded?

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,251 @@
...
+    min_charge = sc.mean(proton_charge.data)
+    min_charge *= minimum_threshold
+
+    good_pulse = (proton_charge >= min_charge) & (proton_charge < max_charge)
```
SimonHeybrock: We should probably make everything until here available as a separate function?

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,251 @@
...
+    return attribute_data[time_name, start:end]
+
+
+def _generate_random_string(length: int = 20):
+    charecters = string.ascii_letters + string.digits + string.punctuation
+    return "".join(random.choice(charecters) for ii in range(length))
+
+
+def filter_attribute_by_value(da: sc.DataArray, attribute_name: str,
+                              min_value: sc.Variable, max_value: sc.Variable):
```
SimonHeybrock: Why is this used for in practice?

Shouldn't this be made more generic to simply do this for a data array, rather than the value of an attribute of a data array? That is, is there any need for the restriction and the ensuring complicated function signature?

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,251 @@
...
+    """
+    # Define what data is in range
+    attribute = da.attrs[attribute_name].value
+    greater_than = attribute > min_value
+    less_than = attribute < max_value
+    in_range = greater_than == less_than
+
+    # Group up data based on the in range data that is present and clean up groupby
+    # coordinate
+    filter_coord_name = _generate_random_string()
```
SimonHeybrock: You are hoping to avoid name clashes like this? Have you considered using something like `'_'.join(da.coords)`?

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,251 @@
...
+    attribute.coords[filter_coord_name] = in_range.data
```
SimonHeybrock: I don't think we should add attributes to an existing data array. Make a shallow copy beforing doing this? (`da.copy(deep=False)`)

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,251 @@
...
+    data_in_range = attribute.groupby(filter_coord_name).copy(True)
+    del data_in_range.coords[filter_coord_name]
+    del attribute.coords[filter_coord_name]
+
+    return data_in_range
+
+
+def filter_data_array_on_attribute(da: sc.DataArray,
```
SimonHeybrock: I don't think the `data_array` in the function names adds anything?

This function does not actually seem to filter by an attribute, rather it filters/extract intervals?

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,251 @@
...
+    """
+    good_data_coord = good_data_lut.coords[mapping_coord]
+    edge = good_data_lut.data
+    edge = edge[mapping_coord, :-1] ^ edge[mapping_coord, 1:]
+    good_edges = sc.Dataset(data={
+        mapping_coord: good_data_coord[mapping_coord, 1:],
+        'good': good_data_lut.data[mapping_coord, :-1]
+    },
+                            coords={'edge': edge})
+    good_edges = good_edges.groupby(group='edge').copy(True)
```
SimonHeybrock: Move edge finding to separate function?

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,251 @@
...
+    good_data = sc.DataArray(data=good_edges['good'].data
+                             ^ good_data_lut[mapping_coord, 0].data,
+                             coords={mapping_coord: good_edges[mapping_coord].data})
+    good_data = sc.concatenate(good_data_lut[mapping_coord, 0], good_data,
+                               mapping_coord)
+    good_data = sc.concatenate(good_data, good_data_lut[mapping_coord, -1],
+                               mapping_coord)
+    good_pulse_lut = sc.lookup(good_data, mapping_coord)
+
+    da.bins.coords[data_slices_name] = good_pulse_lut[da.bins.coords[mapping_coord]]
```
SimonHeybrock: Adding coord to input... maybe not ideal?

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,251 @@
...
+
+        data_slices_name: str, The name of the dimension that will be added to the
+         passed sc.DataArray, da, that can then be sliced later for good and bad data.
+
+    Returns:
+        The originally passed sc.DataArray, da, with an extra dimension that can then be
+         sliced for good data on slice 1, and sliced for bad data on slice 0. For
+         example, da["slice_data", 1] for good data.
+    """
+    good_data_coord = good_data_lut.coords[mapping_coord]
```
SimonHeybrock: Is there a reason for supporting anything other than the dimension-coordinate, i.e., the one where the name matches the dim?

SimonHeybrock (changes requested):

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,251 @@
...
+                      proton_charge_time_dim_name: str = "time",
+                      data_time_coord: str = "pulse_time"):
+    """
+    Filters out any bad pulses based on an attribute named "proton_charge" if the
+     "proton_charge" value is less than 95% of the mean (by default, the percentage can
+     be changed using "minimum_threshold") it will be filtered out, if a proton charge
+     is 10% higher than the maximum (This won't actually make any difference in
+     comparison to just the max value) based on how Mantid did filtering of bad pulses.
+
+    Args:
```
nvaytet: You may have copied the docstring style from the reflectometry code that is already in the `ess` repo, but I have actually asked Drew to change this to the Sphinx docstring format (https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html#the-sphinx-docstring-format), so that it's the same as in `scipp` and `scippneutron`.
The google style that you are using causes issues with the formatting/indentation of long lines in sphinx.

Sorry, we should really write this in our style guidelines.

nvaytet (changes requested):

src/ess/diffraction/filter.py
Pasarus: I just used the default that pycharm gave me here, I will update it before final version.

src/ess/diffraction/filter.py
Pasarus: I have made this automatic

src/ess/diffraction/filter.py
Pasarus: We can do this yeah.

src/ess/diffraction/filter.py
Pasarus: Now an optional threshold passed into the function.

src/ess/diffraction/filter.py
Pasarus: I'm not sure it makes sense to do this, would you look at this again (as I can appreciate it's been 2 weeks since you said this) and let me know why we should do this?

src/ess/diffraction/filter.py
Pasarus: The main thing this function does is provide filtering for date inputs before putting them through the normal filtering, if you think I should remove it, I can do so. The aim is to simplify the use of Scipp and writing of workflows with dates in whatever format they are present, and provide the functionality available in Mantid.

Perhaps we should have a discussion as to what we are aiming to achieve with these functions long term? Perhaps with Celine?

src/ess/diffraction/filter.py
Pasarus: In some data produced at the SNS it uses both `time` and `pulse_time` interchangeably for attributes and the event dimensions/coordinates, hence all of this mapping, as I am not sure we know how the ESS intends to do this long term? I suppose we don't want to support hard coding of dim/coord names.

So we should just ensure/rely on whatever is passed in the good_data_lut has the correct dim-coordinate names?

src/ess/diffraction/filter.py
Pasarus: Not ideal, would you suggest something like a shallow copy instead?

src/ess/diffraction/filter.py
Pasarus: I can, however, it would at present only be used in this function and thus private so I am not sure it would be worth it. It will look like this:
```
def _find_edges_based_on_lut(da: sc.DataArray, good_data_lut, mapping_coord):
    good_data_coord = good_data_lut.coords[mapping_coord]
    edge = good_data_lut.data
    edge = edge[mapping_coord, :-1] ^ edge[mapping_coord, 1:]
    good_edges = sc.Dataset(data={
        mapping_coord: good_data_coord[mapping_coord, 1:],
        'good': good_data_lut.data[mapping_coord, :-1]
    },
                            coords={'edge': edge})
    return good_edges.groupby(group='edge').copy(True)
    

def filter_on_attribute(da: sc.DataArray,
                        good_data_lut,
                        mapping_coord: str,
                        data_slices_name: str = "slice_data"):
    edges = _find_edges_based_on_lut(da, good_data_lut, data_time_coord)
    good_data = sc.DataArray(data=good_edges['good'].data
                             ^ good_data_lut[mapping_coord, 0].data,
                             coords={mapping_coord: good_edges[mapping_coord].data})
    good_data = sc.concatenate(good_data_lut[mapping_coord, 0], good_data,
                               mapping_coord)
    good_data = sc.concatenate(good_data, good_data_lut[mapping_coord, -1],
                               mapping_coord)
    good_pulse_lut = sc.lookup(good_data, mapping_coord)

    da.bins.coords[data_slices_name] = good_pulse_lut[da.bins.coords[mapping_coord]]
    groups = sc.array(dims=[data_slices_name], values=[False, True])
    grouped = sc.bin(da, groups=[groups])
    return grouped
```

src/ess/diffraction/filter.py
SimonHeybrock: > So we should just ensure/rely on whatever is passed in the good_data_lut has the correct dim-coordinate names?

Yes. We have to avoid adding hacks for bad input values (this is imo one of the big problems of Mantid). This seriously complicates unit testing and documentation. It is hard to understand and not sustainable.

src/ess/diffraction/filter.py
arm61: I knew I hadn't pulled this style out of nowhere!

Pasarus: They should be up to standard now!

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,147 @@
...
+    good_data = sc.concatenate(good_data_lut[dim, 0], good_data, dim)
+    good_data = sc.concatenate(good_data, good_data_lut[dim, -1], dim)
+    good_pulse_lut = sc.lookup(good_data, dim)
+
+    return good_pulse_lut[da.bins.coords[dim]]
+
+
+def filter_by_attribute_edges(da: sc.DataArray,
+                              good_data_lut,
+                              data_slices_name: str = "slice_data"):
```
SimonHeybrock: This function name and the param names feel odd?
- The function does not "filter", but bin/split
- No attribute is involved.
- It is not splitting by edges, but based on data values, and actually only supports two specific valiues (true and false).
- `good_data_lut` does not make much sense?

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,147 @@
...
+      max_charge) where proton_charge is the attribute (sc.Variable) being filtered.
+
+    :param data_slices_name: str, The name of the dimension that will be added to the
+      passed sc.DataArray, da, that can then be sliced later for good and bad data.
+
+    :return: The originally passed sc.DataArray, da, with an extra dimension that can
+      then be sliced for good data on slice 1, and sliced for bad data on slice 0. For
+      example, da["slice_data", 1] for good data.
+    """
+    da.bins.coords[data_slices_name] = _find_da_coords_for_slicing(da, good_data_lut)
```
SimonHeybrock: Should this be deleted after the operation?

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,147 @@
...
+
+    # Group up data based on the in range data that is present and clean up groupby
+    # coordinate
+    filter_coord_name = '_'.join(da.coords)
+    da.coords[filter_coord_name] = in_range.data
+    data_in_range = da.groupby(filter_coord_name).copy(True)
+    del data_in_range.coords[filter_coord_name]
+    del da.coords[filter_coord_name]
+
+    return data_in_range
```
SimonHeybrock:
```suggestion
    filter_coord_name = '_'.join(da.coords)
    da.coords[filter_coord_name] = da.data
    data_in_range = da.groupby(filter_coord_name, bins=sc.concat([min_value, max_value], filter_coord_name)).copy(0)
    del data_in_range.coords[filter_coord_name]
    del da.coords[filter_coord_name]

    return data_in_range
```

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,147 @@
...
+
+    :return: The originally passed sc.DataArray, da, with an extra dimension that can
+      then be sliced for good data on slice 1, and sliced for bad data on slice 0. For
+      example, da["slice_data", 1] for good data.
+    """
+    da.bins.coords[data_slices_name] = _find_da_coords_for_slicing(da, good_data_lut)
+
+    groups = sc.array(dims=[data_slices_name], values=[False, True])
+    grouped = sc.bin(da, groups=[groups])
+    return grouped
```
SimonHeybrock: Given the difficulty of finding a good function name, maybe this should not exist at all (for now)? It is always hard to remove things once they were added (unless we make it really clear that it is experimental).

Maybe we can think about the underlying things we want to do (not saying this has to be done here and now, just thinking out loud):

1. We have a (time-)series-log, defining a value for each (time-)interval.
2. We are optimizing the log to merge intervals with identical values. Since this is simply to avoid performance issues when mapping the events into intervals we would probably want to avoid exposing this to the user.
3. Create new coord that will be used for binning.
4. Create grouping coord based on all unique values of the log.
5. Apply `sc.bin`

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,147 @@
...
+    # Auto find the time dim name from attribute, assume 1D DataArray
+    proton_charge_time_dim_initial_name = proton_charge.dims[0]
+    proton_charge = proton_charge \
+        .rename_dims({proton_charge_time_dim_initial_name: data_time_coord})
+    proton_charge.coords[data_time_coord] = \
+        proton_charge.coords.pop(proton_charge_time_dim_initial_name)
+
+    if center:
+        proton_charge.data.values[:-1] = 0.5 * (proton_charge.data.values[1:] +
+                                                proton_charge.data.values[:-1])
```
SimonHeybrock: To be honest I still don't really understand this bit.

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,147 @@
...
+
+    max_charge = sc.max(proton_charge.data)
+    max_charge *= maximum_threshold
```
SimonHeybrock, line 47: With this definition and `maximum_threshold > 1.0` below we will *always* fulfill `proton_charge < max_charge`?

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,147 @@
...
+    min_charge = sc.mean(proton_charge.data)
+    min_charge *= minimum_threshold
+
+    good_pulse = (proton_charge >= min_charge) & (proton_charge < max_charge)
```
SimonHeybrock: Instead of having a special function that does two things but only works for pulse-time filtering, would it make sense to have:

```python
# better name may be needed
# `select_band` could also to the joining of adjacent intervals
good = select_band(proton_charge, 0.95, 1.1).rename_dims({'time', 'pulse_time'})
da.bins.coords['good'] = sc.lookup(good)[da.bins.coords['pulse_time']]
sc.bin(da, groups=[sc.unique(good.data)])  # sc.unique is not implemented yet
```

SimonHeybrock (commented):
Some thoughts below, with some of the initial things cleaned up, maybe we can now see a bit better through the forest, which might eventually let us come up with a generic and simple API? Maybe be worth discussing in a call.

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,125 @@
...
+                        coords={mapping_coord: edges[mapping_coord].data})
+    data = sc.concatenate(condition_intervals[mapping_coord, 0], data, mapping_coord)
+    data = sc.concatenate(data, condition_intervals[mapping_coord, -1], mapping_coord)
+    return data
+
+
+def filter(da: sc.DataArray, condition_intervals: sc.DataArray, dim: str):
+    """
+    Filter a data array based on passed condition_intervals. Example input:
+     filter(da, ((temp >= min_temp) & (temp < max_temp)), "dim") where temp is the a
```
SimonHeybrock:
```suggestion
     filter(da, ((temp >= min_temp) & (temp < max_temp)), "temperature") where temp is the a
```

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,125 @@
...
+     sc.Variable or sc.DataArray.
+
+    :param da: sc.DataArray, the data that is being filtered
+    :param condition_intervals: sc.DataArray, a lookup table for each event time, each
+     event time coordinate should have true or false as the corresponding data of the
```
SimonHeybrock: Since we have generalized this it can be any event parameter, not just time?

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,125 @@
...
+     DataArray, to determine whether or not it should be filtered into slice 0 (False),
+     or 1 (True).
```
SimonHeybrock, line 106: Is this due to the current limitation of `_find_edges`?

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,125 @@
...
+    :param dim: str, the name of the coordinate, for which the slices will be
+     attributed.
+    :return: The original da, with  slices of given dim, sliced based on the event bands
```
SimonHeybrock: It is returning a new data array, not the original.

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,125 @@
...
+     in condition_intervals. Use this to access the results of your filter.
+    """
+    edges = _find_edges(condition_intervals)
+    coord_name = edges.dims[-1]
+    da.bins.coords[dim] = sc.lookup(edges, coord_name)[da.bins.coords[coord_name]]
```
SimonHeybrock, line 115: Should this be removed after the operation?

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,125 @@
...
+    :return: A DataArray that contains a dimension called "pulse_slices" in which "good"
+     data in slice 1, and "bad" data in slice "0". To access the good data you would
+     slice this return result like this: da["pulse_slices", 1].
+    """
+    # Auto find the time dim name from attribute, assume 1D DataArray
+    proton_charge_time_dim_initial_name = proton_charge.dims[0]
+    proton_charge = proton_charge \
+        .rename_dims({proton_charge_time_dim_initial_name: data_time_coord})
+    proton_charge.coords[data_time_coord] = \
+        proton_charge.coords.pop(proton_charge_time_dim_initial_name)
```
SimonHeybrock:
```suggestion
    proton_charge = proton_charge.transform_coords(data_time_coord,
        {data_time_coord: proton_charge.dims[0]},
        keep_inputs=False)
```

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,125 @@
...
+# Copyright (c) 2021 Scipp contributors (https://github.com/scipp)
+
+import scipp as sc
+import numpy as np
+
+
+def filter_bad_pulses(da: sc.DataArray,
+                      proton_charge: sc.DataArray,
+                      minimum_threshold: float = .95,
+                      maximum_threshold: float = 1.0,
```
SimonHeybrock: With this default, the interval that saw the maximum proton charge will be dropped, given how the condition is implemented below.

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,125 @@
...
+        proton_charge.coords.pop(proton_charge_time_dim_initial_name)
+
+    max_charge = sc.max(proton_charge.data)
+    max_charge *= maximum_threshold
+    min_charge = sc.mean(proton_charge.data)
+    min_charge *= minimum_threshold
+
+    good_pulse = (proton_charge >= min_charge) & (proton_charge < max_charge)
+
+    return filter(da=da, condition_intervals=good_pulse, dim="pulse_slices")
```
SimonHeybrock: Why did you chose `pulse_slices`, which does not say much? How about `'good_pulse'` or something like that?

SimonHeybrock (commented):
Some more comments below. I think we are getting closer to having found a more generic mechanism, but I suspect the API will need more iterations (later).

@nvaytet how should we proceed with this? We may want to avoid having too many people rely on this, given that we will change the API later on. Would it make sense to have an `experimental` submodule? Other ideas?

src/ess/diffraction/filter.py
Pasarus: So at the moment it is returning both good and bad pulses, hence `pulse_slices`

src/ess/diffraction/filter.py
Pasarus: Then I'll increase it back to 1.1

src/ess/diffraction/filter.py
Pasarus: Implemented

src/ess/diffraction/filter.py
Pasarus: Updated

src/ess/diffraction/filter.py
Pasarus: Yes

src/ess/diffraction/filter.py
Pasarus: Yes/no I am not sure where this would be useful, besides using it on event time?

src/ess/diffraction/filter.py
Pasarus: Done

src/ess/diffraction/filter.py
SimonHeybrock: You have a coord that says "False" and "True". So you would:
```python
da['pulse_slices', sc.scalar(True)]
```
to select the good data. That is meaningless?

If you name the coord `good_pulse` you have:
```python
da['good_pulse', sc.scalar(True)]
```
which clearly expresses intent.

src/ess/diffraction/filter.py
SimonHeybrock: Or change the math to include the upper bound.

src/ess/diffraction/filter.py
Pasarus: Ok I'll change it to good_pulse

src/ess/diffraction/filter.py
Pasarus: Yes, added the removal.

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,124 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2021 Scipp contributors (https://github.com/scipp)
+
+import scipp as sc
+import numpy as np
+
+
+def filter_bad_pulses(da: sc.DataArray,
```
nvaytet: Is this name from Mantid? Looking at the other functions below, I would have thought something like `filter_by_proton_charge` would be a better name?
It is subjective what one calls a "bad" pulse. The criterion may not always be proton charge?

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,124 @@
...
+                      maximum_threshold: float = 1.0,
+                      data_time_coord: str = "pulse_time"):
+    """
+    Filters out any bad pulses based on an attribute named "proton_charge" if the
+     "proton_charge" value is less than 95% of the mean (by default, the percentage can
+     be changed using "minimum_threshold") it will be filtered out, if a proton charge
+     is 10% higher than the maximum (This won't actually make any difference in
+     comparison to just the max value) based on how Mantid did filtering of bad pulses.
+
+    :param da: sc.DataArray, the data array that is to be filtered.
```
nvaytet: I believe the type hints should provide the information on the input type to sphinx, so I don't think you need to repeat them in the docstring. I think all you need is the whether it is optional, and the default value if applicable.

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,124 @@
...
+    """
+    Return all data points between a minimum and maximum values of da in a DataArray
+
+    :param da: sc.DataArray, the sc.DataArray that is being filtered
+
+    :param min_value: sc.Variable, the lower bound of the filtering for the attribute
+
+    :param max_value: sc.Variable, the upper bound of the filtering for the attribute
+
+    :return: sc.DataArray or sc.Variable, whichever the attribute is in the original da
```
nvaytet: I don't really understand in what case it could be a Variable, since `data_in_range` appears to always have coords below?
> `del data_in_range.coords[filter_coord_name]`

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,124 @@
...
+    da.coords[filter_coord_name] = da.data
+    data_in_range = da.groupby(filter_coord_name,
+                               bins=sc.concatenate(min_value, max_value,
+                                                   filter_coord_name)).copy(0)
+    del data_in_range.coords[filter_coord_name]
+    del da.coords[filter_coord_name]
+    return data_in_range
+
+
+def _find_edges(condition_intervals):
```
nvaytet: Could you add type hints and docstring here also? Thanks!

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,124 @@
...
+    mapping_coord = condition_intervals.dims[0]
```
nvaytet:
```suggestion
    mapping_dim = condition_intervals.dims[0]
```

tests/diffraction/filter_test.py
```diff
@@ -0,0 +1,89 @@
...
+from ess.diffraction.filter import filter_bad_pulses, filter_by_value, filter
+
+
+def _test_filter_sums_for_bad_pulses(filtered_da, bad_sum, good_sum):
+    good_filtered_da = filtered_da["good_pulse", 1]
+    bad_filtered_da = filtered_da["good_pulse", 0]
+    good_da_sum = sc.sum(good_filtered_da.bins.sum()).value
+    bad_da_sum = sc.sum(bad_filtered_da.bins.sum()).value
+
+    np.testing.assert_almost_equal(good_da_sum, good_sum)
```
nvaytet, line 16: we prefer to use plain `assert` instead of `np.testing.assert_almost_equal`,
e.g. `assert sc.identical(good_da_sum, bad_da_sum)` or use `sc.allclose()`

tests/diffraction/filter_test.py
```diff
@@ -0,0 +1,89 @@
...
+    np.testing.assert_almost_equal(bad_da_sum, bad_sum)
+
+
+def test_results_of_bad_pulse_defaults():
+    da = scn.data.tutorial_event_data()
```
nvaytet, line 21: Should we use generated data we have more control over, instead of actual data? so that we don't end up with magic numbers below.

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,124 @@
...
+    :return: sc.DataArray or sc.Variable, whichever the attribute is in the original da
+     that is passed.
+    """
+    filter_coord_name = '_'.join(da.coords)
+    da.coords[filter_coord_name] = da.data
+    data_in_range = da.groupby(filter_coord_name,
+                               bins=sc.concatenate(min_value, max_value,
+                                                   filter_coord_name)).copy(0)
+    del data_in_range.coords[filter_coord_name]
+    del da.coords[filter_coord_name]
```
SimonHeybrock: When the `bins` argument is used, `groupby` accepts a variable as the first argument, so we can avoid the temporary insertion of a new coord:
```suggestion
    data_in_range = da.groupby(da.data,
                               bins=sc.concatenate(min_value, max_value,
                                                   filter_coord_name)).copy(0)
    del data_in_range.coords[filter_coord_name]
```

Pasarus: Even with this change, it had the coord there, so I still needed to remove it, perhaps clean up isn't properly handled by groupby?

SimonHeybrock: I don't think `groupby` is adding coords to the input, can you double check?

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,124 @@
...
+     or 1 (True).
+    :param dim: str, the name of the coordinate, for which the slices will be
+     attributed.
+    :return: A new DataArray, a copy of the original da, with  slices of given dim,
+     sliced based on the event bands in condition_intervals. Use this to access the
+     results of your filter.
+    """
+    edges = _find_edges(condition_intervals)
+    coord_name = edges.dims[-1]
+    da.bins.coords[dim] = sc.lookup(edges, coord_name)[da.bins.coords[coord_name]]
```
SimonHeybrock, line 115: @nvaytet @Pasarus I occurs to me that it may make sense to do this "find edges" optimization within `sc.lookup`? The result should be unchanged, and it is essentially just for internal speedup. What do you think?

src/ess/diffraction/filter.py
Pasarus: I just looked at it and it seems it was removing a coordinate that shouldn't be removed. So you are correct.

tests/diffraction/filter_test.py
Pasarus: Whilst I agree plain assert would be better, assert almost equal is necessary due to float addition and rounding errors.

src/ess/diffraction/filter.py
Pasarus: :+1:

src/ess/diffraction/filter.py
Pasarus: This is a private function, not meant for users to use, hence no docs.

src/ess/diffraction/filter.py
Pasarus: I think this was a left over from a previous interpretation of filter by value, good spot!

src/ess/diffraction/filter.py
Pasarus: I'll remove the types from the doc strings. I will add the Optional and defaults for those that are.

src/ess/diffraction/filter.py
Pasarus: I think it is a better name I shall change it.

src/ess/diffraction/filter.py
nvaytet: Yes, but it still helps the code review process, you could add just a normal comment instead of a proper docstring if you prefer?
Not a big deal, but I would have spent less time understanding what the code was trying to achieve with a comment at the top.

tests/diffraction/filter_test.py
nvaytet: Yes, and you can use `sc.allclose` for that? and then you don't need to get the `.value`

src/ess/diffraction/filter.py
SimonHeybrock: This can actually be removed entirely, due to https://github.com/scipp/scipp/pull/2264. This should also remove the restriction to boolean inputs.

Pasarus: Should I remove this function and wait for scipp to release a new version?

src/ess/diffraction/filter.py
```diff
@@ -0,0 +1,122 @@
...
+
+    unique_values = np.unique(edges.data.values)
+    groups = sc.array(dims=[dim], values=unique_values)
+    grouped = sc.bin(da, groups=[groups])
```
SimonHeybrock, line 119: Should we consider supporting `str` in the list of `groups`, such that groups will be automatically determined? That is what `groupby` does, so it would be consistent. That is, we could then write:

```python
grouped = sc.bin(da, groups=[dim])
```
instead of the 3 lines above.

SimonHeybrock: Maybe not, because it could be quite inefficient in practice?

Pasarus: I am struggling to think of the benefit of supporting `str` in the list of groups? Would this be so we don't need to support `datetime64`? or would it be so we can use a dimension in the `da` opposed to pulling it out and then putting it back into `sc.bin?`

SimonHeybrock: The benefit would be to spare two lines of code to determine the possible groups, instead doing it automatically.

Pasarus: Yeah I see that now, would this be clear to users of scipp? Like do we do this in other places? and should we add them if not?

SimonHeybrock: > Like do we do this in other places?

We do (see above): https://scipp.github.io/user-guide/groupby.html

Pasarus: So it makes sense to support this in Scipp itself yes.

SimonHeybrock:
I had a final brief look at this. With the recent enhancements in Scipp (either already available, or with the next release) I think this is fully obsolete. This includes the units test. I'll there close this, unless there are objections.

PR #60, opened 2021-11-02
SANS Loki basic workflow

wpotrzebowski:
Basic workflow for LoKI demonstarted on SANS2D data.


src/ess/loki/load_files.py
```diff
@@ -0,0 +1,53 @@
+import scipp as sc
+import scippneutron as scn
+
+
+def load_isis(filename, spectrum_size, tof_bins):
+    """
+    Loading wrapper for ISIS files
+    """
+    events = scn.load(filename=filename, mantid_args={"LoadMonitors": True})
+    return sc.histogram(events["spectrum", :spectrum_size], tof_bins)
```
SimonHeybrock: Is this for the SANS2D files? Then maybe `load_sans2d` may be a better name?

src/ess/loki/load_files.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def load_and_apply_masks(idf_filename, mask_files, sample, background, spectrum_size):
+    """
+    Loading masks files from the list and add them to sample and background
+    """
+    for i, mask_file in enumerate(mask_files):
+        mask_xml = load_mask(idf_filename, mask_file)
+        sample.masks[f"mask_{i}_xml"] = mask_xml["spectrum", :spectrum_size].data
+        background.masks[f"mask_{i}_xml"] = mask_xml["spectrum", :spectrum_size].data
```
SimonHeybrock: - Looks SANS2D specific again (with the `spectrum_size` argument)?
- Is there a more sensible mask name than just enumerating? Maybe parts of the filename?

src/ess/sans/contrib.py
```diff
@@ -0,0 +1,80 @@
...
+    return 0.5 * (var[dim, 1:] + var[dim, :-1])
+
+
+def to_bin_centers(d, dim):
+    """
+    Utility function for setting centers of bins
+    """
+    edges = d.coords[dim].copy()
+    del d.coords[dim]
+    d.coords[dim] = 0.5 * (edges[dim, 1:] + edges[dim, :-1])
```
SimonHeybrock:
```suggestion
    d.coords[dim] = midpoints(d.coords.pop(dim))
```

src/ess/sans/contrib.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+def to_bin_edges(d, dim):
+    """
+    Utility function for setting bin edges
+    """
+    centers = d.coords[dim].copy()
```
SimonHeybrock:
```suggestion
    centers = d.coords[dim]
```
The copy is not necessary for any scipp>=0.7

src/ess/sans/contrib.py
```diff
@@ -0,0 +1,80 @@
...
+    del d.coords[dim]
+    first = 1.5 * centers[dim, 0] - 0.5 * centers[dim, 1]
+    last = 1.5 * centers[dim, -1] - 0.5 * centers[dim, -2]
+    bulk = 0.5 * (centers[dim, 1:] + centers[dim, :-1])
```
SimonHeybrock:
```suggestion
    bulk = midpoints(centers)
```

src/ess/sans/reduction.py
```diff
@@ -17,19 +22,22 @@ def reduce_to_q(data, *, q_bins, reducer, wavelength_bands=None):
...
     >>> reduced = reduce_to_q(data, q_bins=q_bins, reducer=simple_reducer('spectrum'))  # noqa: E501
     """
-    # TODO Backup of the coord is necessary until `convert` can keep original
     wavelength = data.coords['wavelength']
     data = scn.convert(data, 'wavelength', 'Q', scatter=True)
     if wavelength_bands is None:
         data = sc.histogram(data, q_bins)
         return reducer(data)
+
+    wavelength.rename_dims({'wavelength': 'Q'})
```
SimonHeybrock: In the most recent version of scipp this does not do anything: `rename_dims` does not change the input, but returns a new array.

src/ess/sans/reduction.py
```diff
@@ -17,19 +22,22 @@ def reduce_to_q(data, *, q_bins, reducer, wavelength_bands=None):
...
     data.coords['wavelength'] = wavelength
     bands = None
+
     for i in range(wavelength_bands.sizes['wavelength'] - 1):
         low = wavelength_bands['wavelength', i]
         high = wavelength_bands['wavelength', i + 1]
-        band = sc.histogram(data['wavelength', low:high], q_bins)
+        band = sc.histogram(data['wavelength', low:high].copy(), q_bins)
```
SimonHeybrock:
```suggestion
        band = sc.histogram(data['wavelength', low:high], q_bins)
```
`sc.histogram` does not modify the input, no need to copy.

src/ess/sans/reduction.py
```diff
@@ -17,19 +22,22 @@ def reduce_to_q(data, *, q_bins, reducer, wavelength_bands=None):
     Example:
     >>> reduced = reduce_to_q(data, q_bins=q_bins, reducer=simple_reducer('spectrum'))  # noqa: E501
     """
-    # TODO Backup of the coord is necessary until `convert` can keep original
     wavelength = data.coords['wavelength']
```
SimonHeybrock: Keeping the input is actually possible when using `transform_coords` directly. I think we should use this here instead of the `convert` wrapper. We can help you with making this change.

src/ess/sans/normalization.py
```diff
@@ -0,0 +1,51 @@
+import scipp as sc
+import scippneutron as scn
+
+
+def solid_angle(data, pixel_size, pixel_length):
+    """
+    Solid angle function taking pixel_size and pixel_lenght as parameters
```
SimonHeybrock: Mention that this assumes rectangular pixels and small angles?

src/ess/sans/normalization.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def transmission_fraction(
+    sample,
+    direct,
+    wavelength_bins,
+    min_bin_mon2,
+    max_bin_mon2,
+    min_bin_mon4,
+    max_bin_mon4,
```
SimonHeybrock: Give explicit names to parameters? 2 and 4 have no meaning.

src/ess/sans/normalization.py
```diff
@@ -0,0 +1,51 @@
...
+    Note: TOF ranges fro
+    """
+
+    def setup(data, begin, end, scatter):
+        transformed = data - sc.mean(data["tof", begin:end], "tof")
+        transformed = scn.convert(transformed, "tof", "wavelength", scatter=scatter)
+        transformed = sc.rebin(transformed.copy(), "wavelength", wavelength_bins)
+        return transformed
+
+    us = sc.units.us
```
SimonHeybrock: unused?

src/ess/sans/normalization.py
```diff
@@ -0,0 +1,51 @@
...
+    p = \frac{S_T}{D_T}\frac{D_I}{S_I}
+    This is equivalent to mantid.CalculateTransmission without fitting
+
+    Note: TOF ranges fro
+    """
+
+    def setup(data, begin, end, scatter):
+        transformed = data - sc.mean(data["tof", begin:end], "tof")
+        transformed = scn.convert(transformed, "tof", "wavelength", scatter=scatter)
+        transformed = sc.rebin(transformed.copy(), "wavelength", wavelength_bins)
```
SimonHeybrock:
```suggestion
        transformed = sc.rebin(transformed, "wavelength", wavelength_bins)
```

src/ess/sans/normalization.py
```diff
@@ -0,0 +1,51 @@
...
+    max_bin_mon2,
+    min_bin_mon4,
+    max_bin_mon4,
+):
+    """
+    Approximation based on equations in CalculateTransmission documentation
+    p = \frac{S_T}{D_T}\frac{D_I}{S_I}
+    This is equivalent to mantid.CalculateTransmission without fitting
+
+    Note: TOF ranges fro
```
SimonHeybrock: Truncated note?

src/ess/sans/normalization.py
```diff
@@ -0,0 +1,51 @@
...
+    def setup(data, begin, end, scatter):
+        transformed = data - sc.mean(data["tof", begin:end], "tof")
+        transformed = scn.convert(transformed, "tof", "wavelength", scatter=scatter)
+        transformed = sc.rebin(transformed.copy(), "wavelength", wavelength_bins)
+        return transformed
+
+    us = sc.units.us
+
+    sample_incident = setup(
+        sample.attrs["monitor2"].value, min_bin_mon2, max_bin_mon2, scatter=False
```
SimonHeybrock: So this only works for SANS2D? Should either make monitor names configurable, or change function/module name to include sans2d.

src/ess/sans/normalization.py
```diff
@@ -0,0 +1,51 @@
...
+):
+    """
+    Approximation based on equations in CalculateTransmission documentation
+    p = \frac{S_T}{D_T}\frac{D_I}{S_I}
+    This is equivalent to mantid.CalculateTransmission without fitting
+
+    Note: TOF ranges fro
+    """
+
+    def setup(data, begin, end, scatter):
```
SimonHeybrock: We appear to have `scatter=False` in all cases, maybe this param is not necessary?

src/ess/loki/sans2d/transform_coordinates.py
```diff
@@ -0,0 +1,31 @@
...
+    direct_beam,
+    sample_pos_z_offset,
+    bench_pos_y_offset,
+    monitor4_pos_z_offset,
+):
+    """
+    Transformin coordinates according to instrument setup
+    """
+    for item in [sample, sample_trans, background, background_trans, direct_beam]:
+        # for item in [sample,sample_trans,background,background_trans,directbeam]:
```
SimonHeybrock: commented code?

src/ess/sans/sans.py
```diff
@@ -0,0 +1,107 @@
...
+    masks,
+    wavelength_bins,
+    tof_bins_monitors,
+    pixel_size,
+    pixel_length,
+):
+    """
+    TOF to wavelength conversion
+    """
+    data = data.copy()
```
SimonHeybrock: The only modifying change to `data` below is setting masks, so a shallow copy should be enough here (and faster):
```suggestion
    data = data.copy(deep=False)
```

src/ess/sans/sans.py
```diff
@@ -0,0 +1,107 @@
...
+        direct,
+        wavelength_bins,
+        min_bin_mon2,
+        max_bin_mon2,
+        min_bin_mon4,
+        max_bin_mon4,
+    )
+    for name, mask in masks.items():
+        data.masks[name] = mask
+    data = scn.convert(data, "tof", "wavelength", out=data, scatter=True)
```
SimonHeybrock:
```suggestion
    data = scn.convert(data, "tof", "wavelength", scatter=True)
```
Recent versions of scippneutron do not support the `out` argument any more

src/ess/sans/sans.py
```diff
@@ -0,0 +1,107 @@
...
+    data = sc.rebin(data, "wavelength", wavelength_bins)
+
+    monitor = data.attrs["monitor2"].value
+    monitor = monitor - sc.mean(monitor["tof", min_bin_mon2:max_bin_mon2], "tof")
+    monitor = scn.convert(monitor, "tof", "wavelength", out=monitor, scatter=False)
+    monitor = sc.rebin(monitor, "wavelength", wavelength_bins)
```
SimonHeybrock: This looks identical to what is done above in the nested `setup()` function when computing transmission fraction? Can we reuse that function, or setup the monitors *outside* the transmission calculation function, so this is done only once?

src/ess/sans/contrib.py
```diff
@@ -0,0 +1,80 @@
...
+    bin_width.unit = sc.units.one
+    data *= bin_width
+    data = sc.rebin(data, dim, edges)
+    bin_width = edges[dim, 1:] - edges[dim, :-1]
+    bin_width.unit = sc.units.one
+    data /= bin_width
+    return data
+
+
+def select_bins(array, dim, start, end):
```
SimonHeybrock: Unused?

src/ess/sans/contrib.py
```diff
@@ -0,0 +1,80 @@
...
+    edges = coord.shape[0]
+    # scipp treats bins as closed on left and open on right: [left, right)
+    first = sc.sum(sc.less_equal(coord, start), dim).value - 1
+    last = edges - sc.sum(sc.greater(coord, end), dim).value
+    assert first >= 0
+    assert last < edges
+    return array[dim, first : last + 1]
+
+
+def make_slices(var, dim, cutting_points):
```
SimonHeybrock: Unused?

SimonHeybrock (commented):
Thanks for this addition! A number of comments below (if have not looked at the notebook yet). I think either @nvaytet or myself would pair-program with you to address some potentially more complex bits, and to ensure that this is in line with our latest guidelines (for example, helping in setting up unit tests).

src/ess/loki/load_files.py
```diff
@@ -0,0 +1,53 @@
+import scipp as sc
+import scippneutron as scn
+
+
+def load_isis(filename, spectrum_size, tof_bins):
```
nvaytet: You could put these functions into an `io` submodule

src/ess/loki/sans2d/SANS2D_reduction.ipynb
```diff
@@ -0,0 +1,532 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "try:\n",
+    "    import dataconfig # run make_config.py to create this\n",
```
nvaytet: We'll keep the `dataconfig` for now, but soon we will move all this to use `pooch` for a much nicer mechanism

src/ess/loki/sans2d/SANS2D_reduction.ipynb
```diff
@@ -0,0 +1,532 @@
...
+    "except:\n",
+    "    print(\"ERROR: dataconfig.py not find please run `make_config.py`\\n\")\n",
+    "    !python make_config.py -h # change to `-f path` or run in terminal\n",
+    "\n",
+    "path = dataconfig.data_root\n",
+    "direct_beam_file = 'DIRECT_SANS2D_REAR_34327_4m_8mm_16Feb16.dat'\n",
```
nvaytet: All these files should be put somewhere so that the notebook can find them so it can run. We should put them on the dmsc web server.

src/ess/loki/sans2d/SANS2D_reduction.ipynb
```diff
@@ -0,0 +1,532 @@
...
+    "#TODO: Are we using moderator file?\n",
+    "moderator_file = 'ModeratorStdDev_TS2_SANS_LETexptl_07Aug2015.txt'\n",
+    "\n",
+    "sample_run_number = 63114\n",
+    "sample_transmission_run_number = 63114\n",
+    "background_run_number = 63159\n",
+    "background_transmission_run_number = 63159 \n",
+    "direct_run_number = 63091\n",
+    "\n",
+    "mask_1 = f'{path}/MASK_SANS2D_REAR_Edges_16Mar2015.xml'\n",
```
nvaytet: I don't really understand the point of having all these masks in xml files.
From the names, it seems they are masking either tubes or physical regions in space.

We should be able to come up with a programmatic way of masking the tubes, maybe if we reshape the data into logical (`tube,straw,pixel`) dimensions?

Maybe some masking based on count values would also be more compact?

src/ess/loki/sans2d/SANS2D_reduction.ipynb
```diff
@@ -0,0 +1,532 @@
...
+    "### Do coordinate transformations"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "setup_offsets(sample, sample_trans, background, background_trans, direct, sample_pos_z_offset, bench_pos_y_offset, monitor4_pos_z_offset)"
```
nvaytet: I think the interface to this function is slightly odd. It assumes you always have the same list of input DataArrays, which hampers portability.
The function should probably just take in a single DataArray to apply corrections on, and then you call the function on the DataArrays you wish to correct.

In addition, have you thought about putting your data in a Dataset, so you can do these kind of things in a single call?

src/ess/loki/sans2d/SANS2D_reduction.ipynb
```diff
@@ -0,0 +1,532 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "setup_geometry(sample, background, direct, x_offset, y_offset, z_offset)"
```
nvaytet: Same as above. (and you can see that here the function only expects 3 input DataArrays, which is confusing)

src/ess/loki/load_files.py
nvaytet: Is it intentional to convert immediately to histogram? You lose some flexibility here.

src/ess/loki/sans2d/SANS2D_reduction.ipynb
```diff
@@ -0,0 +1,532 @@
...
+    "min_bin_mon2 = 85000.0 * sc.units.us\n",
+    "max_bin_mon2 = 98000.0 * sc.units.us\n",
+    "min_bin_mon4 = 85000.0 * sc.units.us\n",
+    "max_bin_mon4 = 98000.0 * sc.units.us\n",
+    "tof_bins_monitors = [min_bin_mon2, max_bin_mon2, min_bin_mon4, max_bin_mon4] \n",
+    "#Solid angle values\n",
+    "pixel_size = 0.00405 * sc.units.m\n",
+    "pixel_length = 0.002033984375 * sc.units.m\n",
+    "\n",
+    "#Coordinate trasnformation\n",
```
nvaytet:
```suggestion
    "#Coordinate transformation\n",
```

src/ess/loki/sans2d/SANS2D_reduction.ipynb
```diff
@@ -0,0 +1,532 @@
...
+   "outputs": [],
+   "source": [
+    "reduced_by_wavelength = sans.normalize_and_subtract(sample_q_reduce_wav, background_q_reduce_wav)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Comoparison reduced vs reduced_by_wavelength"
```
nvaytet:
```suggestion
    "## Comparison reduced vs reduced_by_wavelength"
```

src/ess/loki/sans2d/transform_coordinates.py
```diff
@@ -0,0 +1,31 @@
...
+
+def setup_offsets(
+    sample,
+    sample_trans,
+    background,
+    background_trans,
+    direct_beam,
+    sample_pos_z_offset,
+    bench_pos_y_offset,
+    monitor4_pos_z_offset,
```
nvaytet: I think function definitions should not have trailing commas (although style guides are a bit unclear on this)

src/ess/loki/sans2d/transform_coordinates.py
```diff
@@ -0,0 +1,31 @@
+import scipp as sc
+
+
+def setup_offsets(
```
nvaytet: Change to accept only one input data array.
You should also add type hints to the arguments

src/ess/sans/reduction.py
nvaytet: Why is the copy needed?

src/ess/sans/sans.py
nvaytet: Why the copy?

src/ess/sans/sans.py
```diff
@@ -0,0 +1,107 @@
...
+    pixel_size,
+    pixel_length,
+):
+    """
+    TOF to wavelength conversion
+    """
+    data = data.copy()
+    min_bin_mon2, max_bin_mon2, min_bin_mon4, max_bin_mon4 = tof_bins_monitors
+    transmission_frac = transmission_fraction(
+        transmission,
```
nvaytet: This would be a good place to use keyword arguments

src/ess/sans/sans.py
```diff
@@ -0,0 +1,107 @@
+# SANS specific functions
+import scipp as sc
+import scippneutron as scn
+from .contrib import map_to_bins, to_bin_centers
+from .reduction import reduce_to_q, simple_reducer, grouping_reducer
+from .normalization import solid_angle, transmission_fraction
+
+
+def to_wavelength(
```
nvaytet: Can we do this in a more succint way via `transform_coords`?

nvaytet (commented):
Some initial thoughts/comments.

src/ess/loki/sans2d/SANS2D_reduction.ipynb
SimonHeybrock: I agree, but I believe this is simply for reproducing specific SANS2D results, so maybe ok for that, as long as we ensure that for ESS instruments a better mechanism is put in place?

src/ess/sans/sans.py
SimonHeybrock: Maybe not, because this is essentially doing half of a SANS data reduction?

wpotrzebowski:
@SimonHeybrock @nvaytet thanks for useful comments! I will start off from addressing obvious cases and then indeed pair programing session(s) would be very useful. I can also see some more open questions e.g. xml mask files that may require some broader discussion but I guess we can take it at the later stage.

src/ess/sans/contrib.py
wpotrzebowski: @SimonHeybrock following up on previous suggestion shouldn't it be something like `bulk = midpoints(centers.pop(dim))`?

src/ess/sans/reduction.py
wpotrzebowski: @SimonHeybrock do you suggest to do anything with this code or leave it as it is for now?

src/ess/sans/contrib.py
SimonHeybrock: `centers` is a variable, you cannot `pop` from it, unless I misunderstand your question?

src/ess/sans/reduction.py
SimonHeybrock: I am talking about the released version of scipp (0.8), so no point in leaving code that only works with 0.7 and below?

src/ess/loki/load_sans2d.py
```diff
@@ -0,0 +1,69 @@
...
+        mantid_alg="LoadMask",
+        mantid_args={"InputFile": mask_file},
+    )
+
+
+def load_and_apply_masks(
+    idf_filename,
+    mask_files,
+    data_set,
+    spectrum_size,
```
SimonHeybrock: Why is this parameter required? Isn't it just `data_set.sizes['spectrum']`?

Note that we typically use `ds` as a generic variable names for datasets (and `da` for data arrays).

src/ess/sans/contrib.py
```diff
@@ -0,0 +1,47 @@
...
+    Utility function for setting bin edges
+    """
+    centers = d.coords[dim]
+    del d.coords[dim]
+    first = 1.5 * centers[dim, 0] - 0.5 * centers[dim, 1]
+    last = 1.5 * centers[dim, -1] - 0.5 * centers[dim, -2]
+    bulk = midpoints(centers, dim)
+    edges = sc.concatenate(first, bulk, dim)
+    edges = sc.concatenate(edges, last, dim)
+    d.coords[dim] = edges
```
SimonHeybrock: `sc.concatenate` is deprecated, use
```suggestion
    d.coords[dim] = sc.concat([first, bulk, last], dim)
```
unless you are stuck with and old scipp for some reason.

src/ess/sans/normalization.py
```diff
@@ -0,0 +1,63 @@
...
+def solid_angle(data, pixel_size, pixel_length):
+    """
+    Solid angle function taking pixel_size and pixel_lenght as parameters.
+    The method assumes pixels are rectangular.
+    """
+    L2 = scn.L2(data)
+    return (pixel_size * pixel_length) / (L2 * L2)
+
+
+def covert_and_rebin(data, wavelength_bins, begin, end):
```
SimonHeybrock:
```suggestion
def convert_and_rebin(data, wavelength_bins, begin, end):
```
... but convert to what?

src/ess/sans/normalization.py
```diff
@@ -0,0 +1,63 @@
...
+    """
+    Convertining and rebinning shifted data
```
SimonHeybrock:
```suggestion
    Convertining and rebinning shifted data
```
This docstring does not say more than the function name? What is the actual function purpose?

src/ess/sans/sans.py
```diff
@@ -0,0 +1,111 @@
...
+
+    if wavelength_bands == None:
+        return reduce_to_q(wav, q_bins=q_bins, reducer=reducer)
+    else:
+        if groupby != None:
+            reducer = grouping_reducer(dim="spectrum", group=groupby)
+
+        return reduce_to_q(
+            wav, q_bins=q_bins, reducer=reducer, wavelength_bands=wavelength_bands
+        )
```
SimonHeybrock: Why is grouping not supported if there are no wavelength bands?

Unless there is an issue with this, we could then write this simpler:

```suggestion
    if groupby is None:
      reducer = simple_reducer(dim="spectrum")
    else:
      reducer = grouping_reducer(dim="spectrum", group=groupby)
    return reduce_to_q(wav, q_bins=q_bins, reducer=reducer, wavelength_bands=wavelength_bands)
```

SimonHeybrock (commented):
Just a few minor more comments/questions from my side:

src/ess/sans/normalization.py
wpotrzebowski: Before this function was called `setup ` and I thought I can give it more descriptive name. Can revert it to `setup` if it works better.

SimonHeybrock: I think if a function is just used within a file then a good name may not be so important, but it seems this is imported elsewhere (`sans.py`) so we probably want something that is truly descriptive?

wpotrzebowski: Indeed, it was initially used internally in one file but then it turned out that can be applied elswhere. Maybe `contrib.py `is anyway better place for such helper function?

SimonHeybrock: `contrib.py` was menat for things that may eventuallt make it into scipp itself. I think that is quite unlikely for this particular function.

nvaytet:
@SimonHeybrock @wpotrzebowski 
Please have another look at this. There is now a `sans.q1d` function.
The details of what the `q1d` function does are illustrated in the `sans2d_q1d` notebook.
The `q1d` function is then used on the `sample` and the `background` in the `sans2d_reduction` notebook.

@wpotrzebowski we need to add some descriptive comments in the `sans2d_q1d` notebook (for example, describing the unit conversion with gravity, a bit like in the amor notebook). Maybe you can have a go at this?

src/ess/imaging/__init__.py
```diff
@@ -1,2 +1,4 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2022 Scipp contributors (https://github.com/scipp)
+
+from . import mantid
```
SimonHeybrock: Are there no components in the imaging module than can run without Mantid?

src/ess/loki/data.py
```diff
@@ -0,0 +1,35 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2022 Scipp contributors (https://github.com/scipp)
+_version = '1'
+
+__all__ = ['get_path']
+
+
+def _make_pooch():
```
SimonHeybrock, line 8: Why did you choose to make this `ess.loki.data` instead of a central `ess.data`?

src/ess/sans/conversions.py
```diff
@@ -0,0 +1,56 @@
...
+    out = sc.asin(out, out=out)
+    return out
+
+
+def sans_elastic() -> dict:
+    """
+    Generate a coordinate transformation graph for SANS elastic scattering.
+    """
+    graph = {**conversions.beamline(scatter=True), **conversions.elastic_Q("tof")}
+    graph["two_theta"] = two_theta
```
SimonHeybrock: There is a quite substantial cost to conversion with gravity. Have you considered having a `gravity=False` default option of `sans_elastic`, replacing this node only if requested?

src/ess/sans/conversions.py
```diff
@@ -0,0 +1,56 @@
...
+    return graph
+
+
+def sans_monitor() -> dict:
+    """
+    Generate a coordinate transformation graph for SANS monitor (no scattering).
+    """
+    return {**conversions.beamline(scatter=False), **conversions.elastic_Q("tof")}
```
SimonHeybrock: Q? Don't we have a wavelength graph somewhere? Otherwise we risk misuse of this graph.

src/ess/sans/normalization.py
```diff
@@ -0,0 +1,43 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2022 Scipp contributors (https://github.com/scipp)
+
+import scipp as sc
+import scippneutron as scn
+
+
+def solid_angle(data, pixel_width, pixel_height):
+    """
+    Solid angle computed from rectangular pixels with a 'width' and a 'height'.
```
SimonHeybrock, line 11: Maybe the rectangular nature should be reflected in the function name? And also the fact that this is an approximation (for small angles w.r.t. to the rectangle normal).

src/ess/sans/normalization.py
```diff
@@ -0,0 +1,43 @@
...
+
+def subtract_background_and_rebin(data, wavelength_bins, threshold):
+    """
+    Subtracts background value from data counts and performs a wavelength rebin.
+    The background is computed as the mean value of all the counts below the given
+    threshold.
+    """
+    data_original = data.copy(deep=False)
+    data_original.masks['background'] = data_original.data > threshold
+    background = sc.mean(data_original)
```
SimonHeybrock: This feels like shady business. Masking based on value risks masking naturally low points in a "region" with valid data. This is a bit like "let us strip peaks and call what is left background". The original approach (selecting a known region with "no" data), while not perfect, seems less fudged?

src/ess/sans/normalization.py
```diff
@@ -0,0 +1,43 @@
...
+def transmission_fraction(data_incident_monitor, data_transmission_monitor,
+                          direct_incident_monitor, direct_transmission_monitor):
+    """
+    Approximation based on equations in CalculateTransmission documentation
+    p = \frac{S_T}{D_T}\frac{D_I}{S_I}
+    This is equivalent to mantid.CalculateTransmission without fitting.
+
+    TODO: It seems we are always multiplying this by data_incident_monitor to compute
+    the normalization term. We could consider just returning
+    data_transmission_monitor * direct_incident_monitor / direct_transmission_monitor
```
SimonHeybrock: You mean elsewhere, outside this function?

src/ess/sans/q1d.py
```diff
@@ -0,0 +1,128 @@
...
+
+def q1d(data,
+        data_incident_monitor,
+        data_transmission_monitor,
+        direct_incident_monitor,
+        direct_transmission_monitor,
+        direct_beam,
+        wavelength_bins,
+        q_bins,
+        number_of_wavelength_bands=1,
```
SimonHeybrock: Indirect control over wavelength bands? What not make the function signature consistent and flexible and have `wavelength_bands` accepting a `Variable`, just like `wavelength_bins` and `q_bins`? Make the default `None` to not have bands?

src/ess/sans/q1d.py
```diff
@@ -0,0 +1,128 @@
...
+        monitor_background_threshold=sc.scalar(30.0, unit='counts'),
```
SimonHeybrock: I feel this will go badly wrong. Sooner or later it will be forgotten to set this value.

src/ess/sans/q1d.py
```diff
@@ -0,0 +1,128 @@
...
+        histogram_output=True):
```
SimonHeybrock: Flags (True/False) often indicate that the function is too complicated and is doing too much.

src/ess/sans/q1d.py
```diff
@@ -0,0 +1,128 @@
...
+        'direct_incident_monitor': direct_incident_monitor,
+        'direct_transmission_monitor': direct_transmission_monitor
+    }
+
+    # Create unit conversion graphs
+    graph = conversions.sans_elastic()
+    graph_monitor = conversions.sans_monitor()
+
+    # Add gravity coordinate
+    data.coords["gravity"] = sc.vector(value=[0, -1, 0]) * g
```
SimonHeybrock: Define the gravity vector outside this function so it is accessible from `ess.sans`? This makes it transparent and less likely that we will later end up with two different definitions of gravity direction for the same instrument.

src/ess/sans/q1d.py
```diff
@@ -0,0 +1,128 @@
...
+# Copyright (c) 2022 Scipp contributors (https://github.com/scipp)
+import scipp as sc
+from scipp.constants import g
+from . import conversions
+from . import normalization
+from ..wfm.tools import to_bin_centers, to_bin_edges
+from scipp.interpolate import interp1d
+
+
+def q1d(data,
```
SimonHeybrock: I see below that this function is doing basically the entire data reduction, maybe this should be indicated in the name? Also, this is doing way more than Mantid's `Q1D`, so I am not sure about the naming?

src/ess/sans/q1d.py
```diff
@@ -0,0 +1,128 @@
...
+                to_bin_edges(direct_beam.coords['wavelength'], dim='wavelength'),
+                wavelength_bins):
+            interpolate_direct_beam = False
+    if interpolate_direct_beam:
+        func = interp1d(sc.values(direct_beam), 'wavelength')
+        direct_beam = func(wavelength_bins, midpoints=True)
+        logger = sc.get_logger()
+        logger.warning('ess.sans.q1d: An interpolation was performed on the '
+                       'direct_beam function. The variances in the direct_beam '
+                       'function have been ignored.')
```
SimonHeybrock: Dropped, not just ignored?

src/ess/sans/q1d.py
```diff
@@ -0,0 +1,128 @@
...
+
+    # Compute denominator and copy coords needed for conversion to Q
+    denominator = (solid_angle * direct_beam * monitors_wav['data_incident_monitor'] *
+                   transmission_fraction)
+    denominator.coords['position'] = data.meta['position']
+    denominator.coords['gravity'] = data.meta['gravity']
+    denominator.coords['sample_position'] = data.meta['sample_position']
+    denominator.coords['source_position'] = data.meta['source_position']
+    denominator.coords['wavelength'] = to_bin_centers(denominator.coords['wavelength'],
+                                                      dim='wavelength')
```
SimonHeybrock: This feels a bit error-prone. Can we just copy the final `Q`? Or transform coords on a dataset? The former may not be possible because we may have event data in the numerator?

src/ess/sans/q1d.py
```diff
@@ -0,0 +1,128 @@
...
+
+    # 3. Convert from wavelength to Q ==================================================
+
+    wavelength_bands = sc.linspace(dim='wavelength',
+                                   start=wavelength_bins.min().value,
+                                   stop=wavelength_bins.max().value,
+                                   num=number_of_wavelength_bands + 1,
+                                   unit=wavelength_bins.unit)
```
SimonHeybrock: See above, I would avoid hard-coding this here.

If there are no wavelength bands you end up computing `min` and `max` for nothing, and that is not entirely for free.

src/ess/sans/q1d.py
```diff
@@ -0,0 +1,128 @@
...
+
+    data = data.transform_coords("Q", graph=graph)
+    q_boundaries = sc.concat([q_bins.min(), q_bins.max()], dim='Q')
+
+    # TODO: once scipp-0.12 is out, we no longer need to move the attr into the coords
+    data.bins.coords['wavelength'] = data.bins.attrs.pop('wavelength')
```
SimonHeybrock: We could also make a patch release.

src/ess/sans/q1d.py
```diff
@@ -0,0 +1,128 @@
...
+
+    q_binned = sc.bin(data, edges=[wavelength_bands, q_boundaries])
+    q_summed = q_binned.bins.concat('spectrum')
+
+    denominator_bands = []
+    for i in range(number_of_wavelength_bands):
+        band = denominator['wavelength',
+                           wavelength_bands['wavelength',
+                                            i]:wavelength_bands['wavelength', i + 1]]
```
SimonHeybrock: Note that you can now write:
```suggestion
        band = denominator['wavelength', wavelength_bands[i]:wavelength_bands[i + 1]]
```

src/ess/sans/q1d.py
```diff
@@ -0,0 +1,128 @@
...
+        q_band = band.transform_coords("Q", graph=graph)
+        denominator_bands.append(sc.histogram(q_band, bins=q_bins))
+    denominator_q_summed = sc.concat(denominator_bands, 'wavelength').sum('spectrum')
```
SimonHeybrock: Wouldn't it be equivalent but more efficient to `sum('spectrum')` *before* `concat`?

src/ess/sans/q1d.py
```diff
@@ -0,0 +1,128 @@
...
+
+    q_binned = sc.bin(data, edges=[wavelength_bands, q_boundaries])
+    q_summed = q_binned.bins.concat('spectrum')
+
+    denominator_bands = []
+    for i in range(number_of_wavelength_bands):
+        band = denominator['wavelength',
+                           wavelength_bands['wavelength',
+                                            i]:wavelength_bands['wavelength', i + 1]]
+        q_band = band.transform_coords("Q", graph=graph)
```
SimonHeybrock: Have you considered calling `transform_coords` only once, before slicing? That should be more efficient (may actually matter a bit with many pixels, since geometry calculations may be costly, especially with gravity).

src/ess/sans/q1d.py
```diff
@@ -0,0 +1,128 @@
...
+                                   num=number_of_wavelength_bands + 1,
+                                   unit=wavelength_bins.unit)
+
+    data = data.transform_coords("Q", graph=graph)
+    q_boundaries = sc.concat([q_bins.min(), q_bins.max()], dim='Q')
+
+    # TODO: once scipp-0.12 is out, we no longer need to move the attr into the coords
+    data.bins.coords['wavelength'] = data.bins.attrs.pop('wavelength')
+
+    q_binned = sc.bin(data, edges=[wavelength_bands, q_boundaries])
```
SimonHeybrock: If there are no wavelength bands this forces an iteration over all events (or rather their wavelength coord) for nothing, slowing things down.

src/ess/sans/q1d.py
```diff
@@ -0,0 +1,128 @@
...
+    denominator_q_summed = sc.concat(denominator_bands, 'wavelength').sum('spectrum')
+
+    # 4. Normalize =====================================================================
+
+    normalized = q_summed.bins / sc.lookup(func=denominator_q_summed, dim='Q')
+    if histogram_output:
+        normalized = sc.histogram(normalized, bins=q_bins)
+    if number_of_wavelength_bands == 1:
+        normalized = normalized['wavelength', 0]
+    return normalized
```
SimonHeybrock: I feel many parts of this function should be extracted into functions. Otherwise the user will have to copy and modify the entire 120 lines for every tweak they may need. You have already used comments to indicate "sections" of this function, which goes to show that such a split may be very natural. The function here would simply make calls to the lower-level blocks.

tests/sans/normalization_test.py
```diff
@@ -0,0 +1,32 @@
...
+    direct_incident_monitor = sc.scalar(4., unit='counts')
+    direct_transmission_monitor = sc.scalar(5., unit='counts')
+    expected = (data_transmission_monitor / direct_transmission_monitor) * (
+        direct_incident_monitor / data_incident_monitor)
+    result = normalization.transmission_fraction(
+        data_incident_monitor=data_incident_monitor,
+        data_transmission_monitor=data_transmission_monitor,
+        direct_incident_monitor=direct_incident_monitor,
+        direct_transmission_monitor=direct_transmission_monitor)
+    assert sc.allclose(result, expected)
```
SimonHeybrock: It looks like this is just repeating the equation from the function? That is, it is not testing anything?

SimonHeybrock (commented):
I have not looked at the notebooks yet:

src/ess/imaging/__init__.py
nvaytet: Nice catch, thanks, i think this left over from trying to debug the mantid tests. Will remove

src/ess/loki/data.py
nvaytet: This is what I had done for `amor`, so I did the same here.
Should we leave it for now and rework this and the `amor.data` in a follow-up PR?
(cf discussion on slack)

src/ess/sans/conversions.py
nvaytet: Yes, but I need the `Q` as well further in my workflow. So I might as well use a single graph?
Unless I misunderstood something

nvaytet: hmm actually, i think you're right, monitors are only converted to wavelength.
i'll change

src/ess/sans/normalization.py
nvaytet: Yes, i know you mentioned this a couple of times now, and i keep forgetting about it. I will change to a wavelength range.

src/ess/sans/normalization.py
nvaytet: yes

src/ess/loki/data.py
SimonHeybrock: 👍

src/ess/sans/q1d.py
nvaytet: This is actually a bit of a trick, because before I introduced wavelength bands, i still needed to do a `sc.bin` operation into a single wavelength bin in order to select the same wavelength range as the denominator (the issue I was having with normalizing that we discussed over a call).

It is now kind of hidden, so I should make this clearer, and I'm going to re-work the wavelength bands anyway, so it should be more explicit

src/ess/sans/q1d.py
nvaytet: make a helper in `ess.sans` or just let the user add it manually to the data in the notebook?
If the former, any suggestions for a good place to put it?

SimonHeybrock: Since users cannot change the direction of gravity of say, LoKI, this should be in `ess.loki`. Problem is of course that this is needed in `ess.sans`. Unless there is a common definition for all (SANS) instruments at ESS we would thus need to inject this somehow.

src/ess/sans/q1d.py
SimonHeybrock: 👍

src/ess/sans/q1d.py
nvaytet: I'm also thinking of the `q1d` function signature (whose name may change).
we could
1. have a `gravity=True/False` argument
2. use gravity correction if a `'gravity'` coordinate is found in the input DataArray

In case 1, this means if you load your data and set all your coordinates, including the gravity coordinate, one could experiment with turning on and off the gravity effects in the unit conversion without having to modify the DataArray.
However. if we set `gravity=False` by default, and a user manually sets a `gravity` coordinate, and forgets or doesn't know of the `gravity` argument to the function, they might think gravity is being applied while in fact it isn't?

In case 2, if you want to experiement turning gravity on/off, you have to modify your data in-place, which may not be desireable.

SimonHeybrock: I don't think `gravity` should be taken from a data coord/attr. It is not really something that can change, so it should be hard-coded in the module (`ess`, `ess.sans`, or `ess.loki`). So option 1, but don't rely on coords/attrs.

src/ess/sans/q1d.py
nvaytet: Not sure what you mean by copying the final `Q`.
- The `solid_angle` is just a Variable, it doesn't keep the coords of the original DataArray, just uses its positions to compute solid angle
- The `direct_beam` is loaded from a file, and only has counts vs wavelength.
- The incident monitor has monitor coords, does not have the original pixels coords, nor the sample position.
- The transmission fraction is a ratio of monitor counts, i.e. same as above.

I don't know where else to get the coords from.

SimonHeybrock: My point was: Instead of using `transform_coords` on the denominator, just copy over the Q coord from transforming for the numerator.

nvaytet: but the numerator is event data, with a single Q bin...

nvaytet: you do raise an important point though, that in its current state the `q1d` function only works with event data, and we should think about making it accept both event and histogrammed data...

SimonHeybrock: > but the numerator is event data, with a single Q bin...

Yes, as I wrote in the original comment.

> you do raise an important point though, that in its current state the q1d function only works with event data, and we should think about making it accept both event and histogrammed data...

This should be possible more easily if we extract sections of this long function into helpers.

src/ess/sans/q1d.py
nvaytet: Done.

src/ess/sans/q1d.py
nvaytet: Done.

src/ess/sans/q1d.py
nvaytet: Removed

src/ess/sans/q1d.py
nvaytet: Don't have any good ideas.
It's basically making as an output a quantity that is 1D that represents `I(Q)`
`i_of_q_1d` looks very weird

src/ess/sans/q1d.py
nvaytet: I split it up

src/ess/sans/q1d.py
nvaytet: Done

tests/sans/normalization_test.py
nvaytet: Yes i agree. At the same time, i don't find it particularly useful to just compute the values I want using the formula, and then hard-coding those as the expected values in the test.

I have to say I've really struggled to come up with ideas for testing all this. It basically feels like it should be a system test, with an approved reference solution. Because testing the individual parts doesn't seem very useful?

So for now, maybe I will just remove them?

src/ess/sans/q1d.py
SimonHeybrock: No really good idea here either:
- `reduce` -- there may be a benefit of naming all reduction functions across techniques the same, so you always know what to look for, but maybe too generic?
- `to_intensity`?
- `I_of_Q`?

tests/sans/normalization_test.py
SimonHeybrock: > Yes i agree. At the same time, i don't find it particularly useful to just compute the values I want using the formula, and then hard-coding those as the expected values in the test.

We can try asking ourselves what behavior we expect. For `solid_angle`, this could be:

- If width or height is doubled, the solid angle should double.
- If distance is doubled, solid angle should be 1/4

src/ess/sans/q1d.py
nvaytet: > `reduce` -- there may be a benefit of naming all reduction functions across techniques the same, so you always know what to look for, but maybe too generic?

I think this is not really possible because you might want to reduce to q1d or to Qxy?

> `I_of_Q`

Uppercase would suggest it constructs a class?

SimonHeybrock: `to_I_of_Q`? I think in cases where there is a physical equation with standard symbols we should keep the case.

nvaytet: agreed. I will go with `to_I_of_Q`

src/ess/sans/normalization.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def solid_angle_of_rectangular_pixels(data: sc.DataArray, pixel_width: sc.Variable,
+                                      pixel_height: sc.Variable) -> sc.Variable:
+    """
+    Solid angle computed from rectangular pixels with a 'width' and a 'height'.
+
+    Note that this is an approximation which is only valid for small angles
+    between the line of sight and the rectangle normal.
+    This is usually a reasonable approximation for SANS experiments.
```
SimonHeybrock: Is this actually true for the LoKI window frame? It is quite a large angle.

src/ess/sans/normalization.py
```diff
@@ -0,0 +1,77 @@
...
+        does not constitute background. Everything outside this range is treated as
+        background counts.
+    """
+    if non_background_range is not None:
+        data_original = data.copy(deep=False)
+        wav_midpoints = 0.5 * (data_original.coords['wavelength'][1:] +
+                               data_original.coords['wavelength'][:-1])
+        data_original.masks['non_background'] = (
+            wav_midpoints > non_background_range[0]) & (wav_midpoints <
+                                                        non_background_range[1])
```
SimonHeybrock: `scipp.core.get_slice_params` might be able to do this in a nicer/safer manner. Have a look at how it is used in the plotting code.

src/ess/sans/to_I_of_Q.py
```diff
@@ -0,0 +1,274 @@
...
+    data_q = data.transform_coords("Q", graph=graph)
+    data_q.coords['wavelength'] = data_q.attrs.pop('wavelength')
+    for i in range(wavelength_bands.sizes['wavelength'] - 1):
+        band = data_q['wavelength', wavelength_bands[i]:wavelength_bands[i + 1]]
+        bands.append(sc.histogram(band, bins=q_bins).sum('spectrum'))
+    q_summed = sc.concat(bands, 'wavelength')
+    return q_summed
+
+
+def _normalize(numerator: sc.DataArray,
```
SimonHeybrock: I think we would want to make these functions publicly accessible, maybe?

nvaytet: By 'these functions', do you mean all of the ones in the file, or just the `normalize` here?
As I see that one indeed being useful, but not sure the others are so useful?

SimonHeybrock: I mean all components of the top-level I(Q) function. They could be kept in this submodule (not imported in `__init__.py`), but there is no need to hide them?

src/ess/sans/normalization.py
nvaytet: maybe that's a little overkill?
what about
```Py
before = data['wavelength', :non_background_range[0]].data
after = data['wavelength', non_background_range[1]:].data
background = (before.sum() + after.sum()) / sc.scalar(before.sizes['wavelength'] + after.sizes['wavelength'])
```
?

src/ess/sans/to_I_of_Q.py
nvaytet: Should the notebook that describes in more detail the reduction workflow (`sans2d_to_I_of_Q.ipynb`) then just call these functions one after the other, instead of repeating the code from the python file in the notebook?

Because I have been wondering about how I was going to keep those two things in-sync...
But maybe this is a nice idea. The notebook would just call those steps and make visualizations of the intermediate results?

src/ess/sans/normalization.py
```diff
@@ -39,13 +38,11 @@ def subtract_background_and_rebin(
...
-                               data_original.coords['wavelength'][:-1])
-        data_original.masks['non_background'] = (
-            wav_midpoints > non_background_range[0]) & (wav_midpoints <
-                                                        non_background_range[1])
-        background = sc.mean(data_original)
+        dim = non_background_range.dim
+        below = data[dim, :non_background_range[0]].data
+        above = data[dim, non_background_range[1]:].data
+        background = (below.sum() + above.sum()) / sc.scalar(below.sizes[dim] +
+                                                             above.sizes[dim])
```
SimonHeybrock: What if someone puts a mask on a monitor?

src/ess/sans/to_I_of_Q.py
SimonHeybrock: I agree.

src/ess/sans/normalization.py
```diff
@@ -0,0 +1,98 @@
...
+    function to perform the division.
+    The denominator cannot contain event data, it must be histogrammed data.
+
+    :param numerator: The data whose counts will be divided by the denominator. This
+        can either be event or dense (histogrammed) data.
+    :param denominator: The divisor for the normalization operation. This cannot be
+        event data, it must contain histogrammed data.
+    :param dim: In the case of a numerator containing event data, this is the dimension
+        along which the lookup operation should be performed. This can be omitted in
+        case the denominator has only one dimension.
```
SimonHeybrock: I appreciate the idea to keep this flexible, but is this ever not `Q` in SANS?

src/ess/sans/to_I_of_Q.py
nvaytet: This is now making these functions public, and we are illustrating their use quite prominently in a notebook.
I just made these names up as I went along, but as soon as we publish that notebook, we have to keep those names.
Should we have a think if we can come up with better names?
@wpotrzebowski your input would be useful here also.

src/ess/sans/normalization.py
nvaytet: ok, I hard-coded the `Q` in the function body instead.

docs/instruments/loki/sans2d_to_I_of_Q.ipynb
```diff
@@ -0,0 +1,712 @@
...
+   "execution_count": null,
+   "id": "37664f8c-0004-4ecb-876b-52fc2a010c43",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "mask_tof_min = sc.scalar(13000.0, unit='us')\n",
+    "mask_tof_max = sc.scalar(15750.0, unit='us')\n",
+    "tof_masked_region = sc.concat([sample.coords['tof']['tof', 0],\n",
+    "                               mask_tof_min, mask_tof_max,\n",
+    "                               sample.coords['tof']['tof', -1]], dim='tof')\n",
```
SimonHeybrock, line 267: This assumes event data. Maybe mention this on top of the notebook?

docs/instruments/loki/sans2d_reduction.ipynb
```diff
@@ -0,0 +1,415 @@
...
+   "metadata": {},
+   "source": [
+    "## Masking\n",
+    "\n",
+    "The next step is to mask noisy and saturated pixels,\n",
+    "as well as a time-of-flight range that contains spurious artifacts from the beamline components.\n",
+    "\n",
+    "**Note:** We use programatic masks here and not those stored in xml files.\n",
+    "\n",
+    "# Mask bad pixels\n",
```
SimonHeybrock:
```suggestion
    "### Mask bad pixels\n",
```
?

docs/instruments/loki/sans2d_reduction.ipynb
```diff
@@ -0,0 +1,415 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "f3715d62-e58b-410a-97cd-cfc16bc32a28",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "result = sample_q.bins.sum() - background_q.bins.sum()\n",
```
SimonHeybrock, line 351: Two notes here:
- Instead of `sum()` once could use `histogram()` here to define different Q bins.
- This can be done in event mode, see https://scipp.github.io/user-guide/binned-data/computation.html#Subtraction.

docs/instruments/loki/sans2d_reduction.ipynb
```diff
@@ -0,0 +1,415 @@
...
+   "execution_count": null,
+   "id": "dba9fe02-5c65-4142-b94b-ccca2df879e2",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "background_q = sans.to_I_of_Q(data=binned['background'],\n",
+    "    data_incident_monitor=binned['background'].attrs[\"monitor2\"].value,\n",
+    "    data_transmission_monitor=binned['background'].attrs[\"monitor4\"].value,\n",
+    "    direct_incident_monitor=binned['direct'].attrs[\"monitor2\"].value,\n",
+    "    direct_transmission_monitor=binned['direct'].attrs[\"monitor4\"].value,\n",
```
SimonHeybrock: This is repeated many times. How about changing the functions to take a dict of monitors as input?

SimonHeybrock (commented):
A few extra comments from reviewing the notebooks:

docs/instruments/loki/sans2d_reduction.ipynb
nvaytet: Then we would be relying on special keys inside the `to_I_of_Q` function?

docs/instruments/loki/sans2d_reduction.ipynb
nvaytet: I thought I would keep the same binning as was requested at the top of the notebook?

docs/instruments/loki/sans2d_reduction.ipynb
SimonHeybrock: Well, you are anyway relying on special function parameter names?

docs/instruments/loki/sans2d_reduction.ipynb
SimonHeybrock: Possibly, but there may be performance advantages to use a coarse Q binning when the computing I(Q) numerator. Not saying the change this now, but you may put a note, maybe?

docs/instruments/loki/sans2d_reduction.ipynb
nvaytet: I added dicts, and used recursion in some functions to make them accept dict for convenience.
I think it works quite nicely.

src/ess/sans/i_of_q.py
```diff
@@ -64,49 +110,56 @@ def resample_direct_beam(direct_beam: sc.DataArray,
...
-        data = data.copy(deep=False)
-        data.coords["gravity"] = gravity_vector()
-
-    if data.bins is not None:
-        out = _convert_events_to_q_and_merge_spectra(data=data,
-                                                     graph=graph,
-                                                     q_bins=q_bins,
-                                                     wavelength_bands=wavelength_bands)
+    if isinstance(data, dict):
+        return {
```
SimonHeybrock: Note sure about having this recursion everywhere?

SimonHeybrock (approved):

src/ess/sans/i_of_q.py
nvaytet: ok i'll remove it here

SimonHeybrock: I meant everywhere. Can this be solved in a more Pythonic way, without having to handle dicts in every function?

nvaytet:
@SimonHeybrock
Wojciech says to wait before merging, as he needs to be sure that we can have the data publicly available online...

wpotrzebowski:
It is fine to use the data. I will provide reference to the publication.

wpotrzebowski:
Reduced data has been published in https://aip.scitation.org/doi/full/10.1063/5.0059238 (it is one of data sets in Fig5d - hd-DES_10_h-C16EO8). It should be sufficent to reference the paper

nvaytet:
@wpotrzebowski can you have a quick look to see if the way I added the citation is ok?
Thanks

PR #61, opened 2021-11-08
Reflectometry rewrite

nvaytet:
This is a major re-write of the reflectometry code. This aims to be a first step in the rewrite, where some functionality has been removed/disabled (see illumination correction fx) for simplicity, and will be added back in as part of follow-up PRs.

The idea is to write the reflectometry module in a more functional way, rather than an object-oriented approach.
This way, we are passing Scipp data arrays and variables around, instead of custom classes, which makes functions more re-useable/portable, as well as more in-line with the rest of the `ess` codebase.

In addition, the gravity correction calculation has been replaced by a custom conversion graph that is then supplied to `transform_coords` to perform the calculation of the `two_theta` scattering angle and the `Q` vector.

The illumination correction has been removed from the reduction workflow (for now), but not from the codebase.

Many tests have been deleted, trying to keep relevant tests for the functions (the transform_coords is being tested in scipp and scippneutron, so we don't need to test again all aspects of it here). Some of these tests will come back in future PRs, when we add the full functionality back in.

A basic reduction notebook, based on the ones from the `ess-notebooks` repo has been added, and this should in the future become the reference reduction workflow, once we add all the missing bits.


src/ess/amor/tools.py
```diff
@@ -1,33 +1,26 @@
...
-def q_grid(q_min=0.008 * sc.Unit('1/angstrom'),
-           q_fix=0.005 * sc.Unit('1/angstrom'),
-           q_max=0.08 * sc.Unit('1/angstrom'),
-           d_q=None):
+def q_grid(q_min: sc.Variable = 0.008 * sc.Unit('1/angstrom'),
+           q_fix: sc.Variable = 0.005 * sc.Unit('1/angstrom'),
+           q_max: sc.Variable = 0.08 * sc.Unit('1/angstrom'),
+           d_q: sc.Variable = None) -> sc.Variable:
     """
-    Obtain a q_grid based on the custom linear-log grid from Jochen Stahn at PSI. The units of the bins are reciprocal angstrom.
```
arm61, line 18: Can we keep some comment (even if not in a dock string) reflecting where this grid has come from?

src/ess/reflectometry/conversions.py
```diff
@@ -0,0 +1,49 @@
...
+    c = (4 * pi).astype(dtype)
+    return c * sc.sin(two_theta.astype(dtype, copy=False)) / wavelength
+
+
+def reflectometry_graph() -> dict:
+    """
+    Generate a coordinate transformation graph for reflectometry.
+    """
+    graph = {**conversions.beamline(scatter=True), **conversions.elastic("tof")}
+    graph["two_theta"] = reflectometry_theta
```
arm61: Can we also produce something called `'theta'` alongside this?

arm61 (commented):
~~How come the `write.py` is removed? Just cause it needs to be rewritten from scratch?~~

Ignore I have remembered our conversation. i.e. the writing will be done with `orsopy`.

src/ess/reflectometry/conversions.py
nvaytet: yes, it seems we can add something that is called `theta` which is just a renaming of `two_theta` by doing
```Py
    graph["theta"] = "two_theta"
```

SimonHeybrock: Are you sure we need/want a double meaning of `theta`? This seems like it will nearly with certainty lead to bugs.

arm61: I want a dimension called `theta` in my final dataset, which is equal to a half of `two_theta`

SimonHeybrock: Then @nvaytet must have misunderstood you?

src/ess/amor/io/load.py
```diff
@@ -0,0 +1,90 @@
...
+    TODO: generalise mechanism to fold any number of pulses.
+    """
+    tof_offset = tau * chopper_phase / (180.0 * sc.units.deg)
+    # Make 2 bins, one for each pulse
+    edges = sc.array(dims=['tof'],
+                     values=[
+                         -tof_offset.value, (tau - tof_offset).value,
+                         (2 * tau - tof_offset).value
+                     ],
+                     unit=tau.unit)
```
SimonHeybrock:
```suggestion
    edges = sc.concat([-tof_offset, tau - tof_offset, 2 * tau - tof_offset], 'tof')
```

src/ess/amor/io/load.py
```diff
@@ -0,0 +1,90 @@
...
+    data = sc.bin(data, edges=[edges])
+    # Make one offset for each bin
+    offset = sc.concatenate(tof_offset, tof_offset - tau, 'tof')
```
SimonHeybrock:
```suggestion
    offset = sc.concat([tof_offset, tof_offset - tau], 'tof')
```

src/ess/amor/io/load.py
```diff
@@ -0,0 +1,90 @@
...
+    # Apply the offset on both bins
+    data.bins.coords['tof'] += offset
+    # Rebin to exclude second (empty) pulse range
+    return sc.bin(data, edges=[sc.concatenate(0. * sc.units.us, tau, 'tof')])
```
SimonHeybrock: concat

src/ess/amor/io/load.py
```diff
@@ -0,0 +1,90 @@
...
+    filename,
+    sample_angle_offset: sc.Variable = 0 * sc.units.deg,
+    beam_size: sc.Variable = 0.001 * sc.units.m,
+    sample_size: sc.Variable = 0.01 * sc.units.m,
+    detector_spatial_resolution: sc.Variable = 0.0025 * sc.units.m,
+    chopper_sample_distance: sc.Variable = 15.0 * sc.units.m,
+    chopper_speed: sc.Variable = 20 / 3 * 1e-6 / sc.units.us,
+    chopper_detector_distance: sc.Variable = 19.0 * sc.units.m,
+    chopper_chopper_distance: sc.Variable = 0.49 * sc.units.m,
+    chopper_phase: sc.Variable = -8.0 * sc.units.deg,
```
SimonHeybrock: Would it make sense to introduce a `dataclass` that combines all the chopper params into one?

src/ess/amor/io/load.py
```diff
@@ -0,0 +1,90 @@
...
+    data = refl.io.load(filename=filename,
+                        sample_angle_offset=sample_angle_offset,
+                        gravity=gravity,
+                        beam_size=beam_size,
+                        sample_size=sample_size,
+                        detector_spatial_resolution=detector_spatial_resolution)
+
+    # Convert tof nanoseconds to microseconds for convenience
+    # TODO: is it safe to assume that the dtype of the binned wrapper coordinate is
+    # the same as the dtype of the underlying event coordinate?
```
SimonHeybrock: Scipp does not impose any restrictions here, so it may in principle differ. It depends on what the loader does.

src/ess/amor/io/load.py
```diff
@@ -0,0 +1,90 @@
...
+    if data.coords['tof'].dtype != sc.dtype.float64:
+        data.bins.coords['tof'] = data.bins.coords['tof'].astype('float64')
+        data.coords['tof'] = data.coords['tof'].astype('float64')
```
SimonHeybrock:
```suggestion
    data.bins.coords['tof'] = data.bins.coords['tof'].astype('float64', copy=False)
    data.coords['tof'] = data.coords['tof'].astype('float64', copy=False)
```

src/ess/amor/io/load.py
```diff
@@ -0,0 +1,90 @@
...
+    data.bins.coords['tof'] = sc.to_unit(data.bins.coords['tof'], 'us')
+    data.coords['tof'] = sc.to_unit(data.coords['tof'], 'us')
```
SimonHeybrock:
```suggestion
    data.bins.coords['tof'] = sc.to_unit(data.bins.coords['tof'], 'us', copy=False)
    data.coords['tof'] = sc.to_unit(data.coords['tof'], 'us', copy=False)
```

src/ess/amor/io/load.py
```diff
@@ -0,0 +1,90 @@
...
+
+    # These are Amor specific parameters
+    tau = 1 / (2 * chopper_speed)
+    # The source position is not the true source position due to the
+    # use of choppers to define the pulse.
+    data.coords["source_position"] = sc.geometry.position(0.0 * sc.units.m,
+                                                          0.0 * sc.units.m,
+                                                          -chopper_sample_distance)
+    data = _tof_correction(data, tau, chopper_phase)
+    return data
```
SimonHeybrock:
```suggestion
    return _tof_correction(data, tau, chopper_phase)
```

src/ess/amor/tools.py
```diff
@@ -1,33 +1,26 @@
...
+        Default is `0.008 Å^{-1}`.
+    :param q_fix: The last point in the linear binning and first in the log binning.
+        Default is `0.005 Å^{-1}`.
+    :param q_max: The maximum q-value to be present in the data.
+        Default is `0.08 Å^{-1}`.
+    :param q_d: The q-spacing for the linear region, from which the log-spacing is
+        defined. Default is `0.05*q_fix Å^{-1}`.
     """
     q_min = sc.to_unit(q_min, q_max.unit)
     q_fix = sc.to_unit(q_fix, q_max.unit)
```
SimonHeybrock: Regarding code below: May be worth checking if a simple combination of `sc.linspace` and `sc.geomspace` or `sc.logspace` can do the same, in a more transparent way. Like this nobody can tell what is going on.

src/ess/reflectometry/conversions.py
```diff
@@ -0,0 +1,49 @@
...
+                        scattered_beam: sc.Variable) -> sc.Variable:
+    """
+    Compute the reflectometry theta angle, including gravity correction.
+    It is similar to the classical two_theta in other techniques,
+    but we also neglect the x component of the scattered beam.
+    """
+    scattered_beam = scattered_beam.copy()
+    scattered_beam.fields.x *= 0
+    # Arbitrary internal convention: beam=z, gravity=y
+    grav = sc.norm(gravity)
```
SimonHeybrock: This comment does not appear to be correct?

- We are changing `fields.x`, so no *internal* convention here?
- It uses input gravity direction.

Can we instead define a sample-normal vector, which, together with the incident beam, might define which component should be set to 0?

May also consider introduced a pre-processing step in the graph which converts `scattered_beam` into the scattered beam you actually need. Or, even better(?), do not use `scattered_beam` at all, since apparently the point-like sample definition it uses is not what you need here?

src/ess/reflectometry/conversions.py
```diff
@@ -0,0 +1,49 @@
...
+    wavelength = sc.to_unit(wavelength, "m", copy=False)
+    drop = grav * m_n**2 / (2 * h**2) * wavelength**2 * L2**2
+    return sc.asin(sc.sqrt(x**2 + (y + drop)**2) / L2)
+
+
+def reflectometry_q(wavelength: sc.Variable, two_theta: sc.Variable) -> sc.Variable:
+    """
+    The reflectometry theta angle corresponds to what is usually called theta in other
+    techniques. Hence, the reflectometry q calculation overrides the default by
+    taking `sin(two_theta)` instead of `sin(two_theta/2)`.
```
SimonHeybrock: This contradicts what @arm61 said earlier in the comment thread below.

src/ess/reflectometry/io/load.py
```diff
@@ -0,0 +1,45 @@
...
+import scipp as sc
+import scippneutron as scn
+import warnings
+
+
+def load(filename: str,
+         sample_angle_offset: sc.Variable = 0 * sc.units.deg,
+         beam_size: sc.Variable = 0.001 * sc.units.m,
+         sample_size: sc.Variable = 0.01 * sc.units.m,
+         detector_spatial_resolution: sc.Variable = 0.0025 * sc.units.m,
```
SimonHeybrock: Unless this values are good for *all* ESS reflectometres, this may not be a good place to set them?

src/ess/reflectometry/io/load.py
```diff
@@ -0,0 +1,45 @@
...
+         gravity: sc.Variable = sc.vector(value=[0, -1, 0]) * sc.constants.g,
```
SimonHeybrock: If there is an ESS standard for beamline reference frames it may be worth defining this in the ESS module, something like `ess.gravity`

SimonHeybrock (commented):
Did not look at notebooks or tests. Some comments  and questions below:

src/ess/reflectometry/conversions.py
arm61: I think so. For now, this is fine, I will open an issue for this to be included.

src/ess/amor/io/load.py
arm61: I like this idea. `dataclass` is a fun addition to Python

src/ess/reflectometry/conversions.py
SimonHeybrock: I think I am very wary of using theta because I have things going wrong with that in the past. If *think* reflectometry may be an exception, since a mixup with `\theta_{spherical}` is not likely (unlike other techniques, which define `z` along `incident_beam`.

src/ess/reflectometry/conversions.py
arm61: Let me play with the code today to try and work out the specific nomenclature on this.

src/ess/reflectometry/io/load.py
arm61: The problem on this is that ESS with have two reflectometers. For ESTIA gravity will be defined as above, for FREIA (and Amor) it will be `sc.vector(value=[0, 0, -1]) * sc.constants.g`. This is because reflectometry has a specific definition of what is x, y, and z (see https://orsopy.readthedocs.io/en/latest/orsopy.fileio.base.html#orsopy.fileio.base.ValueVector)

SimonHeybrock: Can you double check if this is "allowed"? I vaguely remember being in a meeting a couple years ago, where there was talk about standardized reference frame definitions for ESS beamlines.

arm61: A standard reference frame is fine for raw data, etc. but the reflectometry community have historically defined the scattering vector in the z-dimension, and I know (from discussion with the Amor instrument scientist) that if this isn't respected we are gonna have a lot of trouble on our hands.

src/ess/reflectometry/conversions.py
```diff
@@ -0,0 +1,49 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2021 Scipp contributors (https://github.com/scipp)
+import scipp as sc
+from scipp.constants import m_n, h, pi
+from scippneutron.tof import conversions
+from scippneutron.core.conversions import _elem_dtype
+
+
+def reflectometry_theta(gravity: sc.Variable, wavelength: sc.Variable,
```
arm61:
```suggestion
def gamma(gravity: sc.Variable, wavelength: sc.Variable,
```

src/ess/reflectometry/conversions.py
```diff
@@ -0,0 +1,49 @@
...
+                        incident_beam: sc.Variable,
+                        scattered_beam: sc.Variable) -> sc.Variable:
+    """
+    Compute the reflectometry theta angle, including gravity correction.
+    It is similar to the classical two_theta in other techniques,
+    but we also neglect the x component of the scattered beam.
```
arm61:
```suggestion
    Compute the gamma angle, including gravity correction.
    It is similar to the classical two_theta in other techniques,
    but we also neglect the x component of the scattered beam.
    See the schematic in Fig 5 of doi: 10.1016/j.nima.2016.03.007.
```

src/ess/reflectometry/conversions.py
```diff
@@ -0,0 +1,49 @@
...
+    grav = sc.norm(gravity)
+    L2 = sc.norm(scattered_beam)
+    y = sc.dot(scattered_beam, gravity) / grav
+    n = sc.cross(incident_beam, gravity)
+    n /= sc.norm(n)
+    x = sc.dot(scattered_beam, n)
+    wavelength = sc.to_unit(wavelength, "m", copy=False)
+    drop = grav * m_n**2 / (2 * h**2) * wavelength**2 * L2**2
+    return sc.asin(sc.sqrt(x**2 + (y + drop)**2) / L2)
+
```
arm61, line 38:
```suggestion

def theta(gamma: sc.Variable, omega: sc.Variable) -> sc.Variable:
    """
    Determine the value of theta from the gamma and omega (the sample angle offset) angles.
    """
    return gamma - omega
    
```

src/ess/reflectometry/conversions.py
arm61:
```suggestion
    graph["gamma"] = gamma
    graph["theta"] = theta
```

arm61: I have added suggestions based on the nomenclature used in the Amor paper (fig 5 [https://doi.org/10.1016/j.nima.2016.03.007](10.1016/j.nima.2016.03.007)).

src/ess/reflectometry/io/load.py
```diff
@@ -0,0 +1,45 @@
...
+# Copyright (c) 2021 Scipp contributors (https://github.com/scipp)
+# @author Andrew R. McCluskey (arm61)
+
+import scipp as sc
+import scippneutron as scn
+import warnings
+
+
+def load(filename: str,
+         sample_angle_offset: sc.Variable = 0 * sc.units.deg,
```
arm61:
```suggestion
         omega: sc.Variable = 0 * sc.units.deg,
```

src/ess/reflectometry/io/load.py
```diff
@@ -0,0 +1,45 @@
...
+        Default is `True`.
+    """
+    if disable_warnings:
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore", category=UserWarning)
+            da = scn.load_nexus(filename)
+    else:
+        da = scn.load_nexus(filename)
+
+    da.coords["sample_angle_offset"] = sample_angle_offset
```
arm61:
```suggestion
    da.coords["omega"] = omega
```

src/ess/reflectometry/io/load.py
```diff
@@ -0,0 +1,45 @@
...
+         beam_size: sc.Variable = 0.001 * sc.units.m,
+         sample_size: sc.Variable = 0.01 * sc.units.m,
+         detector_spatial_resolution: sc.Variable = 0.0025 * sc.units.m,
+         gravity: sc.Variable = sc.vector(value=[0, -1, 0]) * sc.constants.g,
+         disable_warnings: bool = True) -> sc.DataArray:
+    """
+    The general reflectometry data loader.
+
+    :param filename: The path to the file to be reduced.
+    :param sample_angle_offset: Correction for omega or possibly misalignment of sample.
```
arm61:
```suggestion
    :param omega: Correction for omega or possibly misalignment of sample.
```

src/ess/reflectometry/conversions.py
SimonHeybrock: So `gamma = 2 theta` (apart from the x component thing)?

arm61: I guess so, the naming convention comes from the Amor paper. I expect the rational from the Amor instrument scientist was that 2 theta is a more common thing for transmission methods and he was wanting to ensure clarity.

src/ess/amor/io/load.py
```diff
@@ -0,0 +1,84 @@
...
+    offset = sc.concat([tof_offset, tof_offset - tau], 'tof')
+    # Apply the offset on both bins
+    data.bins.coords['tof'] += offset
+    # Rebin to exclude second (empty) pulse range
+    return sc.bin(data, edges=[sc.concatenate(0. * sc.units.us, tau, 'tof')])
+
+
+def load(
+    filename,
+    sample_angle_offset: sc.Variable = 0 * sc.units.deg,
```
arm61:
```suggestion
    omega: sc.Variable = 0 * sc.units.deg,
```

src/ess/amor/io/load.py
```diff
@@ -0,0 +1,84 @@
...
+    chopper_speed: sc.Variable = 20 / 3 * 1e-6 / sc.units.us,
+    chopper_detector_distance: sc.Variable = 19.0 * sc.units.m,
+    chopper_chopper_distance: sc.Variable = 0.49 * sc.units.m,
+    chopper_phase: sc.Variable = -8.0 * sc.units.deg,
+    gravity: sc.Variable = sc.vector(value=[0, -1, 0]) * sc.constants.g
+) -> sc.DataArray:
+    """
+    Loader for a single Amor data file.
+
+    :param sample_angle_offset: Correction for omega or possibly misalignment of sample.
```
arm61:
```suggestion
    :param omega: Correction for omega or possibly misalignment of sample.
```

src/ess/amor/io/load.py
```diff
@@ -0,0 +1,84 @@
...
+        Default is `6.6666... e-6 µs^{-1}`.
+    :param chopper_detector_distance: Distance from chopper to detector.
+        Default is `19 m`.
+    :param chopper_chopper_distance: The distance between the wavelength defining
+        choppers. Default `0.49 m`.
+    :param chopper_phase: Phase offset between chopper pulse and ToF zero.
+        Default is `-8. degrees of arc`.
+    """
+    data = refl.io.load(filename=filename,
+                        sample_angle_offset=sample_angle_offset,
```
arm61:
```suggestion
                        omega=omega,
```

src/ess/amor/io/load.py
nvaytet: The thing is that they are not all just chopper parameters. The `chopper_detector_distance` involves the detector, the `chopper_sample_distance` involves the sample. For chopper only things, we would only have `chopper_speed` and `chopper_phase` left.
Maybe we should change the `chopper_detector_distance` to be computed from a `chopper_position` and a `sample_position`? etc...

nvaytet: We could also use the Chopper class in the WFM code?
https://github.com/scipp/ess/blob/main/src/ess/wfm/choppers.py#L8

SimonHeybrock: > The thing is that they are not all just chopper parameters

That may strictly speaking be true, but I don't see this as a real problem? Also, you *can* actually argue that the distance of a chopper from "other things" *is* a parameter of the chopper, and a very important one.

SimonHeybrock: > We could also use the Chopper class in the WFM code? https://github.com/scipp/ess/blob/main/src/ess/wfm/choppers.py#L8

Feels like overkill, but maybe you have some future change in mind I am not aware of?

src/ess/amor/tools.py
```diff
@@ -1,67 +1,44 @@
...
-            unit=q_max.unit)
-        return q_log
-    else:
-        raise ValueError("The given values of input parameters are not compatible.")
+    if not isinstance(scale, list):
+        scale = [scale]
+    if not isinstance(num, list):
+        num = [num]
+    if len(scale) != len(edges) - 1:
+        raise ValueError("Sizes do not match")
```
arm61: Can this error message be more specific?
```suggestion
        raise ValueError("Length of scale should be one greater than edges.")
```

src/ess/reflectometry/conversions.py
```diff
@@ -0,0 +1,56 @@
...
+import scipp as sc
+from scipp.constants import m_n, h, pi
+from scippneutron.tof import conversions
+from scippneutron.core.conversions import _elem_dtype
+
+
+def incident_beam(source_chopper, sample_position):
+    """
+    Compute the incident beam vector from the source chopper position vector,
+    instead of the source_position vector.
```
arm61: This is only true for Amor instrument not for all reflectometers (Amor is a ToF reflectometer where the ToF is defined from the chopper, cause PSI is a constant source).

src/ess/reflectometry/conversions.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+    return sample_position - source_chopper.value.position
+
+
+def theta(gravity: sc.Variable, wavelength: sc.Variable, incident_beam: sc.Variable,
+          scattered_beam: sc.Variable, sample_rotation: sc.Variable) -> sc.Variable:
```
arm61, line 10: `sample_rotation` not `omega`?

arm61 (commented):
Sorry if I am jumping the gun on this review

src/ess/amor/tools.py
nvaytet: Isn't it the other way round? `edges` should be one greater?

src/ess/reflectometry/conversions.py
nvaytet: So we should make a conversion graph for Amor, i guess

src/ess/reflectometry/conversions.py
nvaytet: I feel omega is very non-descript if you don't have the figure...

src/ess/reflectometry/conversions.py
nvaytet: Ok i made a new graph for amor that imports the more generic reflectometry graph

nvaytet:
I also made an `amor.instrument_view` which adds the sample and source chopper as beamline components, and sets the pixel size.

src/ess/reflectometry/conversions.py
arm61: But the figure reference is in the doctstring.

src/ess/amor/tools.py
nvaytet: I made a new implementation which can be used in the following way:

Linear mesh:
```Py
q_grid(dim='qz',
                         edges=[0.008, 0.08],
                         scale='linear',
                         num=50,
                         unit='1/angstrom')
```

Log mesh:
```Py
q_grid(dim='qz',
                         edges=[0.008, 0.08],
                         unit='1/angstrom',
                         scale='log',
                         num=50)
```

Linear-log mesh:
```Py
q_grid(dim='qz',
                            edges=[0.008, 0.03, 0.08],
                            unit='1/angstrom',
                            scale=['linear', 'log'],
                            num=[16, 20])
```

Linear-log-linear mesh:
```Py
q_grid(dim='qz',
                               edges=[0.008, 0.03, 0.08, 0.12],
                               unit='1/angstrom',
                               scale=['linear', 'log', 'linear'],
                               num=[16, 20, 10])
```
Please have a look and tell me what you think.

src/ess/reflectometry/io/load.py
nvaytet: I made an Amor beamline.
The `reflectometry.io.load` is gone for now. Only `amor.load()` remains.

src/ess/reflectometry/conversions.py
nvaytet: Sure but the docstring is not immediately visible to the user in the notebook.

arm61 (approved):

src/ess/amor/beamline.py
```diff
@@ -0,0 +1,49 @@
...
+def make_beamline(
+    sample_rotation: sc.Variable = None,
+    beam_size: sc.Variable = 0.001 * sc.units.m,
+    sample_size: sc.Variable = 0.01 * sc.units.m,
+    detector_spatial_resolution: sc.Variable = 0.0025 * sc.units.m,
+    gravity: sc.Variable = sc.vector(value=[0, -1, 0]) * g,
+    chopper_frequency: sc.Variable = sc.scalar(20 / 3, unit='Hz'),
+    chopper_phase: sc.Variable = sc.scalar(-8.0, unit='deg'),
+    chopper_position: sc.Variable = sc.vector(value=[0, 0, -15.0], unit='m')
+) -> dict:
```
SimonHeybrock, line 17: Have you considered using a `dataclass`?

src/ess/amor/conversions.py
```diff
@@ -0,0 +1,22 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2021 Scipp contributors (https://github.com/scipp)
+import scipp as sc
+from ..reflectometry.conversions import reflectometry_graph
+
+
+def incident_beam(source_chopper: sc.Variable,
```
SimonHeybrock: I don't know if anyone will use this function directly (instead of just via a transform graph), but maybe make the arguments keyword-only, since this one is easy to mess up?

src/ess/amor/conversions.py
```diff
@@ -0,0 +1,22 @@
...
+    instead of the source_position vector.
+    """
+    return sample_position - source_chopper.value.position
+
+
+def specular_reflection_graph() -> dict:
+    """
+    Generate a coordinate transformation graph for Amor reflectometry.
+    """
+    graph = reflectometry_graph()
```
SimonHeybrock: In `scippneutron` the graphs do not have a `_graph` suffix, should we be consistent?

src/ess/amor/instrument_view.py
```diff
@@ -0,0 +1,39 @@
...
+                'size': sc.vector(value=[0.2, 0.01, 0.2], unit=sc.units.m),
+                'type': 'box'
+            },
+            "source_chopper": {
+                'center': da.meta['source_chopper'].value.position,
+                'color': 'blue',
+                'size': sc.vector(value=[0.5, 0, 0], unit=sc.units.m),
+                'type': 'disk'
+            }
+        }
```
SimonHeybrock: This feels a bit unusual: If we provide `components`, what is the point of this function, since it is not merging with the pre-defined ones? In that case the user could just use `sc.instrument_view` directly? Did you intend to provide dict-merging instead?

Can you move the setup of the component dict to a separate function? Then this function here becomes generic or trivial, something like:

```python
def instrument_view(da, **kwargs):
   return ess.instrument_view(da, amor.beamline.components(), **kwargs)
```

where `ess.instrument_view` could deal with dict merging, and `amor.beamline.components()` returns the components (feel free to change names).

src/ess/amor/beamline.py
```diff
@@ -0,0 +1,49 @@
...
+        'sample_rotation': sample_rotation,
+        'beam_size': beam_size,
+        'sample_size': sample_size,
+        'detector_spatial_resolution': detector_spatial_resolution,
+        'gravity': gravity
+    }
+    beamline["source_chopper"] = sc.scalar(
+        Chopper(frequency=chopper_frequency,
+                phase=chopper_phase,
+                position=chopper_position))
```
SimonHeybrock: We should very carefully consider whether we want to start storing arbitrary Python objects. This comes with a number of limitations:

- Cannot be saved to a file (at least for now).
- Invokes the GIL in many operations.
- It don't know if string-formatting or other things work nicely.

src/ess/amor/tools.py
```diff
@@ -1,67 +1,48 @@
...
-    :param q_min: The minimum q-value to be present in the data. Optional, defaults to 0.008 Å^{-1}
-    :type q_min: scipp._scipp.core.Variable
-    :param q_fix: The last point in the linear binning and first in the log binning. Optional, defaults to 0.005 Å^{-1}
-    :type: q_fix: scipp._scipp.core.Variable
-    :param q_max: The maximum q-value to be present in the data. Optional, defaults to 0.08 Å^{-1}
-    :type q_max: scipp._scipp.core.Variable
-    :param q_d: The q-spacing for the linear region, from which the log-spacing is defined. Optional, defaults to 0.05*q_fix Å^{-1}
-    :type q_d: scipp._scipp.core.Variable
+    Generate bin edges for a `q_grid` inspired by the custom linear-log grid from
+    Jochen Stahn at PSI.
```
SimonHeybrock: This does not explain much. The function name indicates that this is a (2d) grid (at least personally I think if something 2d or higher when I read "grid")? But it does not look like that below, and is instead making a 1d range with linear and log sections? I feel that both function name and comment may benefit from renaming/clarification?

src/ess/reflectometry/conversions.py
```diff
@@ -0,0 +1,47 @@
...
+from scippneutron.tof import conversions
+from scippneutron.core.conversions import _elem_dtype
+
+
+def theta(gravity: sc.Variable, wavelength: sc.Variable, incident_beam: sc.Variable,
+          scattered_beam: sc.Variable, sample_rotation: sc.Variable) -> sc.Variable:
+    """
+    Compute the theta angle, including gravity correction,
+    This is similar to the theta calculation in SANS (see
+    https://docs.mantidproject.org/v3.9.0/algorithms/Q1D-v2.html#algm-q1d), but we
```
SimonHeybrock: Very old docs?

src/ess/reflectometry/conversions.py
```diff
@@ -0,0 +1,47 @@
...
+    ignore the horizontal `x` component.
+    See the schematic in Fig 5 of doi: 10.1016/j.nima.2016.03.007.
+    """
+    grav = sc.norm(gravity)
+    L2 = sc.norm(scattered_beam)
+    y = sc.dot(scattered_beam, gravity) / grav
+    wavelength = sc.to_unit(wavelength, "m", copy=False)
+    drop = grav * m_n**2 / (2 * h**2) * wavelength**2 * L2**2
```
SimonHeybrock: I encourage you to count how many operations this does on every event, and then strategically place parentheses, change order of operations, or do operations in-place (also in the line below). Otherwise you can easily make coord transforms 2x or 3x times slower.

src/ess/reflectometry/conversions.py
```diff
@@ -0,0 +1,47 @@
...
+    return sc.asin(sc.abs(y + drop) / L2) - sc.to_unit(sample_rotation, 'rad')
```
SimonHeybrock: Where is the factor 1/2 that is present in the SANS equation?

src/ess/reflectometry/conversions.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def reflectometry_q(wavelength: sc.Variable, theta: sc.Variable) -> sc.Variable:
+    """
+    Compute the Q vector from the theta angle computed as the difference
+    between gamma and omega.
+    """
+    dtype = _elem_dtype(wavelength)
+    c = (4 * pi).astype(dtype)
+    return c * sc.sin(theta.astype(dtype, copy=False)) / wavelength
```
SimonHeybrock: This looks just like the "normal" Q, why does it need redefining here?

src/ess/reflectometry/conversions.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def reflectometry_graph() -> dict:
+    """
+    Generate a coordinate transformation graph for reflectometry.
+    """
+    graph = {**conversions.beamline(scatter=True), **conversions.elastic("tof")}
+    del graph['two_theta']
+    del graph['dspacing']
+    del graph['Q']
```
SimonHeybrock, line 59: See above, why not just keep original?

src/ess/reflectometry/corrections.py
```diff
@@ -1,101 +1,35 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2021 Scipp contributors (https://github.com/scipp)
-# flake8: noqa: E501
-"""
-Corrections to be used for neutron reflectometry reduction processes.
-"""
 import numpy as np
 import scipp as sc
 from scipy.special import erf
```
SimonHeybrock: `erf` is available in scipp.

src/ess/reflectometry/corrections.py
```diff
@@ -1,101 +1,35 @@
...
-        (:py:class:`scipp._scipp.core.Variable`): Correction factor.
+    :param beam_size: Width of incident beam.
+    :param sample_size: Width of sample in the dimension of the beam.
+    :param theta: Incident angle.
     """
     beam_on_sample = beam_size / sc.sin(theta)
     fwhm_to_std = 2 * np.sqrt(2 * np.log(2))
     scale_factor = erf((sample_size / beam_on_sample * fwhm_to_std).values)
-    return sc.Variable(values=scale_factor, dims=theta.dims)
+    return sc.array(values=scale_factor, dims=theta.dims)
```
SimonHeybrock: See above, use `sc.erf`

src/ess/reflectometry/resolution.py
```diff
@@ -30,16 +25,12 @@ def detector_resolution(spatial_resolution, pixel_position, sample_position):
...
-        offset_value (:py:class:`scipp._scipp.core.Variable`): z-dimension offset value.
+    Compute new positions that have been offset in the z-dimension.
 
-    Returns:
-        (:py:class:`scipp._scipp.core.Variable`): New position variables that has been offset.
+    :param position: Position vectors (dtype `vector_3_float64`).
+    :param offset_value: Offset to be applied along the z-dimension.
     """
     return sc.geometry.position(position.fields.x, position.fields.y,
                                 position.fields.z + offset_value)
```
SimonHeybrock:
```python
position = position.copy()
position.fields.z += offset_value
```
would be a more modern way to write this. Why is the parameter named `offset_value` instead of just `offset`?

src/ess/reflectometry/conversions.py
nvaytet: The normal Q uses `two_theta / 2` while this uses `theta` directly.

SimonHeybrock: Can you explain that in a comment? Otherwise the next reader will have the same question.

src/ess/amor/beamline.py
nvaytet: Since this is basically making a dict of coordinates (that are to be added to the data), I thought returning a dict was more appropriate?

SimonHeybrock: If the only purpose is to add coords, then yes. If not, note that there is `dataclasses.asdict(obj) `

src/ess/amor/instrument_view.py
nvaytet: I made it this way in case someone wanted to not show the components.
I guess I could do this by adding a kwarg `hide_components` in the suggested `ess.instrument_view` that would basically clear the dict of components if `True`. But it kind of feels like a strange way of doing things.

Also, what would the names of the arguments of the `ess.instrument_view` function? `default_components` and `components` and merge those two together?

I would say dict merging should either be done manually, or in the `amor.beamline.components()` function. And then there is no need for `ess.instrument_view`?

SimonHeybrock: > I would say dict merging should either be done manually

Agree, but at least we need to provide a function to get the default dict.

src/ess/reflectometry/conversions.py
nvaytet: It's actually a bit different from the SANS calculation. It comes from the fact that we are assuming all reflections are specular.
In the figure
fig
the centre of the detector is at an angle `gamma_d = 2*omega`, but the detector has a physical extent, and it measures counts not only at `2*omega`, but also other angles.

If you consider the bottom part of the yellow area at the detector (on the right), the only way neutrons could have arrived here via specular reflection, is if they arrived from the bottom edge of the yellow beam area (on the left before the sample).

Hence, the `theta` angle for these neutrons are the `gamma` angle (which is basically the same as our classical `two_theta` calculation that considers the angle between incident and scattered beam) minus the sample rotation `omega`.

So we don't want the factor of 1/2 because we want classical `two_theta` minus `omega`.

I will update the docstring and also add a better explanation with a drawing in the notebook.

src/ess/amor/tools.py
nvaytet: `linlogspace`?

src/ess/amor/beamline.py
nvaytet: I changed the whole chopper thing to use a Dataset and free functions.

src/ess/amor/tools.py
```diff
@@ -5,14 +5,24 @@
...
+    spacings.
+
+    Examples:
+
+    - Create linearly spaced edges (equivalent to `sc.linspace`):
+        linlogspace(dim='x', edges=[0.008, 0.08], scale='linear', num=50, unit='m')
+    - Create logarithmically spaced edges (equivalent to `sc.geomspace`):
+        linlogspace(dim='x', edges=[0.008, 0.08], scale='log', num=50, unit='m')
+    - Create edges with a linear and a logarithmic part:
+        linlogspace(dim='x', edges=[1, 3, 8], scale=['linear', 'log', num=[16, 20])
```
SimonHeybrock:
```suggestion
        linlogspace(dim='x', edges=[1, 3, 8], scale=['linear', 'log'], num=[16, 20])
```
I guess?

src/ess/choppers/__init__.py
```diff
@@ -2,4 +2,4 @@
 # Copyright (c) 2021 Scipp contributors (https://github.com/scipp)
 # flake8: noqa
 
-from .chopper import Chopper, ChopperKind
+from .chopper import *
```
SimonHeybrock: Really?

src/ess/odin/beamline.py
```diff
@@ -15,106 +14,158 @@ def make_beamline() -> dict:
...
+                sc.array(dims=[dim],
+                         values=[91.93, 142.23, 189.40, 233.63, 275.10, 313.99],
+                         unit='deg'),
+                "cutout_angles_width":
+                sc.array(dims=[dim],
+                         values=[5.74, 8.98, 12.01, 14.85, 17.52, 20.02],
+                         unit='deg'),
+                "kind":
+                "wfm"
+            }))
```
SimonHeybrock: If this code is too cumbersome to write (instead of the old `Chopper` constructor), you can consider replacing `Chopper.__init__` by a free function, that returns a dataset. The additional advantage would be that you can then verify arguments, etc.

src/ess/reflectometry/conversions.py
```diff
@@ -11,31 +11,36 @@ def theta(gravity: sc.Variable, wavelength: sc.Variable, incident_beam: sc.Varia
...
+    but we ignore the horizontal `x` component.
     See the schematic in Fig 5 of doi: 10.1016/j.nima.2016.03.007.
     """
     grav = sc.norm(gravity)
     L2 = sc.norm(scattered_beam)
     y = sc.dot(scattered_beam, gravity) / grav
     wavelength = sc.to_unit(wavelength, "m", copy=False)
-    drop = grav * m_n**2 / (2 * h**2) * wavelength**2 * L2**2
-    return sc.asin(sc.abs(y + drop) / L2) - sc.to_unit(sample_rotation, 'rad')
+    drop = wavelength**2 * (L2**2)
```
SimonHeybrock: Since wavelength is basically never in `m`, consider using `copy=True` and then do things in-place? Not sure if that is better, you'd have to try.

SimonHeybrock: `drop = wavelength**2 * (L2**2)` allocates a temporary. Have you considered:
```python
drop = wavelength**2
drop *= (L2**2)
```
?

src/ess/reflectometry/conversions.py
```diff
@@ -11,31 +11,36 @@ def theta(gravity: sc.Variable, wavelength: sc.Variable, incident_beam: sc.Varia
...
+    # Note: make use of in-place operations for better performance
+    drop *= grav * (m_n**2 / (2 * h**2))
```
SimonHeybrock, line 24: Can this factor not be combined with the `L2**2` above, to avoid one extra large operation?

src/ess/reflectometry/conversions.py
```diff
@@ -11,31 +11,36 @@ def theta(gravity: sc.Variable, wavelength: sc.Variable, incident_beam: sc.Varia
...
+    drop += y
+    return sc.asin(sc.abs(drop) / L2) - sc.to_unit(sample_rotation, 'rad')
```
SimonHeybrock:
```suggestion
    out = sc.abs(drop, out=drop)
    out /= L2
    out = sc.asin(out, out=out)
    out -=sc.to_unit(sample_rotation, 'rad')
    return out
```
does no extra allocation, vs. FOUR large allocations, I think?

src/ess/reflectometry/conversions.py
```diff
@@ -11,31 +11,36 @@ def theta(gravity: sc.Variable, wavelength: sc.Variable, incident_beam: sc.Varia
...
 
 
 def reflectometry_q(wavelength: sc.Variable, theta: sc.Variable) -> sc.Variable:
     """
     Compute the Q vector from the theta angle computed as the difference
     between gamma and omega.
+    Note that this is different from the 'normal' Q defined in scippneutron where
```
SimonHeybrock:
```suggestion
    Note that this is identical to the 'normal' Q defined in scippneutron where
```
...except that the input is given as theta instead of two_theta?

src/ess/v20/imaging/beamline.py
```diff
@@ -16,72 +15,110 @@ def make_beamline() -> dict:
...
-                values=np.array([79.78, 136.41, 191.73, 240.81, 287.13, 330.89]) + 15.0,
-                unit='deg'),
-            opening_angles_close=sc.array(
-                dims=[dim],
-                values=np.array([116.38, 172.47, 221.94, 267.69, 311.69, 360.0]) + 15.0,
-                unit='deg'),
-            kind=ChopperKind.FRAME_OVERLAP)
-    }
+    beamline["chopper_wfm_1"] = sc.scalar(
+        sc.Dataset(
```
SimonHeybrock: See above, a chopper "constructor" (even if it returns a dataset) would probably be useful?

src/ess/reflectometry/conversions.py
nvaytet: I did that first, but it failed with
```
~/code/ess/jupyter/ess/reflectometry/conversions.py in theta(gravity, wavelength, incident_beam, scattered_beam, sample_rotation)
     21     wavelength = sc.to_unit(wavelength, "m", copy=False)
     22     drop = wavelength**2
---> 23     drop *= (L2**2)
     24     # Note: make use of in-place operations for better performance
     25     drop *= grav * (m_n**2 / (2 * h**2))

NotFoundError: Expected (wavelength: 2) to include (detector_id: 9216).
```

SimonHeybrock: Ah, right. This is an interesting problem:
- The the dense coord we need a broadcast.
- For event coord we do not need a broadcast and would like to operate in place.

Can this be written nicely, or would we need to branch explicitly?

src/ess/choppers/__init__.py
nvaytet: Left over from debugging. Nice catch

src/ess/reflectometry/conversions.py
nvaytet: > Since wavelength is basically never in m, consider using copy=True and then do things in-place? Not sure if that is better, you'd have to try.

Should this actually be `wavelength = sc.to_unit(wavelength, L2.unit)`?

SimonHeybrock: Yes, probably. See also how this is done in scippneutron, e.g., https://github.com/scipp/scippneutron/blob/main/src/scippneutron/core/conversions.py#L65-L69.

nvaytet: > Ah, right. This is an interesting problem:
> - The the dense coord we need a broadcast.
> - For event coord we do not need a broadcast and would like to operate in place.
> Can this be written nicely, or would we need to branch explicitly?

I'm not sure exactly how to solve this. Can you elaborate a little on how you would operate in-place if you know you have an event coord?

SimonHeybrock: Conceptually:
```python
if needs_broadcast:
  a = a + b
else:
  a += b
```
That is, it doesn't really matter if events or not, the key thing is whether you need a broadcast or not.

src/ess/choppers/make_chopper.py
```diff
@@ -0,0 +1,60 @@
...
+    # Sanitize input parameters
+    if (None not in [cutout_angles_begin, cutout_angles_end
+                     ]) or (None not in [cutout_angles_center, cutout_angles_width]):
+        widths = utils.cutout_angles_width(chopper)
+        if (sc.min(widths) < sc.scalar(0.0, unit=widths.unit)).value:
+            raise ValueError("Negative window width found in chopper cutout angles.")
+        if not np.all(np.diff(utils.cutout_angles_begin(chopper).values) > 0):
+            raise ValueError("Chopper begin cutout angles are not monotonic.")
+        if not np.all(np.diff(utils.cutout_angles_end(chopper).values) > 0):
+            raise ValueError("Chopper end cutout angles are not monotonic.")
```
SimonHeybrock: Could this use https://scipp.github.io/generated/functions/scipp.issorted.html?

SimonHeybrock (approved):
Nice work!

src/ess/choppers/make_chopper.py
```diff
@@ -52,9 +52,9 @@ def make_chopper(frequency: sc.Variable,
         widths = utils.cutout_angles_width(chopper)
         if (sc.min(widths) < sc.scalar(0.0, unit=widths.unit)).value:
             raise ValueError("Negative window width found in chopper cutout angles.")
-        if not np.all(np.diff(utils.cutout_angles_begin(chopper).values) > 0):
+        if not sc.issorted(utils.cutout_angles_begin(chopper), dim=widths.dim).value:
```
SimonHeybrock: https://scipp.github.io/release/0.8.4/generated/functions/scipp.allsorted.html is your friend, forget that we have it.

nvaytet: ok, and just to make sure: do we also have a function that doesn't require the `dim`?

SimonHeybrock: I don't think we do, we have always avoided that (even though there are a couple of exceptions).

PR #78, opened 2022-01-03
Try github actions

nvaytet:


.github/workflows/weekly_and_release.yml
```diff
@@ -0,0 +1,15 @@
+name: Weekly and Release
+
+on:
+  release:
+    types: [published]
+  schedule:
+    - cron: "0 0 * * 0"
```
SimonHeybrock: In https://github.com/scipp/scipp/pull/2351 I am running on Thursday, i.e., after most of the work week, but with Friday to fix any potential problems.

.github/workflows/weekly_and_release.yml
```diff
@@ -0,0 +1,15 @@
...
+
+jobs:
+  build:
+    uses: scipp/ess/.github/workflows/build.yml@try_github_actions
+
+  deploy:
+    needs: build
+    uses: scipp/ess/.github/workflows/deploy.yml@try_github_actions
```
SimonHeybrock: Remember to remove branch name.

SimonHeybrock (commented):
Looks good, missing `.github/dependabot.yml`?

.github/workflows/weekly_and_release.yml
nvaytet: actually, is this not returning us to the situation we were in with the azure templates, that we constantly have to remember to change back the names of the branch to `main` after we have made changes to the workflow?

Maybe I should instead have a single workflow file with conditionals?

SimonHeybrock: I think it is fine: In practice you will rarely change both at the same time?

SimonHeybrock (approved):

.github/workflows/weekly_and_release.yml
nvaytet: Actually you are running on Thursday at 0:00, so you have Thursday and Friday to fix things.

Anyway, I think I will offset by a few hours or a day, so that we are not running everything at the same time. What do you think?

SimonHeybrock: I don't know, is there a limit of parallel worker *across* repos?

nvaytet: I think it's 20 workers for the whole organization: https://github.com/github/docs/discussions/1724

PR #87, opened 2022-01-18
Small post-release changes

nvaytet:
- remove unused `theme_overrides`
- add new section in release notes
- add missing entry in version dropdown menu


jl-wynen (approved):
Do we generally want apply those changes after a release? It makes sense to not show the empty release notes for the next version.

nvaytet:
> Do we generally want apply those changes after a release? It makes sense to not show the empty release notes for the next version.

Yeah i think that makes sense.
However, the entry in the version dropdown menu should go in when the release is made, not after. I just forgot about it, and we should make something like this script also for `ess`: https://github.com/scipp/scipp/pull/2380

PR #89, opened 2022-01-21
Fix typos and add neutron velocity equation in WFM intro notebook

### LINKED ISSUE #70: Small typos in WFM docs
- Small error in section "using a wfm chopper" the text refers to figure 3 where I guess it should be figure 4.
- Reading on in the same section: L1 and L2 are referenced but not defined or illustrated in a figure

nvaytet:
Fixes #70


jl-wynen (commented):
What about the second comment?
> Reading on in the same section: L1 and L2 are referenced but not defined or illustrated in a figure

Do you think it is explained sufficiently clearly?

nvaytet:
> What about the second comment?
> 
> > Reading on in the same section: L1 and L2 are referenced but not defined or illustrated in a figure
> 
> Do you think it is explained sufficiently clearly?

I think we are already saying that they are the primary and secondary flight paths, so I would say they are defined. I think anyone interested in neutron physics will know what those flight paths are and I think having a figure just for that is overkill.
So i decided to leave it out.

jl-wynen (approved):

PR #90, opened 2022-01-25
Add more files / folder to gitignore

jl-wynen:
Some of these were taken from scipp, others from my local installation.
- I used `build` for the docs. I'm open to pick a different directory if we want to have a standard location.
- The egg comes from installing `ess` in developer mode (`pip install -e .`)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,5 +13,6 @@ src/ess/_version.py
 
 # Caches
 .ipynb_checkpoints
+.mypy_cache
 .virtual_documents
 __pycache__
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,6 @@
 # Build artifacts
 build
 *.pyc
-_fixed_version.py
 docs/generated
 ess.egg-info
 src/ess/_version.py
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,17 @@
+# Build artifacts
+build
 *.pyc
-.idea
 _fixed_version.py
+docs/generated
+ess.egg-info
+src/ess/_version.py
+
+# IDE / Editor settings
+.idea
+.vscode
+.vs
+
+# Caches
+.ipynb_checkpoints
+.virtual_documents
+__pycache__


.gitignore
```diff
@@ -1,3 +1,17 @@
+# Build artifacts
+build
```
nvaytet, line 2: I think `build` for the docs is not optimal, as some people (including myself) actually have that folder in `docs/build`.
Not sure what to suggest.

.gitignore
```diff
@@ -1,3 +1,17 @@
...
 *.pyc
-.idea
 _fixed_version.py
```
nvaytet: I think this file is no longer being generated. Remove?

.gitignore
jl-wynen: I'd generally avoid placing files in the source tree, with the exception of build folders at the root level. But we already have `docs/generated`, so I would be fine with `docs/build`.

.gitignore
jl-wynen: Ok

.gitignore
jl-wynen: Actually, building in the docs sources is a bad idea because sphinx reads the outputs as inputs. This does not seem to break the output but it does mean that sphinx regenerates a lot of files because it misses its cache every time. See https://github.com/scipp/scipp/pull/1970

nvaytet: hmm that would explain why I always have to delete my build directory!
I still think it's a dangerous name to have in a gitignore, as it could refer to many things
Fx in the scipp dev docs, instructions are to build scipp (not the docs but the scipp library) in a `build` folder in the top level of the source tree.
That said, one probably does not want to have that under version control either. And we have no building in the `ess` repo.
So maybe it's ok to have the `build` folder here?

I personally don't really mind, as I don't use the `.gitignore` ;-)

jl-wynen: I actually have `cmake-build-scipp`, `cmake-build-scippneutron`, and `build` in my ess directory right now because I have to use dev versions of scipp and scippneutron. We could pick a more explicit name than `build`. But it would not affect my workflow. So it is your choice. As long as there is some ignored build dir.

nvaytet (approved):

PR #93, opened 2022-01-27
First pass for logging

jl-wynen:
Implements logging for reflectometry and Amor.
This will probably need many iterations to figure out what, how, and where to log. But the high level API (`get_logger`, `configure[_workflow]`) should be kind of final when the PR is merged so that we can start using it across techniques. Please take that into account when reviewing.

The  TODO about Mantid can only be resolved when Mantid 6.4 is released which provides an API to forward its logs to Python.

### DIFF
--- a/docs/instruments/amor/amor_reduction.ipynb
+++ b/docs/instruments/amor/amor_reduction.ipynb
@@ -42,9 +42,7 @@
   {
    "cell_type": "markdown",
    "id": "dde6c61a-39ed-4e85-af6e-0915baf3ad1b",
-   "metadata": {
-    "tags": []
-   },
+   "metadata": {},
    "source": [
     "## The Amor beamline\n",
     "\n",
--- a/docs/instruments/amor/amor_reduction.ipynb
+++ b/docs/instruments/amor/amor_reduction.ipynb
@@ -19,18 +19,34 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "c55f69ba-80db-4cf7-9b32-3865f952a13d",
+   "id": "455b7a61-5a5d-4d94-993c-02fe41c84e50",
    "metadata": {},
    "outputs": [],
    "source": [
+    "import logging\n",
     "import scipp as sc\n",
-    "from ess import amor"
+    "from ess import amor\n",
+    "import ess"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "1c89f088-0663-49ce-a999-8be271f2e24a",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "ess.logging.configure(level=logging.DEBUG, root='overwrite')\n",
+    "ess.logging.greet()\n",
+    "sc.display_logs()"
    ]
   },
   {
    "cell_type": "markdown",
    "id": "dde6c61a-39ed-4e85-af6e-0915baf3ad1b",
-   "metadata": {},
+   "metadata": {
+    "tags": []
+   },
    "source": [
     "## The Amor beamline\n",
     "\n",
@@ -476,7 +492,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.9.5"
+   "version": "3.9.9"
   }
  },
  "nbformat": 4,
--- a/docs/instruments/amor/amor_reduction.ipynb
+++ b/docs/instruments/amor/amor_reduction.ipynb
@@ -99,7 +99,6 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "logger.info('Loading primary data file')\n",
     "sample = amor.load(amor.data.get_path(\"sample.nxs\"),\n",
     "                   beamline=amor_beamline)\n",
     "sample"
--- a/docs/instruments/amor/amor_reduction.ipynb
+++ b/docs/instruments/amor/amor_reduction.ipynb
@@ -323,7 +323,6 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "logger.info('Loading reference file for normalization')\n",
     "reference = amor.load(amor.data.get_path(\"reference.nxs\"),\n",
     "                      beamline=amor_beamline)\n",
     "reference.coords['position'].fields.y += pixel_position_correction(reference)\n",
--- a/docs/instruments/amor/amor_reduction.ipynb
+++ b/docs/instruments/amor/amor_reduction.ipynb
@@ -36,8 +36,9 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "ess.logging.configure(level=logging.DEBUG, root='overwrite')\n",
+    "ess.logging.configure(level=logging.DEBUG, root=False)\n",
     "ess.logging.greet()\n",
+    "logger = ess.get_logger('amor_reduction')\n",
     "sc.display_logs()"
    ]
   },
@@ -103,6 +104,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "logger.info('Loading primary data file')\n",
     "sample = amor.load(amor.data.get_path(\"sample.nxs\"),\n",
     "                   beamline=amor_beamline)\n",
     "sample"
@@ -151,6 +153,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "logger.info(\"Correcting pixel positions for 'sample.nxs'\")\n",
     "def pixel_position_correction(data: sc.DataArray):\n",
     "    return data.coords['position'].fields.z * sc.tan(2.0 *\n",
     "                                                     data.coords['sample_rotation'] -\n",
@@ -326,6 +329,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "logger.info('Loading reference file for normalization')\n",
     "reference = amor.load(amor.data.get_path(\"reference.nxs\"),\n",
     "                      beamline=amor_beamline)\n",
     "reference.coords['position'].fields.y += pixel_position_correction(reference)\n",
--- a/docs/instruments/amor/amor_reduction.ipynb
+++ b/docs/instruments/amor/amor_reduction.ipynb
@@ -23,7 +23,6 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "import logging\n",
     "import scipp as sc\n",
     "from ess import amor\n",
     "import ess"
@@ -36,10 +35,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "ess.logging.configure(level=logging.DEBUG, root=False)\n",
-    "ess.logging.greet()\n",
-    "logger = ess.get_logger('amor_reduction')\n",
-    "sc.display_logs()"
+    "logger = ess.logging.configure_workflow('amor_reduction',\n",
+    "                                        filename='amor.log')"
    ]
   },
   {
@@ -153,7 +150,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "logger.info(\"Correcting pixel positions for 'sample.nxs'\")\n",
+    "logger.info(\"Correcting pixel positions in 'sample.nxs'\")\n",
     "def pixel_position_correction(data: sc.DataArray):\n",
     "    return data.coords['position'].fields.z * sc.tan(2.0 *\n",
     "                                                     data.coords['sample_rotation'] -\n",
@@ -495,8 +492,7 @@
    "mimetype": "text/x-python",
    "name": "python",
    "nbconvert_exporter": "python",
-   "pygments_lexer": "ipython3",
-   "version": "3.9.9"
+   "pygments_lexer": "ipython3"
   }
  },
  "nbformat": 4,
--- a/src/ess/amor/beamline.py
+++ b/src/ess/amor/beamline.py
@@ -6,7 +6,8 @@ from ..choppers import make_chopper
 from ..logging import log_call
 
 
-@log_call(instrument='amor', message='Constructing AMOR beamline parameters')
+@log_call(instrument='amor',
+          message='Constructing AMOR beamline from default parameters')
 def make_beamline(
     sample_rotation: sc.Variable = None,
     beam_size: sc.Variable = 0.001 * sc.units.m,
--- a/src/ess/amor/beamline.py
+++ b/src/ess/amor/beamline.py
@@ -1,6 +1,5 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2022 Scipp contributors (https://github.com/scipp)
-import logging
 import scipp as sc
 from scipp.constants import g
 from ..choppers import make_chopper
@@ -60,7 +59,7 @@ def make_beamline(
     return beamline
 
 
-@log_call(level=logging.DEBUG)
+@log_call(level='DEBUG')
 def instrument_view_components(da: sc.DataArray) -> dict:
     """
     Create a dict of instrument view components, containing:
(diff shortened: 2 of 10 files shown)


docs/instruments/amor/amor_reduction.ipynb
```diff
@@ -19,18 +19,32 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "1c89f088-0663-49ce-a999-8be271f2e24a",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "logger = ess.logging.configure_workflow('amor_reduction',\n",
```
nvaytet, line 38: It feels a bit strange that one has to manually do this in a notebook. Ideally, users should not have to care about setting up logging, they would probably expect this to happen automatically behind the scenes.
But I am not sure if it is possible to do in another way.

You are doing it this way because you want to log the name of the notebook, but is that really necessary? I guess users know which notebook they are running, and I am not sure how often one would be running multiple notebooks to do a reduction.

Would it be possible instead to initialize this when the `amor` module is imported? So that we can configure the logger to at least have `amor`?

I guess this would mean that when you import `ess`, you spawn the loggers for all the submodules (`amor`, `loki`, `wfm` etc.). Would that be an issue? Can you even have multiple loggers?
You could then do `amor.logger(...)`

docs/instruments/amor/amor_reduction.ipynb
```diff
@@ -19,18 +19,32 @@
...
+    "                                        filename='amor.log')"
    ]
   },
   {
    "cell_type": "markdown",
    "id": "dde6c61a-39ed-4e85-af6e-0915baf3ad1b",
-   "metadata": {},
+   "metadata": {
+    "tags": []
```
nvaytet: Why the empty `tags`?

src/ess/amor/beamline.py
```diff
@@ -1,10 +1,13 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2022 Scipp contributors (https://github.com/scipp)
+import logging
```
nvaytet: Is this the python `logging`, and below we have the local `ess` logging?
This feels quite error prone.

src/ess/amor/beamline.py
```diff
@@ -1,10 +1,13 @@
...
 import scipp as sc
 from scipp.constants import g
 from ..choppers import make_chopper
+from ..logging import log_call
 
 
+@log_call(message='Constructing artificial AMOR beamline parameters')
```
nvaytet:
```suggestion
@log_call(message='Constructing AMOR beamline parameters')
```

src/ess/amor/load.py
```diff
@@ -1,9 +1,11 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2022 Scipp contributors (https://github.com/scipp)
+from typing import Optional
```
nvaytet, line 3: Not sure I understand what using `Optional` brings, generally.
If the argument has a default value, it is optional. You don't really need to speel it out even more?
Sphinx's autodoc will even add the `Optional` tags for you automatically.
But maybe that is a wider discussion we can have somwhere else?

src/ess/amor/load.py
```diff
@@ -28,15 +30,19 @@ def _tof_correction(data: sc.DataArray, dim: str = 'tof') -> sc.DataArray:
 
 
 def load(filename,
-         beamline: dict = make_beamline(),
+         beamline: Optional[dict] = None,
```
nvaytet, line 33: Why is `beamline` `Optional`, but not `disable_warnings` below?

src/ess/amor/load.py
```diff
@@ -53,6 +59,7 @@ def load(filename,
     data.coords['tof'] = sc.to_unit(data.coords['tof'], 'us', copy=False)
 
     # Add beamline parameters
+    beamline = make_beamline() if beamline is None else beamline
```
nvaytet, line 62: This is a better way to do things than creating the beamline in the default args, because it used to get created as soon as you import `ess`, even if you are never going to call it. :+1: 

There are probably other places in the reflectometry where we were doing the same, and should use `None` instead?

src/ess/logging.py
```diff
@@ -0,0 +1,280 @@
...
+
+
+_INSTRUMENTS = [
+    'amor', 'beer', 'bifrost', 'cspec', 'dream', 'estia', 'freia', 'heimdal', 'loki',
+    'magic', 'miracles', 'nmx', 'odin', 'skadi', 'trex', 'v20', 'vespa'
+]
+
+
+def _deduce_instrument_name(f: Any) -> Optional[str]:
+    # Assumes package name: ess.<instrument>[.subpackage]
```
nvaytet, line 216: This sounds like it's going to break at some point (e.g. if we decide to move the instruments into an `instruments` folder).
Do we really need this? Can we require the instrument name to always be specified?

The module inspection to find the name also sounds expensive, if you make 1,000,000 calls to a function that is logging?

src/ess/logging.py
```diff
@@ -0,0 +1,280 @@
...
+
+class Formatter(logging.Formatter):
+    """
+    Logging formatter that indents messages and optionally shows threading information.
+    """
+    def __init__(self, show_thread: bool, show_process: bool):
+        """
+        Initialize the formatter.
+
+        The general as well as date formats are fixed.
```
nvaytet: Not sure I understand the meaning of the sentence. What is `general` here?

src/ess/logging.py
```diff
@@ -0,0 +1,280 @@
...
+              widget_level: Union[str, int] = logging.INFO,
+              show_thread: bool = False,
+              show_process: bool = False,
+              loggers: Optional[Sequence[Union[str, logging.Logger]]] = None):
+    """Set up logging for the ess package.
+
+    This function is meant as a helper for application (or notebook) developers.
+    It configures the loggers of ess, scippneutron, scipp, and some
+    third party packages.
+    *Calling it from a library should be avoided because*
```
nvaytet: So when can it be called? When you are in a notebook?

src/ess/logging.py
```diff
@@ -0,0 +1,280 @@
...
+
+      - *File Handler* Writes files to a file with given name.
+        Can be disabled using the `filename` argument.
+      - *Stream Handler* Writes to `sys.stderr`.
+      - *Widget Handler* Writes to a :py:class:`scipp.logging.LogWidget`
+        if Python is running in a Jupyter notebook.
+
+    :param filename: Name of the log file.
+                     Overwrites existing files.
+                     Settings this to `None` disables logging to file.
```
nvaytet:
```suggestion
                     Setting this to `None` disables logging to file.
```

src/ess/logging.py
```diff
@@ -0,0 +1,280 @@
...
+    :param file_level: Log level for the file handler.
+    :param stream_level: Log level for the stream handler.
```
nvaytet, line 142: You should also list the default values at the end of the `:param` entries.
You would think that sphinx could do that for you automatically, but I tried to use a package to do that once and it messed things up, and I couldn't build the docs anymore :-1:

nvaytet (changes requested):

docs/instruments/amor/amor_reduction.ipynb
jl-wynen: > You are doing it this way because you want to log the name of the notebook, but is that really necessary?

The notebook name is not logged. The name passed in here is only used as the name of the returned logger. This is not really necessary, users can create loggers themselves.

Concerning the main point, there are 2 distinct issues here.

Configuring at import is generally a bad idea as that slows down import. In particular logging the program version would be bad because that creates a new file or overwrites an existing one which one would not expect to happen during import. Also, this would mean that logging can never be customised programmatically, one would need a config file.
Configuration could be deferred until someone first calls `get_logger` to solve this issue.

But this links to the 2nd issue. ess is still primarily a library. As such, configuring logging automatically forces that configuration onto every user. Imagine someone at SNS importing the WFM module. That user would be forced to use our logging setup. Or, if they are aware of it, wipe the setup before doing theirs.

One possible solution would be an IPython magic. At the UX workshop we talked about the boilerplate import code that kind of distracts from the actual meaning of the notebook. We could implement a magic that imports all default modules and configures logging. Not sure this is actually a good solution because it hides the definition of names (like `import scipp as sc`), but I wanted to throw it out there...

There is also a way of automatically executing some code whenever Python starts. We could configure this on ESS machines to do whatever setup we need. But that would then be inconsistent with user's own machines. Or we do it for every env that ess is installed in. Which would again force the setup on everyone using the package.

> Can you even have multiple loggers?

As many as you like.

> I guess this would mean that when you import ess, you spawn the loggers for all the submodules (amor, loki, wfm etc.). Would that be an issue

Constructing a logger does not perform any setup. Creating instances for all submodules would not do any harm. But it would mean that we cannot tie setup to `get_logger` if that is the route we want to take.

docs/instruments/amor/amor_reduction.ipynb
jl-wynen: Don't know how they got there. Going to remove them.

src/ess/amor/beamline.py
jl-wynen: Would you prefer renaming our module?
In this particular case, we don't need to import the std lib logging because we can use strings: `@log_call(level='DEBUG')`. And we probably never need to do so outside of our logging module.

src/ess/amor/load.py
jl-wynen: It's not about sphinx but about type checkers. I don't know if they add Optional automatically.

src/ess/amor/load.py
jl-wynen: Didn't pay attention to it

src/ess/logging.py
jl-wynen: > Do we really need this? Can we require the instrument name to always be specified?

You could also argue that when a function is moved to a different module, the user might have to update the name. With the automatic detection, it would just work.
But I agree with you and I was unsure whether I should implement this. Shall I remove it?

> The module inspection to find the name also sounds expensive, if you make 1,000,000 calls to a function that is logging?

The decorator is only evaluated when the function is defined. After that, the name is stored. So the cost does not really matter.

src/ess/logging.py
jl-wynen: `logging.Formatter` takes two formats, one for the message and one for the date. I'll clarify.

src/ess/logging.py
jl-wynen: Or when you write an app of any sort.

src/ess/logging.py
jl-wynen: It does list them. But not in the params items, only in the signature at the top.

src/ess/amor/load.py
jl-wynen: The `make_beamline` function itself takes a bunch of scipp variables as default values. This is generally bad because mutable defaults should be avoided. But avoiding that would be a lot of faff. Should I do it or is it fine the way it is in this particular case?

src/ess/amor/beamline.py
jl-wynen: I put the log there because I thought that this function would eventually be replaced by loading that data from file. If this is expected to be the default way of getting beamline parameters, we may want to remove `log_call`.

src/ess/amor/load.py
nvaytet: I think we can defer to another PR, it's not really related to logging.

src/ess/amor/beamline.py
nvaytet: You are right that it is not the way it will be done in the future. It will be read from the Nexus files, so logging something is good).
However, I thought saying 'artificial' was a little strange, since Amor is an existing beamline with known parameters. Maybe we should have a different message. Something like
```
The beamline parameters were not found in the Nexus file. Constructing a beamline from default values.
```
?

src/ess/amor/load.py
nvaytet: I guess if we change those, we should add `Optional` to all the functions in the `amor` and `reflectometry` modules.
But should be a different PR.

src/ess/logging.py
nvaytet: > Shall I remove it?

I would say simple is always better, so until someone complains about having to type the instrument name, we can remove?

src/ess/logging.py
nvaytet: True. I always thought we should do like the Numpy docs, who lists them both in the signature and in the `:param:` block. But on second though, it is duplicating info, and it is less likely to get out of sync by just having them in the signature.
Maybe I should have a look at all the docstrings and change accordingly?

docs/instruments/amor/amor_reduction.ipynb
```diff
@@ -87,6 +99,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "logger.info('Loading primary data file')\n",
```
nvaytet: What about having the `load` function do the logging instead?
You pass it the `logger` as an argument, and inside `load`, `if logger is not None`, log a message?
Same for the other functions below?

src/ess/logging.py
```diff
@@ -0,0 +1,293 @@
...
+
+    configure.is_configured = True
+
+
+configure.is_configured = False
+
+
+def configure_workflow(workflow_name: Optional[str] = None,
+                       *,
+                       display: bool = running_in_jupyter(),
```
nvaytet: I'm wondering if, as you did for the beamline, we shoudl avoid calling functions for default parameters?
Set to `None` and make the jupyter check when function is called?

jl-wynen: I don't think there is a problem with calling functions as such. It slows down the creation of the function object a little. But since the result in this case is a (immutable) bool, there is no real harm.
I did it like this so that `running_in_jupyter` shows up in the docs. But it is actually just resolved by sphinx and replaced by `False`, which is not very informative. So I am going to change it.

docs/instruments/amor/amor_reduction.ipynb
jl-wynen: `load` does not know what kind of file it is loading. It makes sense to log every file that is loaded, so I am going to add that to the `load` function. But it would also be nice to show some info, like `loading normalisation`, etc. Any idea how best to do this?

jl-wynen: There already is a log call in `load`...

nvaytet: I would have thought just logging the name of the file would be enough

src/ess/logging.py
nvaytet: In this case it might not slow down very much, but I think we should not encourage this practise because soon you'll find people trying to run long python functions or entire workflows as a default parameter... which could really slow down the `import`.

jl-wynen: Agree

docs/instruments/amor/amor_reduction.ipynb
jl-wynen: Maybe. The names I have seen so far are pretty cryptic and don't tell you what the file actually contains.

nvaytet: The notebook itself should be describing what the file contains, no?

jl-wynen: Sure. But correlating logs with a notebook can be tricky.
I'll remove this log call but we should have something similar somewhere.

docs/instruments/amor/amor_reduction.ipynb
```diff
@@ -310,6 +323,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "logger.info('Loading reference file for normalization')\n",
```
nvaytet: I would also remove this logging, for the same reason as for the sample measurement.

nvaytet (approved):

PR #94, opened 2022-01-27
Commit reformatted files and run nb-clean

jl-wynen:
See https://github.com/scipp/scipp/pull/2387/files?file-filters%5B%5D=.yml&show-viewed-files=true

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,6 +32,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
         with:
+          ref: ${{ github.head_ref }}
           fetch-depth: 0  # history required so cmake can determine version
       - uses: conda-incubator/setup-miniconda@v2
       - run: conda install --channel conda-forge --yes conda-build mamba boa
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,6 +12,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.head_ref }}
       # Run this before any other commands so that git diff does not pick up formatting changes.
       - run: |
           find . -type f -not -path '.git/*' -exec sed -ri "s/[0-9]{4} (Scipp contributors)/$(date +%Y) \1/g" {} +
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,7 +18,8 @@ jobs:
       - run: |
           find . -type f -not -path '.git/*' -exec sed -ri "s/[0-9]{4} (Scipp contributors)/$(date +%Y) \1/g" {} +
           git diff --exit-code
-      - run: pip install flake8 yapf==0.30.0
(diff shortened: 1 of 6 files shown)


.github/workflows/build.yml
```diff
@@ -12,12 +12,16 @@ jobs:
...
       - uses: actions/checkout@v2
-      - run: pip install flake8 yapf==0.30.0
-      - run: yapf --recursive --in-place .
-      - run: flake8 .
+        with:
+          ref: ${{ github.head_ref }}
+      # Run this before any other commands so that git diff does not pick up formatting changes.
       - run: |
           find . -type f -not -path '.git/*' -exec sed -ri "s/[0-9]{4} (Scipp contributors)/$(date +%Y) \1/g" {} +
           git diff --exit-code
```
nvaytet, line 20: Why is the `git diff --exit-code` still being applied here?
Can't we just get the `auto-commit-action` to update the copyright?

.github/workflows/build.yml
```diff
@@ -12,12 +12,16 @@ jobs:
...
+      - run: pip install flake8 nb-clean==2.1.0 yapf==0.30.0
+      - run: find . -type f -regex '.*\.ipynb' | xargs nb-clean clean --remove-empty-cells --preserve-cell-metadata
+      - run: yapf --recursive --in-place .
+      - run: flake8 .
       - uses: stefanzweifel/git-auto-commit-action@v4
```
nvaytet, line 25: In scipp, there is a `with: commit_message` for this action

.github/workflows/build.yml
```diff
@@ -12,12 +12,16 @@ jobs:
     runs-on: ubuntu-latest
```
nvaytet, line 12: Note that any changes that you make in this workflow will not be used, because it's the `pr_and_main.yml` that is being run, and this fetches the `build.yml` from `@main`.
You have to change that to `@git-auto-format`.

The annoying thing, is that you will then have to change that back to `@main` once the PR is merged.
It's so annoying that I am thinking of getting rid of these template workflows and just duplicate the code in both `pr_and_main.yml` and `weekly_and_release.yml`

.github/workflows/build.yml
jl-wynen: Maybe to make sure this happens in a controlled manner and not across multiple PRs?

.github/workflows/build.yml
jl-wynen: Good catch!

.github/workflows/build.yml
jl-wynen: See latest commit. Like this?

nvaytet: Yes! there is a new automatic commit i see

nvaytet:
I guess we should also be applying black or yapf formatting to all the notebooks?

jl-wynen:
> I guess we should also be applying black or yapf formatting to all the notebooks?

No. We have a number of cells that we format by hand because YAPF has a brainfart sometimes and produces hard to read outputs.

nvaytet:
If you change the branch back to `@main`, i will merge

nvaytet (approved):

PR #95, opened 2022-01-28
Remove reuseable workflows

nvaytet:
Remove reuseable workflows `build.yml` and `deploy.yml` to avoid having to change the `@main` version selector every time we want to make changes to the workflows.
Duplicate code instead. There is not so much code, so this is ok I think.


.github/workflows/weekly_and_release.yml
```diff
@@ -7,12 +7,71 @@ on:
...
+  run:
+    shell: bash -l {0}  # required for conda env
+
 jobs:
+  formatting:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.head_ref }}
```
jl-wynen: I'm not actually sure if this is correct. I added it in my PR without thinking about release builds. The docs for auto-commit mention this only in the context or PR builds. So maybe we should remove it. (Also in `build:` below)

nvaytet: is it even needed for the PR builds?

jl-wynen: I don't really understand what it does. But the auto-commit docs say
> Note that we explicitly specify ${{ github.head_ref }} in the checkout Action. This is required in order to work with the pull_request event (or any other non-push event).

nvaytet: But it seems it was working fine without it... hmm

nvaytet:
@jl-wynen did you have any more comments on this?

jl-wynen (approved):
No, looks good

PR #97, opened 2022-02-01
Use normalisation in detector/q space to account for pixel differences

arm61:
A first draft (please comment on the text cause I imagine the explanation isn't great) going through performing the normalisation on dimensions of `'detector_id'` and `'Q'` instead of just `'Q'`.


docs/instruments/amor/amor_reduction.ipynb
```diff
@@ -372,8 +390,46 @@
...
+    "sc.plot(normalized)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "03cc3513-a372-4d79-bcec-abf44a48cda4",
+   "metadata": {},
+   "source": [
+    "The above plot shows the two-dimensional plots of $Q$ and the detector pixel number.\n",
+    "Here, we note that there are a large number of pixels, where there was no neutrons detected in the reference measurements, leading to values of `nan` in the normalised data.\n",
```
jl-wynen:
```suggestion
    "Here, we note that there are a large number of pixels, where there were no neutrons detected in the reference measurements, leading to values of `nan` in the normalised data.\n",
```

jl-wynen: `normalized` can also contain `inf`, right? That is why you can't use `nanmean` and have to resort to masking pixels?

jl-wynen (commented):
I think the explanations are understandable.

docs/instruments/amor/amor_reduction.ipynb
arm61: Exactly, I have slightly changed the comment to reflect this.

nvaytet: I have been thinking for a while that we could add `finitemean`, `finitesum`, `finitemax` etc. to scipp, that would ignore both NaNs and Infs (see e.g. https://github.com/scipp/scipp/issues/1094).
Maybe we can re-visit this idea?

docs/instruments/amor/amor_reduction.ipynb
```diff
@@ -331,38 +331,56 @@
...
    ]
   },
   {
    "cell_type": "markdown",
-   "id": "fa8fb76d-3cd8-4adb-89eb-fc00e47ea6b5",
+   "id": "a3af2ec5-f4ec-400d-ab89-a483262daf34",
    "metadata": {},
    "source": [
-    "And we then convert to $Q$ as well"
+    "However, if we want to obtain the normalisation between the sample data and the reference (accounting for detector pixel discontinuities), we want to work with the data with dimensions of `'detector_id'` and `'Q'` (the `*_q_binned` objects). "
```
nvaytet: I think this is making reading the notebook read a bit more confusing (talking about the `*q_binned` quantities).
I would re-work the notebook so that this is what you want to do from the beginning.
You don't define any `*_q_summed` quantities, you just stick with the `*_q_binned` (and possibly give them a better name), just say that you're making a quick preview in Q by plotting the sum over `detector_id`, just as a sanity check.

And just move on to making the normalization as you want it.
There is no need to keep in the notebook the fact that we started by doing this, and then we improved it. The reader doesn't need to know about the developement history of the workflow.

docs/instruments/amor/amor_reduction.ipynb
```diff
@@ -372,8 +390,46 @@
...
+   "source": [
+    "normalized.masks['no_reference_neutrons'] = (reference_norm == sc.scalar(0)).data"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "64a82569-5c12-480e-a120-85b0215f1e31",
+   "metadata": {},
+   "source": [
+    "Finally, we perform a mean along the `'detector_id'` to obtain the normalised reflectometry profile. "
```
nvaytet: I am not sure I understand why we are now doing a `mean` when we did a `sum` before...

docs/instruments/amor/amor_reduction.ipynb
arm61: This is a good point. I am planning to add the footprint correction soon so I suggest that when I add that we rework the whole thing (I suggest I get a first draft soon and you can go through it). This approach might also help to understand why `mean` instead of `sum` before (cause at the moment I am not 100 % sure why `mean` gives the correct result and `sum` doesn't).

nvaytet: I don't think it's too much work to change what I suggested, probably just a small number of cells that need changing?

nvaytet: > (cause at the moment I am not 100 % sure why mean gives the correct result and sum doesn't)

maybe we should dig down a little bit more and understand 100% before we move forward, as we may forget at some point that we were not 100% sure, and after a while everyone just assumes it's all correct?

The order of the operations is a little different now, maybe if we write them all down we can see where the mean comes from?

arm61: Very much agree, and in the process of second drafting this I think I am grepping it a bit better

arm61:
I have added some new features (specifically the footprint correction and the supermirror calibration which were present in the original iteration of this code. Additionally, there has been a lot of new writing in the notebook. Feedback pls.

I will add tests in the coming days.

arm61:
Have just spoken with the Amor instrument scientist and he recommends a slightly different footprint correction that I will look at tomorrow morning.

arm61:
Okay, I have fixed the footprint correction and sorted when a mistaken division.

arm61:
nb. there are two parts that will be replaced with `scipp-0.12` functionality once it is ready.

docs/instruments/amor/amor_reduction.ipynb
```diff
@@ -46,26 +44,34 @@
...
     "In general the detector centre is located at $\\gamma_{\\rm D} = 2\\omega$.\n",
-    "These are instrument coordinates and should not be confused with the situation on the sample,\n",
-    "where the take-off angle of an individual neutron trajectory is called $\\theta$.\n",
+    "These are instrument coordinates and should not be confused with the situation on the sample, where the take-off angle of an individual neutron trajectory is called $\\theta$.\n",
+    "\n",
+    "### The supermirror reference\n",
+    "\n",
+    "Data collected at the Amor instrument in divergent mode requires normalisation by a reference measurement from a neutron supermirror. \n",
+    "The supermirror is described by some properties; an $m$-value, critical edge and $\\alpha$, which can be used to calibriate the supermirror to that \n",
+    "\n",
```
nvaytet:
```suggestion
```

docs/instruments/amor/amor_reduction.ipynb
```diff
@@ -46,26 +44,34 @@
...
+    "The normalisation of data from the Amor instrument in divergent mode requires a reference measurement of a neutron supermirror.\n",
+    "The supermirror is not a perfect supermirror, and is described with some properties, an *m*-value, critical edge, and $\\alpha$, from which we can calibrate the supermirror. \n",
```
nvaytet:
```suggestion
    "The supermirror is not a perfect supermirror, and is described with some properties, an $m$-value, a critical edge, and an $\\alpha$, from which we can calibrate the supermirror. \n",
```

docs/instruments/amor/amor_reduction.ipynb
```diff
@@ -46,26 +44,34 @@
...
+    "This reference measurement facilitates two normalisation on our data:\n",
```
nvaytet:
```suggestion
    "This reference measurement facilitates two normalisations on our data:\n",
```

docs/instruments/amor/amor_reduction.ipynb
```diff
@@ -46,26 +44,34 @@
...
+    "- normalisation of neutron count per unit time, assuming that the instrument flux is constant between the supermirror measurement and our sample measurement,\n",
+    "- normalisation over the detector pixels, to account for differences in pixel efficiency.\n",
+    "It is important when this normalisation is performed that the differences in count time and beam footprint are accounted for such that the measurements are commensurate.\n",
```
nvaytet, line 66: :+1:

docs/instruments/amor/amor_reduction.ipynb
```diff
@@ -76,20 +82,28 @@
...
     "## Loading the data\n",
     "\n",
-    "Using the `amor.load` function, we load the `sample.nxs` file and perform some early preprocessing:\n",
+    "The `sample.nxs` file is the experimental data file of interest,\n",
+    "while `reference.nxs` is the reference measurement of the neutron supermirror.\n",
+    "The `amor.load` function can be used to load these files and perform some early preprocessing:\n",
     "\n",
     "- The `tof` values are converted from nanoseconds to microseconds.\n",
-    "- The raw data contains events coming from two pulses, and these get folded into a single `tof` range"
+    "- The raw data contains events coming from two pulses, and these get folded into a single `tof` range\n",
```
nvaytet:
```suggestion
    "- The raw data contains events coming from two pulses, and these get folded into a single `tof` range.\n",
```

docs/instruments/amor/amor_reduction.ipynb
```diff
@@ -152,7 +170,8 @@
     "    return data.coords['position'].fields.z * sc.tan(2.0 *\n",
     "                                                     data.coords['sample_rotation'] -\n",
     "                                                     (0.955 * sc.units.deg))\n",
-    "sample.coords['position'].fields.y += pixel_position_correction(sample)"
+    "sample.coords['position'].fields.y += pixel_position_correction(sample)\n",
```
nvaytet, line 166: It might be worth considering putting the sample and reference in a `Dataset`, and then you only need to change one coordinate position.
Note that you would have to align the data arrays to the same bin edges in `tof`.

docs/instruments/amor/amor_reduction.ipynb
```diff
@@ -271,108 +290,216 @@
...
    ]
   },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "0361b1c0-64e4-4b68-b290-9f64960613ee",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "reference_wav = reference.transform_coords([\"wavelength\"], graph=graph)\n",
```
nvaytet, line 293: Same as above, you would perform a single unit conversion on the dataset?

nvaytet: hmm now that I think about it, if you put them in the same dataset, you will need to change the `sample_rotation` from being a coordinate to being an attribute, but maybe that's ok?

docs/instruments/amor/amor_reduction.ipynb
```diff
@@ -271,108 +290,216 @@
...
+   "source": [
+    "sample_theta = sample_wav.transform_coords(['theta'], graph=graph)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "aa3e6be8-a6d4-407b-a168-ffca211c0033",
+   "metadata": {},
+   "source": [
+    "The theta coordinate is a bin-edge coordinate (with only two edges), so we will find the centres of these. "
```
nvaytet: > (with only two edges)

I feel something weird is going on here. You have two edges, because theta is an event coordinate. The two edges encompass the entire range of theta values, but there is a theta value per event.

I think computing the `theta_centres` below to simply get a single number half-way between the 2 boundaries doesn't feel right, byt maybe I am not understanding exactly what you are trying to achieve here?

docs/instruments/amor/amor_reduction.ipynb
```diff
@@ -271,108 +290,216 @@
...
    "outputs": [],
    "source": [
-    "reference = amor.load(amor.data.get_path(\"reference.nxs\"),\n",
-    "                      beamline=amor_beamline)\n",
-    "reference.coords['position'].fields.y += pixel_position_correction(reference)\n",
-    "reference"
+    "fwhm_to_std = sc.scalar(2.) * sc.sqrt(sc.scalar(2.) * sc.log(sc.scalar(2.)))\n",
+    "footprint_scale = sc.erf(sample_theta.coords['sample_size'] / beam_on_sample * fwhm_to_std)\n",
+    "# To be replaced with squeeze\n",
+    "sample_theta.coords['footprint_scale'] = sc.Variable(values=footprint_scale.values[0], unit=sc.units.dimensionless, dims=['detector_id'])"
```
nvaytet: If you can replace with `squeeze`, why are you creating a new Variable? Could you just slice `footprint_scale` instead?

docs/instruments/amor/amor_reduction.ipynb
```diff
@@ -271,108 +290,216 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "64a6b838-c594-48bc-a85f-3ba9a40088a6",
+   "metadata": {
+    "tags": []
+   },
+   "source": [
+    "## Compute the Q vector\n",
+    "\n",
+    "Using the same method, we can compute the $Q$ vector,\n",
```
nvaytet:
```suggestion
    "Once again using `transform_coords`, we can compute the $Q$ vector,\n",
```

docs/instruments/amor/amor_reduction.ipynb
```diff
@@ -271,108 +290,216 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "daf577ef-55c0-4556-aeb0-685a339db05e",
    "metadata": {},
+   "outputs": [],
    "source": [
-    "However, if we want to obtain the normalisation between the sample data and the reference (accounting for detector pixel discontinuities), we want to work with the data with dimensions of `'detector_id'` and `'Q'` (the `*_q_binned` objects). "
+    "reference_q_summed = reference_q_binned.bins.sum()\n",
+    "calibration_factor = amor.conversions.supermirror_calibration(reference_q_summed.coords['Q'])\n",
```
nvaytet: Not sure if `conversions` is the right place to put this. It's not doing a unit conversion?

docs/instruments/amor/amor_reduction.ipynb
```diff
@@ -271,108 +290,216 @@
...
   {
    "cell_type": "markdown",
-   "id": "4312f0a3-6c80-4046-a223-efc5f057f7e7",
+   "id": "a3af2ec5-f4ec-400d-ab89-a483262daf34",
    "metadata": {},
    "source": [
-    "For each of these objects, we should determine the number of counts in each bins and normalise this by the total number of counts in the measurement."
+    "## Normalization by the super-mirror\n",
+    "\n",
+    "For each of the measurements, we should determine the number of counts in each bins and normalise this by the total number of counts in the measurement and the footprint scale factor.\n",
```
nvaytet:
```suggestion
    "For each of the measurements, we should determine the number of counts in each bin and normalise this by the total number of counts in the measurement and the footprint scale factor.\n",
```

docs/instruments/amor/amor_reduction.ipynb
```diff
@@ -285,108 +421,171 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "784c11db-f6d8-4959-8777-3be98aa8ba16",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "normalized = sample_norm / reference_norm\n",
+    "normalized.coords['footprint_ratio'] = ratio\n",
+    "sc.plot(normalized)"
```
nvaytet: This plot isn't very useful in the current state. A bug-fix in scipp helps getting the correct range for the colorbar, but it still doesn't help with the resolution (we only see thin lines and lofs of white soace in between).
We should probably use a coarser resolution.

Incidently, if we do that, we will propagate the nans.
We can either replace the nan values with zeros before we plot, or maybe we could plot the sample and reference separately before normalizing, as is done in the powder reduction workflow? (https://scipp.github.io/scippneutron/tutorials/powder-diffraction.html#From-events-to-histogram)

docs/instruments/amor/amor_reduction.ipynb
```diff
@@ -285,108 +421,171 @@
...
   },
   {
    "cell_type": "markdown",
-   "id": "a3af2ec5-f4ec-400d-ab89-a483262daf34",
+   "id": "64a82569-5c12-480e-a120-85b0215f1e31",
    "metadata": {},
    "source": [
-    "Finally, we divide the sample by the reference to obtain"
+    "Finally, we perform a mean along the `'detector_id'` to obtain the normalised reflectometry profile. \n",
+    "We use mean such that the masks pixels are not accounted for. "
```
nvaytet: Delete this line? We already say above that we need to mask before taking the mean, so it should be obvious that the `mean` operation ignores masked data?

docs/instruments/amor/amor_reduction.ipynb
```diff
@@ -285,108 +421,171 @@
...
    ]
   },
   {
    "cell_type": "markdown",
-   "id": "a3af2ec5-f4ec-400d-ab89-a483262daf34",
+   "id": "64a82569-5c12-480e-a120-85b0215f1e31",
    "metadata": {},
    "source": [
-    "Finally, we divide the sample by the reference to obtain"
+    "Finally, we perform a mean along the `'detector_id'` to obtain the normalised reflectometry profile. \n",
```
nvaytet: Did you ever get to the bottom of why we are now doing `mean` as opposed to `sum`?

docs/instruments/amor/amor_reduction.ipynb
```diff
@@ -447,8 +646,8 @@
    "outputs": [],
    "source": [
     "nbins = 165\n",
-    "theta_edges = sc.linspace(dim='theta', start=0.4, stop=1.2, num=nbins, unit='deg')\n",
-    "wavelength_edges = sc.linspace(dim='wavelength', start=1.0, stop=15.0, num=nbins, unit='angstrom')\n",
+    "theta_edges = sc.linspace(dim='theta', start=0.0, stop=1.2, num=nbins, unit='deg')\n",
+    "wavelength_edges = sc.linspace(dim='wavelength', start=0, stop=15.0, num=nbins, unit='angstrom')\n",
     "binned = sc.bin(sample_theta, edges=[sc.to_unit(theta_edges, 'rad'), wavelength_edges])\n",
     "binned"
```
nvaytet, line 627: There are a lot of html outputs in this section, and I feel it's a little too crowded (this was the case already before this PR). Maybe we could remove at least one of them, e.g. the one above after we have done `sample_theta.bins.concatenate('detector_id')`. For example, instead of just showing the entire `sample_theta`, we could show that the `detector_id` dim has disappeared by showing `sample_theta.dims`?

docs/instruments/amor/amor_reduction.ipynb
```diff
@@ -466,7 +665,10 @@
   {
    "cell_type": "markdown",
    "id": "b4ccc717-7b2d-491e-b202-3f16f0a825af",
-   "metadata": {},
+   "metadata": {
+    "jp-MarkdownHeadingCollapsed": true,
```
nvaytet: remove metadata

docs/instruments/amor/amor_reduction.ipynb
```diff
@@ -271,10 +290,127 @@
...
+    "footprint_scale = sc.erf(reference_theta.coords['sample_size'] / beam_on_sample * fwhm_to_std)\n",
+    "# To be replaced with squeeze\n",
+    "reference_theta.coords['footprint_scale'] = sc.Variable(values=footprint_scale.values[0], unit=sc.units.dimensionless, dims=['detector_id'])"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "64a6b838-c594-48bc-a85f-3ba9a40088a6",
+   "metadata": {
+    "tags": []
```
nvaytet: remove tags?

src/ess/amor/conversions.py
```diff
@@ -20,3 +21,27 @@ def specular_reflection() -> dict:
     graph = spec_relf_graph()
     graph['incident_beam'] = incident_beam
     return graph
+
+
+def supermirror_calibration(
```
nvaytet: You should add this to the API reference for reflectometry, otherwise it will not show up in the docs.

src/ess/amor/beamline.py
```diff
@@ -9,9 +9,9 @@
 @log_call(instrument='amor',
           message='Constructing AMOR beamline from default parameters')
 def make_beamline(
-    sample_rotation: sc.Variable = None,
-    beam_size: sc.Variable = 0.001 * sc.units.m,
-    sample_size: sc.Variable = 0.01 * sc.units.m,
+    sample_rotation: sc.Variable = 45.0 * sc.units.deg,
```
nvaytet: I think having a default value is potentially evil, because if you forget to supply it, it could yield confusion because you are getting strange results?
I am thinking it should have no default value and always be required?

src/ess/amor/beamline.py
```diff
@@ -9,9 +9,9 @@
...
+    beam_size: sc.Variable = 2.0 * sc.units.mm,
```
nvaytet: I think we want to slowly move away from this way of specifying default values.

All these default values are created when importing the modules, and so if someone specifies a really large array as a default value, it will create memory overhead and slow-down when importing `ess.relfectometry`.
Default values should only be created when the functions are called.

The new way is therefore to set them to `None` and then update the values if they are `None` in the function body.
I know it kind of defeats the point of default values, but I think the potential issues listed above make it worthwhile.

src/ess/amor/beamline.py
```diff
@@ -47,7 +47,7 @@ def make_beamline(
         'beam_size': beam_size,
         'sample_size': sample_size,
         'detector_spatial_resolution': detector_spatial_resolution,
-        'gravity': gravity
+        'gravity': gravity,
```
nvaytet: why this change? My opinion on trailing commas is that although it does make it easier in the future to add another argument, as you only have to change one line, i think a big drawback is that it can cause confusion/distraction.

Someone reading the code might think "was that supposed to be there?", "was there something after this that was deleted and maybe shouldn't have been?"
I think it is just safer to avoid the confusion and make it clear that your dictionary (and the same thing goes for lists and function calls) is complete.

I will add this to the style guide notes.

src/ess/amor/conversions.py
nvaytet: As i mentioned above, this should probably go in a different place, as it is not performing a unit conversion.

src/ess/amor/conversions.py
```diff
@@ -20,3 +21,27 @@ def specular_reflection() -> dict:
...
+    Determine calibration factor for the supermirror.
+
+    :param q_bins: Q-bins over which calibration is performed.
+    :param m_value: m-value for the supermirror.
+    :param critical_edge: Supermirror critical edge.
+    :param alpha: Supermirror alpha value.
+
+    :return: Calibration factor at the midpoint of each Q-bin.
+    """
+    q_midpoints = _midpoints(q_bins, 'Q')
```
nvaytet: This assumes that you always have bin edges. Maybe there should be the option to not convert to midpoints if midpoints are already being supplied?

src/ess/amor/beamline.py
arm61: This is a mistake cause in a previous version of the code I had another item in the dictionary that I deleted and forgot to remove the comma 🤣

src/ess/amor/beamline.py
arm61: I think the reason that I set `sample_rotation` to a value was that when it was `None` there was an error thrown when I run the code in the notebook. Let me double check this logic though

docs/instruments/amor/amor_reduction.ipynb
arm61: The answer is stupidity 😄

src/ess/amor/conversions.py
arm61: Is there a method to check if something has bin edges?

jl-wynen: Not really. There is this: https://github.com/scipp/scipp/blob/5e706f0ee292d414c82ae3d7b2014aac1f43a6e5/src/scipp/html/formatting_html.py#L181
But it is not really part of the public API.

src/ess/amor/conversions.py
arm61: It is specific to Amor (currently) so should it go in the instruments section?

src/ess/amor/conversions.py
arm61: https://github.com/scipp/scipp/issues/2467
🚨 feature request coming though 🚨

src/ess/amor/conversions.py
nvaytet: I would just make a new `calibration` file?

src/ess/amor/conversions.py
nvaytet: Note that a variable cannot know whether it itself represents bin edges, only the entire data array can know that.
You could pass an additional argument that tells it if it is bin edges or not.
E.g.
```Py
supermirror_calibration(q_bins, is_bin_edge=q_bins.sizes['dim'] > da.sizes['dim'])
```

docs/instruments/amor/amor_reduction.ipynb
nvaytet: Let me know if you need some help with this step

arm61: I think that keeping them separate is desirable over combining. For example, it may be that the detector is in a different position when measuring the reference vs the sample.

docs/instruments/amor/amor_reduction.ipynb
arm61: The aim is to find the theta for each event, perhaps I should be binning in theta first?

docs/instruments/amor/amor_reduction.ipynb
nvaytet: But if that is the case, then you cannot perform a normalization? If the pixels were not in the same place...

arm61:
The build failing is cause I am using features from `scipp-0.12` but (I think) due to `mantid-framework` being installed we are getting `sscipp-0.11` instead.

jl-wynen:
Correct. We are working on an update to `ess` that works with scipp 0.12. But we first need to release a patch...

docs/instruments/amor/amor_reduction.ipynb
arm61: I added this to explain that "We use mean such that the masks pixels are not accounted for." but I am now not clear what the hell I was talking about, so let me remember and I will improve

docs/instruments/amor/amor_reduction.ipynb
arm61: I am not sure how to remove these and the metadata in general.

docs/instruments/amor/amor_reduction.ipynb
arm61: I  have improved this in a later version.

docs/instruments/amor/amor_reduction.ipynb
arm61: My issue with plotting separately is that it isn't very useful (as they look quite similar).

docs/instruments/amor/amor_reduction.ipynb
arm61: If you were in `('detector_id', 'lambda')` space then it should be possible no?

arm61:
Rebased to main now that there is a `0.4` release.

docs/instruments/amor/amor_reduction.ipynb
nvaytet: just manually edit the notebook in a text editor so that the metadat looks something like
`"metadata": {},`

nvaytet:
Not sure what happened in the rebase, but I am seeing in the diff the changes from the release that are unrelated to this PR...

docs/instruments/amor/amor_reduction.ipynb
nvaytet: I don't think so, you would be comparing two different things.

arm61:
Bugger, suggestions to resolve (squash?)

docs/instruments/amor/amor_reduction.ipynb
nvaytet: you already have the theta for each event from doing `sample_theta = sample_wav.transform_coords(['theta'], graph=graph)`.
The coordinate on the top level is just giving you the ranged spanned by the underlying event coord (which is the theta for each event).

So what do you need for the footprint below? What does the footprint represent? Is it a quantity per event also?

nvaytet:
How did you do your `rebase`? You could try merging instead?

arm61:
Merge main?

arm61:
> How did you do your `rebase`? 

Honestly can't remember now...

docs/instruments/amor/amor_reduction.ipynb
nvaytet: As per our discussion on slack, let's keep these in separate data arrays for now.

docs/instruments/amor/amor_reduction.ipynb
nvaytet: This still needs to be resolved. The theta for each event is in the `sample_theta.bins.coords['theta']` and `sample_theta.coords['theta']`.

docs/instruments/amor/amor_reduction.ipynb
```diff
@@ -271,111 +287,246 @@
...
    "execution_count": null,
-   "id": "a085f3e6-7351-4779-96ce-60600a988cf0",
+   "id": "856b38c4-cc74-4a00-ae19-fbc3e9d2bda9",
    "metadata": {},
    "outputs": [],
    "source": [
-    "reference_wav = reference.transform_coords([\"wavelength\"], graph=graph)\n",
-    "reference_wav = sc.bin(reference_wav, edges=[wavelength_edges])\n",
-    "reference_wav.bins.concatenate('detector_id').plot()"
+    "fwhm_to_std = sc.scalar(2.) * sc.sqrt(sc.scalar(2.) * sc.log(sc.scalar(2.)))\n",
```
nvaytet: I seem to remember these were in a helper function before. Maybe we should do this again?

src/ess/amor/calibrations.py
```diff
@@ -0,0 +1,35 @@
...
+    :param alpha: Supermirror alpha value.
+
+    :return: Calibration factor at the midpoint of each Q-bin.
+    """
+    if is_bin_edge:
+        q = sc.midpoints(coord, 'Q')
+    else:
+        q = coord
+    max_q = m_value * critical_edge
+    lim = (q < critical_edge).astype(float)
```
nvaytet, line 28: As Simon suggested, we can probably make use of `sc.where` here

docs/instruments/amor/amor_reduction.ipynb
arm61: I have added it as a constant in the `amor.tools` directory for now. Let me know if you think a function is more valuable.

docs/instruments/amor/amor_reduction.ipynb
```diff
@@ -271,111 +283,242 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "856b38c4-cc74-4a00-ae19-fbc3e9d2bda9",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from ess.amor.tools import FWHM_TO_STD\n",
```
nvaytet: You could also consider making a function that takes in the `sample_size / beam_on_sample` and then converts to std by multiplying by `fwhm_to_std`.

e.g.
```Py
def fwhm_to_std(x):
    return x * sc.scalar(2.) * sc.sqrt(sc.scalar(2.) * sc.log(sc.scalar(2.)))
```
?

Note that you also need to replace `fwhm_to_std = sc.scalar(2.) * sc.sqrt(sc.scalar(2.) * sc.log(sc.scalar(2.)))` further down below

docs/instruments/amor/amor_reduction.ipynb
nvaytet: Just remove the plot altogether?

docs/instruments/amor/amor_reduction.ipynb
```diff
@@ -385,8 +528,48 @@
...
+   "source": [
+    "normalized.masks['no_reference_neutrons'] = (reference_norm == sc.scalar(0)).data"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "64a82569-5c12-480e-a120-85b0215f1e31",
+   "metadata": {},
+   "source": [
+    "Finally, we perfrom a mean along the `'detector_id'` to obtain the normalised reflectometry profile.\n",
```
nvaytet:
```suggestion
    "Finally, we perform a mean along the `'detector_id'` to obtain the normalised reflectometry profile.\n",
```

nvaytet: And are you sure about the `mean` here? because the values at low q in the final normalized plot went from ~0.1 to ~10

If you had 3 detector pixels, before we were doing
```
(counts_1 + counts_2 + counts_3) / (reference_1 + reference_2 + reference_3)
```
and now we are doing
```
(counts_1/reference_1 + counts_2/reference_2 + counts_3/reference_3) / 3
```
?

We should check the final values against Jochen's code.

docs/instruments/amor/amor_reduction.ipynb
arm61: My rational for the current implementation was that this allows for the inverse to give the equivalent of `STD_TO_FWHM` without another function.

docs/instruments/amor/amor_reduction.ipynb
arm61: Because we have masked pixels on reference I think that the second is the one we want...There is also the fact that using `sum` gives a result at *looks* wrong (i.e. the critical edge should be flat). Jochen's code doesn't do the normalisation, I need to go a level up to `eos.py` which can do the normalisation but will take quite a while to get to work.

nvaytet: To me, a mean doesn't really make sense, but maybe that's just me. We need to be careful about what we mean by zero counts. Is it just zero because we didn't run the experiment long enough, and statistically we would have had some counts if we recorded longer? Or does it mean zero in the sense that there simply cannot be counts there, just like if you did not have a detector pixel there?

nvaytet: I don't think just saying the shape looks better is enough justification, especially since the values have changed by so much. We need to understand what is correct.
The fact that there are nan values would mess up both a sum and and mean operation so that should not be a reason to use one or the other?

arm61: I very much agreed *looks right* isn't good scientific justification and we still need to understand it (but it indicates it is correct). Changed by so much compared to?

The `sum` counts all the pixels and does no division, the `mean` counts the pixels that are un-masked and divides by the number of un-masked pixels. The number of unmasked pixels changes as a function of q, which is why there is a difference between our sum and our mean. And I think it makes sense to consider that weighting (with sum doesn't do).

docs/instruments/amor/amor_reduction.ipynb
```diff
@@ -528,47 +521,47 @@
...
+    "normalized = sample_norm / reference_norm"
    ]
   },
   {
    "cell_type": "markdown",
    "id": "03cc3513-a372-4d79-bcec-abf44a48cda4",
    "metadata": {},
    "source": [
-    "The above plot shows the two-dimensional plots of $Q$ and the detector pixel number.\n",
+    "The plot below shows the reflecivity as a function of `'detector_id'` and `'Q'`, where there is no intensity, there is no neutrons being detected by the sample or reference datasets. \n",
```
nvaytet:
```suggestion
    "The plot below shows the reflectivity as a function of `'detector_id'` and `'Q'`, where there is no intensity, there is no neutrons being detected by the sample or reference datasets. \n",
```

nvaytet: Also, not sure why you need the second part of the sentence? It seems to either be self-evident, or be repeating some of what is said in the sentence just below?

src/ess/amor/tools.py
```diff
@@ -4,9 +4,29 @@
...
+def fwhm_to_std(fwhm: sc.Variable) -> sc.Variable:
+    """
+    Convert from full-width half maximum to standard deviation.
+    
+    :param fwhm: Full-width half maximum.
+    :return: Standard deviation.
+    """
+    # Enables the conversion from full width half
+    # maximum to standard deviation
+    return fwhm * sc.scalar(2.) * sc.sqrt(sc.scalar(2.) * sc.log(sc.scalar(2.)))
```
nvaytet:
```suggestion
    return fwhm * _FWHM_TO_STD
```

src/ess/amor/tools.py
```diff
@@ -4,9 +4,29 @@
...
+def std_to_fwhm(std: sc.Variable) -> sc.Variable:
+    """
+    Convert from standard deviation to full-width half maximum.
+    
+    :param std: Standard deviation.
+    :return: Full-width half maximum.
+    """
+    # Enables the conversion from full width half
+    # maximum to standard deviation
+    return std / sc.scalar(2.) * sc.sqrt(sc.scalar(2.) * sc.log(sc.scalar(2.)))
```
nvaytet:
```suggestion
    return std / _FWHM_TO_STD
```

nvaytet (approved):
Had another look and changes look ok to me.

arm61:
Same here. Happy for you to merge whenever.

PR #110, opened 2022-02-24
Use Napoleon for docs

jl-wynen:


.github/workflows/pr_and_main.yml
```diff
@@ -45,6 +45,7 @@ jobs:
           fetch-depth: 0  # history required so cmake can determine version
       - uses: conda-incubator/setup-miniconda@v2
       - run: conda install --channel conda-forge --yes conda-build mamba boa
+      - run: pip install sphinx-autodoc-typehints>=1.17.0
```
nvaytet: Why do you need this here if it is also being installed inside the `meta.yaml` file?

ess-developer.yml
```diff
@@ -26,8 +26,6 @@ dependencies:
```
nvaytet, line 26: remove file?

nvaytet (approved):

PR #111, opened 2022-02-28
Remove developer env file and add getting started dev docs

nvaytet:


docs/developer/getting-started.rst
```diff
@@ -0,0 +1,62 @@
...
+    --channels=conda-forge,scipp --merge-with=developer-extra.yml
+  conda env create -f ess-developer.yml
+  conda activate ess-developer
+
+Once you have activated the environment, if you want to import the ``ess`` module from
+the source, you will have to update your ``PYTHONPATH``
+
+.. code-block:: bash
+
+  export PYTHONPATH=$PYTHONPATH:src/.
```
jl-wynen: There should be a way to use `conda develop` instead. Haven't tried that though. I always use `pip install -e .`.
Either should be preferred over setting PYTHONPATH manually.

nvaytet: Does `pip install -e .` work with `ess`? I never tried it and don't know if the `setup.py` file handles this properly.

jl-wynen: It does for me

jl-wynen (approved):

PR #113, opened 2022-02-28
Add simple POWGEN workflow

jl-wynen:


docs/index.rst
```diff
@@ -36,11 +36,13 @@ Documentation
 
    instruments/amor/amor
    instruments/loki/loki
+   instruments/powgen/powgen
```
SimonHeybrock: I don't think we should add "top level" instrument submodules for instruments that are not at least related to ESS work?

jl-wynen: What about amor?
I was thinking of removing that module once we have something for DREAM. So it wouldn't be a big issue, I think.

SimonHeybrock: AMOR is closely related to ESTIA (using very similar detector technology, same scientists, used for testing ESS hardware, ...).

src/ess/diffraction/conversions.py
```diff
@@ -0,0 +1,106 @@
...
+    .. math:: \mathsf{tof} = \mathsf{DIFA} * d^2 + \mathsf{DIFC} * d + t_0
+
+    This function can be used with :func:`scipp.transform_coords`.
+
+    See Also
+    --------
+    ess.diffraction.conversions.to_dspacing_with_calibration
+    """
+    if sc.all(difa == sc.scalar(0.0, unit=difa.unit)).value:
+        return (tof - tzero) / difc
```
SimonHeybrock: Could write this with one fewer allocation?

src/ess/diffraction/conversions.py
```diff
@@ -0,0 +1,106 @@
...
+    return _dspacing_from_diff_calibration_generic_impl(tof, tzero, difa, difc)
+
+
+def _restore_tof_if_in_wavelength(data: sc.DataArray) -> sc.DataArray:
+    if 'wavelength' not in data.dims:
+        return data
+
+    # TODO better error message
+    tof_data = data.transform_coords('tof', graph={}, quiet=True)
```
SimonHeybrock: This is to turn this into a coord?

src/ess/diffraction/conversions.py
```diff
@@ -0,0 +1,106 @@
...
+    del tof_data.coords['wavelength']
+    if tof_data.bins:
+        del tof_data.bins.coords['wavelength']
```
SimonHeybrock: Any reason for this, apart from saving memory?

src/ess/diffraction/corrections.py
```diff
@@ -0,0 +1,217 @@
...
+    ess.diffraction.load_calibration
+    """
+    dim = calibration.dim
+    if not sc.identical(into.coords[dim], calibration.coords[dim]):
+        raise ValueError(
+            f'Coordinate {dim} of calibration and target dataset do not agree.')
+    out = into.copy(deep=False)
+    for name in ('difa', 'difc', 'tzero'):
+        out.attrs[name] = calibration[name].data
+    out.masks['cal'] = calibration['mask'].data
```
SimonHeybrock: May silently remove an existing `cal` mask.

src/ess/diffraction/corrections.py
```diff
@@ -0,0 +1,217 @@
...
+    :
+        Copy of `into` with additional coordinates and masks
+        from `calibration`.
+
+    See Also
+    --------
+    ess.diffraction.load_calibration
+    """
+    dim = calibration.dim
+    if not sc.identical(into.coords[dim], calibration.coords[dim]):
```
SimonHeybrock: What is this coord? Pixel IDs?

src/ess/diffraction/conversions.py
```diff
@@ -0,0 +1,106 @@
...
+    else:
+        out = data
+
+    out = _restore_tof_if_in_wavelength(out)
+    graph = {'dspacing': dspacing_from_diff_calibration}
+    out = out.transform_coords('dspacing', graph=graph)
+
+    for name in ('sample_position', 'source_position', 'position'):
+        if name in out.coords:
+            out.attrs[name] = out.coords.pop(name)
```
SimonHeybrock: Why is this necessary? Isn't this what we tried to avoid by carefully designing `transform_coords`?

src/ess/diffraction/corrections.py
```diff
@@ -0,0 +1,217 @@
...
+    for the combination of all inputs:
+
+        ^     | --- |
+    spectrum  |--   |
+        v     |   --|
+              < dim >
+    """
+    def extremum(fn, index):
+        aux_dim = uuid.uuid4().hex
+        return fn(sc.concat([fn(edge[dim, index]) for edge in edges], aux_dim), aux_dim)
```
SimonHeybrock: Looking for https://scipp.github.io/generated/functions/scipp.reduce.html#scipp.reduce?

src/ess/diffraction/corrections.py
```diff
@@ -0,0 +1,217 @@
...
+
+    lo = extremum(sc.min, 0)
+    hi = extremum(sc.max, 1)
```
SimonHeybrock: Is the 1 intentional, or should it be -1?

src/ess/diffraction/corrections.py
```diff
@@ -0,0 +1,217 @@
...
+                                     dim='wavelength')
+    data.coords['wavelength'] = wavelength_edges
+    empty_instr.coords['wavelength'] = wavelength_edges
+
+    if 'tof' in data.attrs:
+        tof_edges = _common_edges(data.attrs['tof'],
+                                  empty_instr.attrs['tof'],
+                                  dim='wavelength')
+        data.attrs['tof'] = tof_edges
+        empty_instr.attrs['tof'] = tof_edges
```
SimonHeybrock: An awful lot of code... can this somehow be ensured by applying the same workflow to both data and empty instrument before calling this? Is there a practical use for different edges?

src/ess/diffraction/corrections.py
```diff
@@ -0,0 +1,217 @@
...
+        `smooth_args` is passed as keyword arguments to
+        :func:`ess.diffraction.fft_smooth`. If ``None``, the monitor is not smoothed.
+
+    Returns
+    -------
+    :
+        `data` normalized by a monitor.
+    """
+    mon = data.meta[monitor].value
+    if 'wavelength' not in mon.coords:
```
SimonHeybrock, line 42: Is the `if` required?

src/ess/diffraction/corrections.py
```diff
@@ -0,0 +1,217 @@
...
+                                   graph={
+                                       **beamline(scatter=False),
+                                       **elastic("tof")
+                                   },
+                                   keep_inputs=False,
+                                   keep_intermediate=False,
+                                   keep_aliases=False)
+
+    if wavelength_edges is not None:
+        mon = sc.rebin(mon, 'wavelength', wavelength_edges)
```
SimonHeybrock, line 53: Can probably worry about this later, but monitors can (and will) be event mode.

src/ess/diffraction/corrections.py
```diff
@@ -0,0 +1,217 @@
...
+    Returns
+    -------
+    :
+        `data` divided by proton charge.
+    """
+    proton_charge = data.meta[proton_charge]
+    if in_place:
+        data /= proton_charge
+        return data
+    return data / proton_charge
```
SimonHeybrock: Is this the way we want to go? This function does extremely little, but requires a lot of code.

src/ess/diffraction/corrections.py
```diff
@@ -0,0 +1,217 @@
...
+    :
+        `data` normalized by `vanadium` and histogrammed according to `edges`.
+    """
+    norm = sc.lookup(sc.histogram(vanadium, bins=edges), dim=edges.dim)
+    if in_place:
+        data.bins /= norm
+        out = data
+    else:
+        out = data.bins / norm
+    return sc.histogram(out, bins=edges)
```
SimonHeybrock: Why does normalization histogram? The user may want to normalize but not histogram yet.

src/ess/diffraction/grouping.py
```diff
@@ -0,0 +1,36 @@
...
+        All events are concatenated along this dimension.
+
+    Returns
+    -------
+    :
+        `data` grouped into two_theta bins.
+    """
+    data = data.copy(deep=False)
+    if 'two_theta' not in data.meta and 'two_theta' not in data.bins.meta:
+        data.coords['two_theta'] = scn.two_theta(data)
```
SimonHeybrock: Have you considered using `transform_coords` (or the diffraction-specific helper functions you implement above) to compute this? Avoids the `if` and reliance on (the possibly soon legacy) `scn.two_theta`.

src/ess/diffraction/load.py
```diff
@@ -0,0 +1,178 @@
...
+import scipp as sc
+import scippneutron as scn
+
+from ..logging import get_logger
+from .corrections import normalize_by_proton_charge, subtract_empty_instrument
+
+
+def _load_aux_file_as_wavelength(filename: Union[str, Path]) -> sc.DataArray:
+    da = scn.load(filename,
+                  advanced_geometry=True,
```
SimonHeybrock: Note that this option can be very slow. Is it needed for getting detector IDs? If not we should try to avoid it.

src/ess/diffraction/load.py
```diff
@@ -0,0 +1,178 @@
...
+                  load_pulse_times=False,
+                  mantid_args={'LoadMonitors': True})
+    return da.transform_coords('wavelength',
+                               graph={
+                                   **scn.tof.conversions.beamline(scatter=True),
+                                   **scn.tof.conversions.elastic("tof")
+                               })
+
```
SimonHeybrock, line 28: Personal bookmark (ignore this)

SimonHeybrock (commented):
Not done yet with review, comments for first part (excluding notebooks).

src/ess/diffraction/load.py
```diff
@@ -0,0 +1,178 @@
...
+
+def _remove_based_on_proton_charge(da: sc.DataArray) -> bool:
+    charge = da.meta['proton_charge'].value.data
+    limit = da.meta['gd_prtn_chrg'].to(unit=charge.unit)
```
SimonHeybrock: Very specific names hard coded in generic diffraction module?

src/ess/diffraction/load.py
```diff
@@ -0,0 +1,178 @@
...
+    """
+    get_logger('diffraction').info('Loading vanadium from file %s.', vanadium_file)
+    vanadium = _load_aux_file_as_wavelength(vanadium_file)
+    normalize_by_proton_charge(vanadium, in_place=True)
+    empty_instrument = _load_aux_file_as_wavelength(empty_instrument_file)
+    if _remove_based_on_proton_charge(empty_instrument):
+        # TODO This is a hack because in the POWGEN test data, all events are
+        #  filtered out for the empty instrument. Ultimately, we need to properly
+        #  filter pulses by proton charge when that functionality is available.
+        return vanadium
```
SimonHeybrock: This should not be in the `diffraction` module.

Filtering by proton charge is simple, what stops us from doing it now?

src/ess/diffraction/load.py
```diff
@@ -0,0 +1,178 @@
...
+    get_logger('diffraction').info(
+        'Subtracting empty instrument loaded from file %s from vanadium.',
+        empty_instrument_file)
+    normalize_by_proton_charge(empty_instrument, in_place=True)
+    return subtract_empty_instrument(vanadium, empty_instrument)
```
SimonHeybrock: Coming back to my question on this function (above): Just switching out the edges to cover the union of the intervals seems like it can go badly wrong. If either of vanadium or empty instrument do not cover the full wavelength range then maybe there is a reason for this?

Furthermore, as soon as someone introduces additional binning, e.g., to mask a tof or wavelength range this seems like it may break very easily.

Can we find a mechanism that deals with imposing common wavelength (or TOF) binning "upstream" rather than attempting to fix it here?

src/ess/diffraction/load.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+def _as_boolean_mask(var: sc.Variable) -> sc.Variable:
+    if var.dtype in ('float32', 'float64'):
+        if sc.any(var != var.to(dtype='int64')).value:
+            raise ValueError(
+                'Cannot construct boolean mask, the input mask has fractional values.')
+    return var.to(dtype=bool)
```
SimonHeybrock: Is this for dealing with Mantid's bin masks?

src/ess/diffraction/load.py
```diff
@@ -0,0 +1,178 @@
...
+    the same Python environment as ess.
+
+    Parameters
+    ----------
+    filename:
+        The name of the calibration file to load.
+    instrument_filename:
+        Instrument definition file.
+    instrument_name:
+        Name of the instrument.
```
SimonHeybrock: Is this the route ESS wants to take? I thought scicat would deal with this? This seems to be replicating a Mantid-like interface, with its drawbacks such as multiple mutually exclusive options, ...?

src/ess/diffraction/smoothing.py
```diff
@@ -0,0 +1,82 @@
...
+            'will be returned without any!'
+            '\n--------------------------------------------------\n'
+            'If you know a good solution for handling uncertainties in such a case, '
+            'please contact the scipp developers! (e.g. via https://github.com/scipp)'
+            '\n--------------------------------------------------\n')
+        return sc.values(var)
+    return var
+
+
+def fft_smooth(da: sc.DataArray,
```
SimonHeybrock: I'd stick with the naming of scipy rather than Mantid?

src/ess/diffraction/smoothing.py
```diff
@@ -0,0 +1,82 @@
...
+               dim: str,
+               order: int,
+               Wn: sc.Variable,
+               coord: Optional[str] = None) -> sc.DataArray:
+    """
+    Smooth data using a lowpass frequency filter.
+
+    Applies a lowpass Butterworth filter to `da.data` based on the sampling rate
+    defined by `coord`.
+    See :py:func:`scipp.signal.butter` for information on filter design.
```
SimonHeybrock, line 41: Mention that this uses the midpoints?

src/ess/external/powgen/beamline.py
```diff
@@ -0,0 +1,40 @@
...
+    if not sc.identical(
+            data.coords['detector'].to(dtype=detector_info.coords['detector'].dtype,
+                                       copy=False), detector_info.coords['detector']):
+        raise sc.CoordError(
+            "The 'detector' coords of `data` and `detector_info` do not match.")
+
+    out = data.copy(deep=False)
+    del out.coords['detector']
+    # Add 1 because spectrum numbers in the data start at 1 but
+    # detector_info contains spectrum indices which start at 0.
```
SimonHeybrock, line 36: This is unfortunately not sufficient. Spectrum indices are from 0 to N-1, but spectrum numbers are arbitrary. We can, for example, load only a single bank, or a bank may be disabled.

jl-wynen: How is this represented in the inputs? Does `detector_info` change?

Do we even need to convert to spectrum for diffraction at ESS? We could just stay in 'detector', no?

SimonHeybrock: `detector_info` would be the same. In Mantid each spectrum has a set of associated detectors, so I think it gets looked up via that.

Currently I do not think that ESS would/should operate in this manner.

src/ess/diffraction/corrections.py
jl-wynen: spectrum

src/ess/diffraction/corrections.py
jl-wynen: It might also overwrite the other attrs. What do you suggest? Raise an exception if something with the same name already exists?

src/ess/diffraction/corrections.py
jl-wynen: The min and max tof values are different for vanadium and empty instrument. And therefore, the wavelengths. I don't think we can enforce the same edges.

But as we discussed, this all goes away if we don't have tof edges in the first place.

src/ess/diffraction/grouping.py
jl-wynen: This would potentially turn the position coords into attrs. But that is probably desirable because we process the position information here anyway. (And in practice, it does not matter because this function is applied to d-spacing...) Changing

src/ess/diffraction/conversions.py
jl-wynen: Done for the actual computation.
Is there a way to avoid the temporary array in `sc.all`?

src/ess/diffraction/conversions.py
jl-wynen: No. We could also preserve it as an attr. But in my tests with WISH data, I sometimes got quite close to the limits of my machine, so removing a spectrum x tof - sized coord can be relevant.

src/ess/diffraction/conversions.py
jl-wynen: Yes

src/ess/diffraction/conversions.py
jl-wynen: They are not used by the transformation. We could make them dummy inputs but then, we would require them to exist.

src/ess/diffraction/corrections.py
jl-wynen: I wanted to give the option to not rebin.

This whole function is kind of untested because it is not needed for POWGEN. So I am a bit unsure about the requirements. But since I had already written / copied it for WISH, I kept it.

src/ess/diffraction/corrections.py
jl-wynen: What do you mean? A lot of code to call? Or a lot of code because of the option to do it in-place?

I am suspecting that this function will become more complicated, depending on if the proton charge is precomputed.

src/ess/diffraction/corrections.py
jl-wynen: To ensure that the data and vanadium use the same edges. Or is it fine if they are different?

src/ess/diffraction/conversions.py
SimonHeybrock: Would it be useful if the conversion to d-spacing (with calibration) on consumed (and removed) those coords?

src/ess/diffraction/corrections.py
SimonHeybrock: Not sure. Probably?

src/ess/diffraction/conversions.py
SimonHeybrock: > Is there a way to avoid the temporary array in sc.all?

That is just one value per pixel, i.e., tiny and irrelevant?

> Done for the actual computation.

You mean for the `else` branch? But don't we frequently hit this branch in practice as well?

src/ess/diffraction/conversions.py
SimonHeybrock: ... but we have to remove them do avoid them interfering later?

src/ess/diffraction/corrections.py
SimonHeybrock: > The min and max tof values are different for vanadium and empty instrument. And therefore, the wavelengths. I don't think we can enforce the same edges.

But isn't this simply due to how we "guess" a good default value for the TOF bounds when loading data? If you impose common TOF edges from that start then at least you do not need to modify wavelength as well?

src/ess/diffraction/corrections.py
SimonHeybrock: I think my question is whether it was worth having it as a function (or whether it should just be done directly within a higher-level function), and having to worry about signatures (such as in-place options).

src/ess/diffraction/corrections.py
SimonHeybrock: Well, you normalized in event-mode. So your question is whether it is fine to later pick a binning different from the normalization-term binning for the normalized data?

I *think* that is fine, and possibly done frequently -- think also of other normalization terms such as monitor or direct-beam (SANS). These terms may vary smoothly and the concrete binning may be very different.

src/ess/diffraction/conversions.py
jl-wynen: No, I pushed a change for the then branch.

src/ess/diffraction/conversions.py
jl-wynen: This is rather tricky. We would have to use two transforms:
```.py
aux = tof_data.transform_coords('aux', graph={'aux': 'tof'}, keep_inputs=False)
aux.transform_coords('tof', graph={'tof': lambda aux, wavelength: aux}, keep_inputs=False)
```
We cannot simply use 
```.py
tof_data.transform_coords('tof', graph={'tof': lambda tof, wavelength: tof}, keep_inputs=False)
```
because that would not hit the node in the graph, it would simply promote `tof` to a coord and leave `wavelength` untouched (including the wavelength dim).

jl-wynen: Regarding handling wavelength and the positions via `transform_coords`:

We can add nodes to the graph to consume them in addition to the calibration parameters, or modify the existing node. But this would make those coords mandatory. (I experimented a bit with adding arguments to function signatures dynamically. It is possible but quite involved.)
I don't think that positions should be mandatory, users might want to not even load them if they do everything in terms of calibration parameters (e.g. in autoreduction).

We could probably pull off some shenanigans by calling `transform_coords` multiple times. But I doubt that it would make the code (or log messages) easier to understand than the current implementation.
Do you have any suggestions?

src/ess/diffraction/load.py
jl-wynen: Agree. The name is hard coded everywhere. Just like `tof` or `position`. I currently don't see a good way of abstracting away the name, though it will probably have to be adjusted for ESS.

src/ess/diffraction/load.py
jl-wynen: I don't know what those are. The masks loaded by Mantid are stored as floats but all have values of 0 or 1.

src/ess/diffraction/load.py
jl-wynen: I hope that we will take a different route. But right now, we need at least one of those options to call the mantid algorithm. We could for now hard code this information in the data array, but until we know how calibrations will be stored at ESS, we have to keep something like the current implementation.

src/ess/diffraction/smoothing.py
jl-wynen: Does it have an analogous function? I think it only has frequency filters. But that naming seems confusing here. That being said, I am happy with renaming this function.

src/ess/diffraction/load.py
SimonHeybrock: Mantid has only two types of masks: Pixel masks, and bin-masks. The latter would depend on the TOF dimension. Is that what you have?

src/ess/diffraction/smoothing.py
SimonHeybrock: Wel,, you are essentially just calling `butter`?

src/ess/diffraction/corrections.py
jl-wynen: I will worry about this when we have a data file with event mode monitors so we can test the code properly.

src/ess/diffraction/load.py
jl-wynen: It is not needed. I changed it to `False` now.
But I guess, this functions should be changed to use `load_nexus` for ESS data anyway.

src/ess/diffraction/load.py
jl-wynen: We should wait for https://github.com/scipp/scipp/pull/2477 and https://github.com/scipp/scippneutron/issues/236. Otherwise, we would have to implement something more complicated now and refactor in the hopefully near future.

src/ess/diffraction/smoothing.py
jl-wynen: But `butter_filtfilt` is not a very descriptive name. Also, I would not but 'butter' into the name because we may want to support different filters (unlikely, but still). We could use `lowpass` or `smooth_using_lowpass`.

SimonHeybrock: One of the latter two sounds good.

src/ess/diffraction/conversions.py
SimonHeybrock: > users might want to not even load them if they do everything in terms of calibration parameters 

Aren't they always needed for conversion to wavelength? Positions are not so large, I think it is unlikely that someone may need to skip loading them.

> Do you have any suggestions?

Note sure.

src/ess/diffraction/corrections.py
SimonHeybrock: iirc you said your answer was based on something that is done below, not my actual question?

src/ess/diffraction/grouping.py
SimonHeybrock: Somehow I do not see those changes, did you push them?

src/ess/diffraction/load.py
SimonHeybrock: I think, e.g., `da.meta['gd_prtn_chrg']` should probably be provided as an input to these functions.

src/ess/diffraction/load.py
SimonHeybrock: Did you mean to link scipp/scippneutron#237 instead of 236? The comment here is longer than the code to do the filtering. It is very unclear if and when scipp/scippneutron#237 will provide a "simpler" solution.

src/ess/diffraction/powder/conversions.py
```diff
@@ -0,0 +1,116 @@
...
+
+def _restore_tof_if_in_wavelength(data: sc.DataArray) -> sc.DataArray:
+    if 'wavelength' not in data.dims:
+        return data
+
+    tof_data = data.copy(deep=False)
+    del tof_data.coords['wavelength']
+    if tof_data.bins:
+        del tof_data.bins.coords['wavelength']
+    return tof_data.rename_dims({'wavelength': 'tof'})
```
SimonHeybrock: Interesting odd case: Essentially we want to `transform_coords` to `tof`, where `tof` exists (as an attr) but we nevertheless want to remove the things it *could* have been computed from.

Can we avoid this whole mess by rewriting the "conversion with calibration" based on `wavelength`? After all, it is related to `tof` by a simple (pixel-dependent) scaling factor?

src/ess/diffraction/powder/corrections.py
```diff
@@ -0,0 +1,41 @@
...
+    out = into.copy(deep=False)
+    for name in ('difa', 'difc', 'tzero'):
+        if name in out.meta:
+            raise ValueError(f"Cannot add calibration parameter '{name}' to data, "
+                             "there already is metadata with the same name.")
+        out.attrs[name] = calibration[name].data
+    if 'calibration' in out.masks:
+        raise ValueError("Cannot add calibration mask 'calibration' tp data, "
+                         "there already is a mask with the same name.")
+    out.masks['calibration'] = calibration['mask'].data
```
SimonHeybrock, line 40: So you are loading the masks now?

src/ess/diffraction/grouping.py
jl-wynen: The diff view can be finicky: https://github.com/scipp/ess/blob/d31c949a5bb696911018345b2e2b1db2c7d742cd/src/ess/diffraction/powder/conversions.py#L34

src/ess/diffraction/powder/corrections.py
jl-wynen: They were always loaded. I only added checks to prevent overwriting existing meta data.

src/ess/diffraction/powder/conversions.py
jl-wynen: We could. But I don't want to make it depend on wavelength. It should be able to bypass that. So we would have to detect whether it exists and then use a different function than when it does not exist.

SimonHeybrock: That would seem cleaner?

src/ess/diffraction/conversions.py
jl-wynen: I changed it to use `transform_coords` for handling wavelength and positions. I tried to use descriptive names so that the log message makes sense. Let me know what you think!

src/ess/diffraction/corrections.py
jl-wynen: Note that I removed this function and used division inline instead.

src/ess/diffraction/powder/conversions.py
```diff
@@ -104,13 +132,11 @@ def to_dspacing_with_calibration(
...
-        if out.bins is not None and name in out.bins.coords:
-            out.bins.attrs[name] = out.bins.coords.pop(name)
-
-    return out
+    graph = {
+        'temp_dspacing_': dspacing_from_diff_calibration,
+        'dspacing': _consume_positions,
+        'position': _make_dummy,
+        'sample_position': _make_dummy,
+        'source_position': _make_dummy
```
SimonHeybrock: I think I understand the trick you are using here, but anyone less familiar with `transform_coords` probably would not. Maybe add a comment explaining it?

src/ess/diffraction/grouping.py
jl-wynen: I forgot to make the change. Pushed it now.

src/ess/diffraction/load.py
jl-wynen: No, I meant 236. Can we currently get the pulse time reliably from the file? Without that, I don't see how we can filter pulses.

SimonHeybrock: We cannot reliably for cases that require high time resolution. The pulse times we get may be off by a couple of pulses. So if the beam is down for 5 minutes the filtering will be fine. If every other pulse is bad then it will be completely wrong.

src/ess/diffraction/load.py
jl-wynen: > Can we find a mechanism that deals with imposing common wavelength (or TOF) binning "upstream" rather than attempting to fix it here?

This would only work if the user never changes the binning or only changes it in a consistent manner.

I agree that simply replacing the binning can go wrong. But that is also the case in tof. We only handle data with a single wavelength bin here, in this case, it is fine to change the bin edges in this way, I think. I can change it to make the change in tof before conversion to wavelength. But that would be entirely equivalent and no less complicated.

SimonHeybrock: > We only handle data with a single wavelength bin here, in this case, it is fine to change the bin edges in this way, I think. I can change it to make the change in tof before conversion to wavelength. But that would be entirely equivalent and no less complicated.

That may be the case right now, but sooner or later someone will add masking along TOF, and then this will cause trouble, without necessarily showing up as an error, won't it?

src/ess/diffraction/load.py
jl-wynen: I thought the point of your work on frame handling was to figure out the correct pulses?

src/ess/diffraction/load.py
jl-wynen: `data.attrs['tof'] = tof_edges` will fail with a `DimensionError`.

I can change such that `subtract_empty_instrument` requires matching binning and move the edge matching into `load_and_preprocess_vanadium`. There, we are in full control over what happens.

src/ess/diffraction/powder/conversions.py
jl-wynen: Can do but that would only help people looking at the code. I think it is more important that the log message makes sense because that is what most people will see. E.g.
```
Transformed coords (difa, difc, position, sample_position, source_position, tof, tzero) -> (dspacing)
  Renamed dimensions:
    dspacing <- tof
  Steps:
    Compute (temp_dspacing_) = dspacing_from_diff_calibration(tof, tzero, difa, difc)
    Compute (dspacing) = _consume_positions(temp_dspacing_, position, sample_position, source_position)
```
or
```
Transformed coords (difa, difc, tof, tzero) -> (dspacing)
  Renamed dimensions:
    dspacing <- tof
  Steps:
    Compute (position) = _make_dummy()
    Compute (sample_position) = _make_dummy()
    Compute (source_position) = _make_dummy()
    Compute (temp_dspacing) = dspacing_from_diff_calibration(tof, tzero, difa, difc)
    Compute (dspacing) = _consume_positions(temp_dspacing, position, sample_position, source_position)
```
I don't think either is particularly easy to understand. I could add a custom log to explain what happens(?)

Also, is the fact that this trick needs a comment a sign that we should manually move / remove coords instead? That would not be pretty but easy to understand?

src/ess/diffraction/load.py
SimonHeybrock: Yes, so shouldn't we assume here that `pulse_time` means `pulse_time`?

src/ess/diffraction/load.py
SimonHeybrock: Yes, let us do that. Is the function even required?

src/ess/diffraction/powder/conversions.py
SimonHeybrock: Until we actually encounter data without the coords, can we just pretend they are always there? It feels like you are trying to fix hypothetical problems?

jl-wynen: Ok. But I would still not like to add them as arguments to `dspacing_from_diff_calibration` because that makes it look as though they are used to compute d-spacing.

docs/instruments/external/powgen/powgen_reduction.ipynb
```diff
@@ -0,0 +1,667 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "id": "094942ff-1bde-46c2-ae50-d00177eca3ad",
+   "metadata": {},
+   "source": [
+    "This notebook shows a basic reduction workflow for powder diffraction for the SNS [POWGEN](https://sns.gov/powgen) instrument.\n",
+    "It serves mainly to develop and present routines for powder diffraction and will eventually be removed in favor of a workflow for DREAM at ESS.\n",
+    "\n",
+    "First, we import all required modules.\n",
```
nvaytet: Delete?

docs/instruments/external/powgen/powgen_reduction.ipynb
```diff
@@ -0,0 +1,667 @@
...
+    "\n",
+    "**Note** that we load functions from `external` modules.\n",
```
nvaytet, line 19: Just add this as an inline comment of the import?

docs/instruments/external/powgen/powgen_reduction.ipynb
```diff
@@ -0,0 +1,667 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "6e5a4d3c-cb93-4875-b2be-9e73df22771d",
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "scn.instrument_view(\n",
```
nvaytet: Make this into a `powgen.instrument_view()` that would define the camera parameters?
Even if it's only a single parameter, more params could be added later (such as other beamline components etc...)

docs/instruments/external/powgen/powgen_reduction.ipynb
```diff
@@ -0,0 +1,667 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "b9494f15-1fb2-4236-95a7-800a04896e2f",
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "vana = diffraction.load_and_preprocess_vanadium(\n",
```
nvaytet: Not sure I understand why there needs to be a function which is specific to vanadium. Is it doing something different that for the sample file above? Could it just be a `load_and_preprocess` function?

docs/instruments/external/powgen/powgen_reduction.ipynb
```diff
@@ -0,0 +1,667 @@
...
+   "source": [
+    "vana"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "a8f39532-e0ee-48cc-8c49-40ac7427ded8",
+   "metadata": {},
+   "source": [
+    "Have a look at the log widget at the top of the notebook.\n",
```
nvaytet: This is reading more like a tutorial than an actual workflow.
I think it's fine to add more details about what the function is doing internally, but I think I would rephrase this.
Something like

`load_and_preprocess_vanadium` applies a crude, preliminary event filtering mechanism whereby events are discarded when the proton charge is too low (a message is visible in the log...)

?

docs/instruments/external/powgen/powgen_reduction.ipynb
```diff
@@ -0,0 +1,667 @@
...
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "57f987ff-55bb-4782-84c5-ffbbf8d8e7b6",
+   "metadata": {},
+   "source": [
+    "The data is still represented as events.\n",
+    "We could plot is as is, but we histogram manually in order to have control over the bins."
```
nvaytet: I would just remove this, as the automatic binning mechanism is going to go away in the future anyway.
I think it's fine to just have the histogramming without justification.

src/ess/diffraction/corrections.py
```diff
@@ -0,0 +1,89 @@
...
+
+from .smoothing import lowpass
+from ..logging import get_logger
+
+
+def normalize_by_monitor(data: sc.DataArray,
+                         *,
+                         monitor: str,
+                         wavelength_edges: Optional[sc.Variable] = None,
+                         smooth_args: Optional[Dict[str, Any]] = None) -> sc.DataArray:
```
nvaytet, line 16: I think this function should do what the name says, i.e. normalize, instead of normalizing and smoothing.
There should be a separate function that does the smoothing that the user can call before calling this one.

src/ess/diffraction/grouping.py
```diff
@@ -0,0 +1,34 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2022 Scipp contributors (https://github.com/scipp)
+import scipp as sc
+from scippneutron.tof import conversions
+
+
+def focus_by_two_theta(data: sc.DataArray,
```
nvaytet: Is 'focus' a Mantid specific term? Should we avoid and use e.g. `group_by_two_theta` which is more explicit as to what is going on inside the function?

SimonHeybrock: It might be powder-diffraction-specific, ask Celine.

docs/instruments/external/powgen/powgen_reduction.ipynb
jl-wynen: Only the last line or everything?

docs/instruments/external/powgen/powgen_reduction.ipynb
jl-wynen: Why? Do we have guidelines for what goes into the code and what into markdown? I typically lean towards putting as much as possible into markdown because it is easier to format.

docs/instruments/external/powgen/powgen_reduction.ipynb
jl-wynen: It subtracts the empty instrument measurement which is not done for the sample.
I am definitely not a fan of this function because it is so specific and high-level. But the only alternative I see is inlining the code in the notebook which would not help readability. Do you have a better idea?

docs/instruments/external/powgen/powgen_reduction.ipynb
jl-wynen: Are these notebooks really meant as production grade workflows? I thought those would be in `ess-notebooks` and here we explain how reduction works in principle.

src/ess/diffraction/corrections.py
jl-wynen: Then, the normalisation would be trivial, just `x.bins /= lookup(y)`. We would require the user to first convert to wavelength, then smooth, and then divide.

This is basically showing the problem of where to draw the line between tools provided by ess and what users need to write in notebooks. In this particular case, my objection to putting this into the notebook is that it is easy to make a mistake in the individual steps or forget to smooth. And ~10 lines of code in a notebook are harder to read than a single function, IMHO.

src/ess/diffraction/grouping.py
jl-wynen: Celine says, it might even be SNS-specific. I'll rename.

src/ess/diffraction/corrections.py
nvaytet: My problem is more that the function name could mislead the user to think it's just normalizing.

An alternative would be to just change the function name to reflect that it is doing more than just normalizing.

docs/instruments/external/powgen/powgen_reduction.ipynb
nvaytet: In my mind it was the other way round. Official reduction under ci control in ess, and ess-notebooks is a place for users to store and share notebooks?

docs/instruments/external/powgen/powgen_reduction.ipynb
nvaytet: The last line. I think it's obvious that we are importing modules from the code cell?

docs/instruments/external/powgen/powgen_reduction.ipynb
nvaytet: My idea was that this would go away in the future when we have a workflow for dream, and it would then be easier to update the notebook.

We don't have guidelines on this. It was just a suggestion, I don't have strong feelings either way.

jl-wynen: The entire markdown cell has to be replaced. So I'd rather leave it as it is.

src/ess/diffraction/corrections.py
jl-wynen: Not sure what would be a good name. Something like `normalize_by_smooth_monitor` does not really work because smoothing is optional.

src/ess/external/powgen/instrument_view.py
```diff
@@ -0,0 +1,40 @@
...
+    if 'camera' not in kwargs:
+        kwargs = {
+            **kwargs, 'camera': {
+                'position': sc.vector(value=[-3, 3, 3], unit=da.coords[positions].unit)
+            }
+        }
+    return scn.instrument_view(da,
+                               positions=positions,
+                               components=components,
+                               pixel_size=pixel_size,
```
nvaytet, line 39: Why did you list these 3 arguments explicitly? They seem to just be forwarded to the scippneutron `instrument_view`.
Can they just be part of the `kwargs`?
Is it because you wanted to have the type-hints? I see they are missing from the scippneutron `instrument_view`. Maybe we should add them there instead?

jl-wynen: These are the explicit arguments of `instrument_view`. Listing them explicitly means that they can be found by `help(powgen.instrument_view)`, autocompletion, etc.

src/ess/diffraction/grouping.py
nvaytet: Don't forget to rename `focus` to something else ;-)

jl-wynen:
I disabled the failing tests because this is a bug in Mantid. See https://github.com/mantidproject/mantid/issues/33690

src/ess/diffraction/corrections.py
nvaytet: ok let's keep it as is.

nvaytet (approved):

nvaytet (changes requested):
There seems to be an issue with the filtering by proton charge.
Screenshot at 2022-03-30 16-46-15

jl-wynen:
I disabled handling of the empty instrument data because I can't seem to get it to work with normalisation by proton charge. See the long comment in `load.py`. Please let me know if you have a better idea! (@celinedurniak)

celinedurniak:
@jl-wynen With the set of POWGEN data files we are using, I think the option you chose after normalisation by proton charge is fine.

docs/instruments/external/powgen/powgen_reduction.ipynb
```diff
@@ -0,0 +1,649 @@
...
+   "source": [
+    "sc.histogram(all_spectra, bins=dspacing_edges).plot()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "df5f7a70-ac69-47dc-b0c1-1139728d2878",
+   "metadata": {},
+   "source": [
+    "### Focus into $2\\theta$ bins"
```
nvaytet: We renamed 'focus' to 'group' in the python function. Should we also remove the term Focus here?

docs/instruments/external/powgen/powgen_reduction.ipynb
```diff
@@ -0,0 +1,649 @@
...
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "two_theta = sc.linspace(dim='two_theta',\n",
+    "                        unit='deg',\n",
+    "                        start=25.0,\n",
+    "                        stop=90.0,\n",
+    "                        num=16)\n",
+    "sample_focussed = diffraction.group_by_two_theta(sample_dspacing,\n",
```
nvaytet:
```suggestion
    "sample_grouped = diffraction.group_by_two_theta(sample_dspacing,\n",
```

docs/instruments/external/powgen/powgen_reduction.ipynb
```diff
@@ -0,0 +1,649 @@
...
+    "                                                 edges=two_theta)\n",
+    "vana_focussed = diffraction.group_by_two_theta(vana_dspacing, edges=two_theta)"
```
nvaytet:
```suggestion
    "vana_grouped = diffraction.group_by_two_theta(vana_dspacing, edges=two_theta)"
```

docs/instruments/external/powgen/powgen_reduction.ipynb
```diff
@@ -0,0 +1,649 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "5ef778f1-e23a-492f-9a08-9a08e889ad83",
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "sc.histogram(all_spectra, bins=dspacing_edges).plot()"
```
nvaytet, line 489: Why the strange unit on the y axis?
Screenshot at 2022-04-25 13-51-38

docs/instruments/external/powgen/powgen_reduction.ipynb
```diff
@@ -0,0 +1,649 @@
...
+   "source": [
+    "### Focus into $2\\theta$ bins"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "fb474efb-692f-4f2d-9243-23ff9319dd76",
+   "metadata": {},
+   "source": [
+    "For a better resolution, we now group ('focus') the sample and vanadium data into a number of bins in the scattering angle [$2\\theta$](https://scipp.github.io/scippneutron/user-guide/coordinate-transformations.html)\n",
```
nvaytet: Something seems to go wrong in sphinx with the rendering of math in a link?
Screenshot at 2022-04-25 13-51-44

docs/instruments/external/powgen/powgen_reduction.ipynb
jl-wynen: Don't know. Must be some strange scaling in the unit from dividing them for normalisation.

jl-wynen: The sample and vanadium have different units for the proton charge. When dividing them, we get the 2.7e-10 scale in the plot. I changed `normalize_by_vanadium` to convert to `unit='one'` to eliminate the scale.

docs/instruments/external/powgen/powgen_reduction.ipynb
jl-wynen: Good catch!

docs/instruments/external/powgen/powgen_reduction.ipynb
jl-wynen: Removed all mentions of 'focus'

docs/instruments/external/powgen/powgen_reduction.ipynb
```diff
@@ -0,0 +1,650 @@
...
+   "source": [
+    "### Group into $2\\theta$ bins"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "fb474efb-692f-4f2d-9243-23ff9319dd76",
+   "metadata": {},
+   "source": [
+    "For a better resolution, we now group the sample and vanadium data into a number of bins in the scattering angle $2\\theta$ (https://scipp.github.io/scippneutron/user-guide/coordinate-transformations.html)\n",
```
nvaytet:
```suggestion
    "For a better resolution, we now group the sample and vanadium data into a number of bins in the scattering angle $2\\theta$ (see [here](https://scipp.github.io/scippneutron/user-guide/coordinate-transformations.html))\n",
```

nvaytet (commented):
Looking good. Just one more suggestion for hiding the long url.

In addition, I don't see anything in the release notes.
Can you add something?

Thanks

nvaytet (approved):

PR #115, opened 2022-04-04
File writing and the Orso object

arm61:
Add necessary functionality and an example of file writing, calculation of resolution function and the including of the orso object that is populated throughout


arm61:
How do I add a `pip` package? `orsopy` is only on `pip` currently.

arm61:
For now (to get things to build) I have added a `wget` and install for `orsopy` but that is obviously not ideal. I couldn't work out how to include `pip` dependencies.

conda/meta.yaml
```diff
@@ -48,6 +48,8 @@ build:
   script:
     - wget https://files.pythonhosted.org/packages/04/51/4b4a0740fbde50408a2f956cbb190bdc01f1c7a562627e7f102b6a30d9b4/sphinx_autodoc_typehints-1.17.0-py3-none-any.whl
     - pip install sphinx_autodoc_typehints-1.17.0-py3-none-any.whl
+    - wget https://files.pythonhosted.org/packages/9d/b4/761a6da7e1dedf793a6c3410cf362e2b61092a8fd8267d5ebc88e71f1fc5/orsopy-0.0.5-py2.py3-none-any.whl
+    - pip install orsopy-0.0.5-py2.py3-none-any.whl
```
nvaytet: This is fine :+1:

docs/instruments/amor/amor_reduction.ipynb
```diff
@@ -135,8 +140,53 @@
   },
   {
    "cell_type": "markdown",
-   "id": "9a43dee7-fea3-4de9-9426-d3fe9741249d",
+   "id": "d5b8ab01-4924-451e-95fb-a04365373c54",
+   "metadata": {},
+   "source": [
+    "The resolution function for these measurements involves combining a few contributing elements. \n",
```
nvaytet: I feel we start talking about resolution function a bit out of the blue. Maybe making a new notebook subsection would help?

docs/instruments/amor/amor_reduction.ipynb
```diff
@@ -135,8 +140,53 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "87829723-a265-4a4d-ba8e-ecd903eb4197",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "sample.coords['sample_size_resolution'] = sc.scalar(0.18) * sc.to_unit(sample.coords['sample_size'], 'm') / sample.coords['position'].fields.z"
```
nvaytet: Where does the `0.18` come from?

docs/instruments/amor/amor_reduction.ipynb
```diff
@@ -108,9 +109,13 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "sample = amor.load(amor.data.get_path(\"sample.nxs\"),\n",
+    "owner = fileio.base.Person('Jochen Stahn', 'Paul Scherrer Institut', 'jochen.stahn@psi.ch')\n",
+    "sample = fileio.data_source.Sample('Ni/Ti Multilayer', 'gas/solid', 'air | (Ni | Ti) * 5 | Si')\n",
+    "creator = fileio.base.Person('Andrew R. McCluskey', 'European Spallation Source', 'andrew.mccluskey@ess.eu')\n",
+    "\n",
+    "sample = amor.load(amor.data.get_path(\"sample.nxs\"), owner, sample, creator, 'amor_reduction.ipynb', \n",
```
nvaytet: Use keyword arguments?

src/ess/reflectometry/save.py
```diff
@@ -0,0 +1,24 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2022 Scipp contributors (https://github.com/scipp)
+
+import numpy as np
+import scipp as sc
+from orsopy import fileio
```
nvaytet: You are making `orso` a hard dependency of `ess.reflectometry`. Is that what we want to do, or do we want to have it optional?

src/ess/reflectometry/save.py
```diff
@@ -0,0 +1,24 @@
...
+
+    :param data_array: Scipp-data array to save.
+    :param filename: Filename.
+    """
+    if filename[:-4] == '.ort':
+        raise UserWarning("The expected output file ending is .ort.")
+    q = sc.midpoints(data_array.mean('detector_id').coords['Q']).values
+    R = data_array.mean('detector_id').data.values
+    sR = sc.stddevs(data_array.mean('detector_id').data).values
+    sq = data_array.coords['sigma_Q_by_Q'].max('detector_id').values * q
```
nvaytet: I would perform the operation on the scipp variables, and extract the `.values` afterwards, as you are dropping variances and also ignoring potentially mismatching units.

docs/instruments/amor/amor_reduction.ipynb
```diff
@@ -567,8 +653,72 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "eb2ef19e-2388-40f6-91b2-bb58b83c3246",
    "metadata": {},
+   "outputs": [],
+   "source": [
+    "normalized.coords['sigma_Q_by_Q'] = sc.sqrt(\n",
```
nvaytet: How is the `sigma_Q` used in the analysis? Is it related to the variances of the `sample_q` quantity, or is it a separate quantity?
Should the variances of `sample_q` contribute to `sigma_Q`?

src/ess/amor/beamline.py
```diff
@@ -46,6 +49,8 @@ def make_beamline(sample_rotation: sc.Variable,
         detector_spatial_resolution = 0.0025 * sc.units.m
     if gravity is None:
         gravity = sc.vector(value=[0, -1, 0]) * g
+    if chopper_chopper_distance is None:
+        chopper_chopper_distance = sc.scalar(0.49, unit='m')
```
nvaytet: Remove parameter and compute distance from chopper positions.

src/ess/amor/load.py
```diff
@@ -63,5 +76,44 @@ def load(filename,
...
+    data.attrs['orso'].value.data_source.experiment.instrument = data.attrs[
+        'instrument_name'].value
+    data.attrs['orso'].value.data_source.experiment.start_date = datetime.strftime(
+        datetime.strptime(data.attrs['start_time'].value[:-3], '%Y-%m-%dT%H:%M:%S.%f'),
+        '%Y-%m-%d')
+    data.attrs['orso'].value.data_source.sample = sample
+    data.attrs['orso'].value.data_source.measurement.data_files = [filename]
+    data.attrs['orso'].value.reduction.creator = creator
+    data.attrs['orso'].value.reduction.script = reduction_script
+
```
nvaytet: Move into `populate_orso()`

nvaytet (commented):
Some more comments, unrelated to our discussion on wrapping code in cells into functions in the `reflectometry` or `amor` modules.

conda/meta.yaml
arm61: This means if someone wants to use `ess.reflectometry` they need to `conda install ess` then `pip install orsopy` yeah?

src/ess/reflectometry/save.py
arm61: I think it should be a hard dependency. If you wanna write files, you need `orsopy`.

nvaytet: Sure, but maybe you want to simply load an amor file and make a plot?
I am worried that we will have 200 different dependencies for all the different techniques. I think we should make it optional? I will ask Simon what he thinks.

conda/meta.yaml
nvaytet: Well we would add `orsopy` in the list of dependencies in the recommended env file for using `ess`

jl-wynen: You should be able to install directly from the remote URL with pip:
```
pip install https://files.pythonhosted.org/packages/9d/b4/761a6da7e1dedf793a6c3410cf362e2b61092a8fd8267d5ebc88e71f1fc5/orsopy-0.0.5-py2.py3-none-any.whl
```

See, e.g., point 8 in https://pip.pypa.io/en/stable/cli/pip_install/

src/ess/amor/load.py
```diff
@@ -30,12 +34,21 @@ def _tof_correction(data: sc.DataArray, dim: str = 'tof') -> sc.DataArray:
 
 
 def load(filename,
+         owner: fileio.base.Person,
+         sample: fileio.data_source.Sample,
+         creator: fileio.base.Person,
+         reduction_script: str,
```
jl-wynen, line 52: Can the ORSO related arguments be combined into a single structure? This might simplify things moving forward in case you need to add more arguments or if they are loaded from a common place (e.g. SciCat)

src/ess/reflectometry/save.py
jl-wynen: Does `save.py` have to be imported by `__init__.py`?

src/ess/reflectometry/save.py
```diff
@@ -0,0 +1,25 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2022 Scipp contributors (https://github.com/scipp)
+
+import numpy as np
+import scipp as sc
+from orsopy import fileio
+
+
+def save(data_array: sc.DataArray, filename: str):
```
jl-wynen:
```suggestion
def save_orso(data_array: sc.DataArray, filename: str):
```
?

src/ess/reflectometry/save.py
```diff
@@ -0,0 +1,25 @@
...
+    """
+    Save a data array with the ORSO .ort file format.
+
+    :param data_array: Scipp-data array to save.
+    :param filename: Filename.
+    """
+    if filename[:-4] == '.ort':
+        raise UserWarning("The expected output file ending is .ort.")
+    q = sc.midpoints(data_array.mean('detector_id').coords['Q'])
```
jl-wynen: You should check if Q is actually bin-edges.

src/ess/reflectometry/save.py
arm61: Fair point, does this maybe suggest at the `ess` level there should be different dependency levels/categories?

conda/meta.yaml
arm61: I tried this (and maybe did it wrong) but then found something saying that this didn't work for conda envs

jl-wynen: ok

docs/instruments/amor/amor_reduction.ipynb
arm61: The `sigma_Q` is the resolution of each `Q` points so not linked to the variances in the `sample_q`.

src/ess/reflectometry/save.py
arm61: What as the conclusion on how to achieve this?

src/ess/reflectometry/save.py
arm61: I am going for `save_ort` cause in future there might be a NeXus version of the ORSO format.

src/ess/amor/load.py
arm61: In future, `owner` and `sample` will be in the raw NeXus, so these will be removed when this is the case.

src/ess/reflectometry/save.py
jl-wynen: Use this? https://scipp.github.io/generated/classes/scipp.Coords.html#scipp.Coords.is_edges

src/ess/reflectometry/conversions.py
```diff
@@ -61,3 +61,28 @@ def specular_reflection() -> dict:
     graph["theta"] = theta
     graph["Q"] = reflectometry_q
     return graph
+
+
+def tof_to_wavelength(data_array: sc.DataArray,
+                      wavelength_edges: sc.Variable,
+                      graph: dict = specular_reflection()) -> sc.DataArray:
```
jl-wynen:
```suggestion
                      graph: dict = None) -> sc.DataArray:
```
And later
```.py
graph = graph if graph is not None else specular_reflection()
```
in order to avoid a mutable default argument.

src/ess/reflectometry/save.py
arm61: This is new?

conda/meta.yaml
arm61: Sorry @jl-wynen that comment was to @nvaytet not yourself

arm61:
Screenshot 2022-04-06 at 15 22 56

`yapf` is rewriting my file and making it fail `flake8` 😢

arm61:
Looks like `yapf` and `flake8` are playing nicely again.

src/ess/reflectometry/save.py
jl-wynen: Added in 0.13

arm61:
Requesting a re-view from you both writing tests in the meantime (though those can be put in a future PR).

docs/techniques/reflectometry/reference.rst
```diff
@@ -28,3 +28,11 @@ Resolution
...
    resolution.detector_resolution
    resolution.z_offset
+
+Save 
+----
+
+.. autosummary:: 
+   :toctree: ../../generated
+   
+   save.save
```
jl-wynen:
```suggestion
   save.save_ort
```
Please build the docs and check for warnings / errors.

src/ess/amor/normalise.py
```diff
@@ -0,0 +1,30 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2022 Scipp contributors (https://github.com/scipp)
+import scipp as sc
+
+
+def normalise_by_supermirror(sample: sc.DataArray,
```
jl-wynen: I believe this should be 'normalize' (also rename file) because we decided to use American spelling.

jl-wynen (commented):
We switched to using numpy-style docstrings because they are more readable (https://github.com/scipp/scipp/pull/2432). sphinx in `ess` uses the napoleon extension to make this work.
I'm not asking you to change the docstrings now, but when you add more functions in the future, you can use numpy-style and step by step change the existing ones.

src/ess/amor/normalise.py
arm61: American spelling makes me sad 🤣

arm61:
> We switched to using numpy-style docstrings because they are more readable ([scipp/scipp#2432](https://github.com/scipp/scipp/pull/2432)). sphinx in `ess` uses the napoleon extension to make this work.
> I'm not asking you to change the docstrings now, but when you add more functions in the future, you can use numpy-style and step by step change the existing ones.

I used to use `numpy` style and now my brain is trained on `sphinx` style 🤣 . Will adapt gradually!

jl-wynen (approved):

src/ess/amor/beamline.py
```diff
@@ -63,10 +68,14 @@ def make_beamline(sample_rotation: sc.Variable,
...
     # We should change this to be a Dataset, which is what we do here.
-    beamline["source_chopper"] = sc.scalar(
+    beamline["source_chopper_2"] = sc.scalar(
         make_chopper(frequency=chopper_frequency,
                      phase=chopper_phase,
-                     position=chopper_position))
+                     position=chopper_2_position))
+    beamline["source_chopper_1"] = sc.scalar(
+        make_chopper(frequency=chopper_frequency,
+                     phase=chopper_phase,
```
nvaytet, line 82: Are the phases the same for the two choppers?
I don't think it affects the workflow here but i'm slightly surprised.
Or maybe it's the cut-out angles that are different, thus making the choppers optically blind?

src/ess/amor/calibrations.py
```diff
@@ -4,26 +4,52 @@
 
 
 def supermirror_calibration(
-        coord: sc.Variable,
-        is_bin_edge: bool = True,
-        m_value: sc.Variable = sc.scalar(5, unit=sc.units.dimensionless),
-        critical_edge: sc.Variable = 0.022 * sc.Unit('1/angstrom'),
-        alpha: sc.Variable = 0.25 / 0.088 * sc.units.angstrom) -> sc.Variable:
+    data_array: sc.DataArray,
+    m_value: sc.Variable = sc.scalar(5, unit=sc.units.dimensionless),
```
nvaytet: Should we also use `None` and set the default values in the function body?

src/ess/amor/load.py
```diff
@@ -63,5 +76,63 @@ def load(filename,
     for key, value in beamline.items():
         data.coords[key] = value
 
+    data.attrs['orso'] = sc.scalar(
+        populate_orso(data, filename, owner, sample, creator, reduction_script))
```
nvaytet: Use keyword args?

src/ess/amor/load.py
```diff
@@ -63,5 +76,63 @@ def load(filename,
...
+
     # Perform tof correction and fold two pulses
     return _tof_correction(data)
+
+
+def populate_orso(data: sc.DataArray, filename: str, owner: fileio.base.Person,
```
nvaytet, line 116: Does the Orso standard say that all these entries are mandatory, or are some of them optional?

src/ess/amor/normalize.py
```diff
@@ -0,0 +1,30 @@
...
+    try:
+        normalized.attrs['orso'] = sample.attrs['orso']
+        normalized.attrs['orso'].value.reduction.corrections = list(
+            set(sample.attrs['orso'].value.reduction.corrections +
+                supermirror.attrs['orso'].value.reduction.corrections))
+        normalized.attrs[
+            'orso'].value.data_source.measurement.reference = supermirror.attrs[
+                'orso'].value.data_source.measurement.data_files
+    except KeyError:
+        raise UserWarning("For metadata to be logged in the data array, "
```
nvaytet: By using `raise` you are generating an actual error, instead of a warning. You should use `warnings.warn()` if you don't want the workflow to be interrupted if orsopy is not installed.

src/ess/amor/beamline.py
```diff
@@ -15,7 +15,8 @@ def make_beamline(sample_rotation: sc.Variable,
                   gravity: sc.Variable = None,
                   chopper_frequency: sc.Variable = None,
                   chopper_phase: sc.Variable = None,
-                  chopper_position: sc.Variable = None) -> dict:
+                  chopper_1_position: sc.Variable = None,
```
nvaytet, line 18: The names `chopper_1` and `chopper_2` end up in many places of the code. In the WFM, I tried to use names that were a little more descriptive, such as `chopper_near` and `chopper_far`.
However, I realised that it could also be unclear, because you could ask "does `chopper_near` mean near the source or near the sample?"
So just a suggestion. I guess `chopper_1` could be interpreted as being the first chopper the neutrons meet along their flight path, so maybe it's just as good?

src/ess/amor/resolution.py
```diff
@@ -0,0 +1,68 @@
...
+# Copyright (c) 2022 Scipp contributors (https://github.com/scipp)
+import scipp as sc
+from .tools import fwhm_to_std
+
+
+def wavelength_resolution(chopper_1_position: sc.Variable,
+                          chopper_2_position: sc.Variable,
+                          pixel_position: sc.Variable) -> sc.Variable:
+    """
+    Find the wavelength resolution contribution as described in Secion 4.3.3 of the Amor
```
nvaytet:
```suggestion
    Find the wavelength resolution contribution as described in Section 4.3.3 of the Amor
```

src/ess/amor/resolution.py
```diff
@@ -0,0 +1,68 @@
...
+    chopper_midpoint = (chopper_1_position.data +
+                        chopper_2_position.data) * sc.scalar(0.5)
+    chopper_detector_distance = (pixel_position.fields.z - chopper_midpoint.fields.z)
+    return fwhm_to_std(distance_between_choppers / chopper_detector_distance)
+
+
+def sample_size_resolution(pixel_position: sc.Variable,
+                           sample_size: sc.Variable) -> sc.Variable:
+    """
+    The resolution from the projected sample size, where is may be bigger
```
nvaytet: "where is may be bigger" ?

src/ess/amor/resolution.py
```diff
@@ -0,0 +1,68 @@
...
+    than the detector pixel resolution as described in Secion 4.3.3 of the Amor
```
nvaytet, line 40:
```suggestion
    than the detector pixel resolution as described in Section 4.3.3 of the Amor
```

src/ess/amor/resolution.py
```diff
@@ -0,0 +1,68 @@
...
+    :param sample_size: Size of sample.
+    :return: Standard deviation of contribution from the sample size.
+    """
+    return fwhm_to_std(sc.to_unit(sample_size, 'm') / pixel_position.fields.z)
+
+
+def angular_resolution(pixel_position: sc.Variable, theta: sc.Variable,
+                       detector_spatial_resolution: sc.Variable) -> sc.Variable:
+    """
+    Determine the angular resolution as described in Secion 4.3.3 of the Amor
```
nvaytet:
```suggestion
    Determine the angular resolution as described in Section 4.3.3 of the Amor
```
(the wonders of copy/paste...)

src/ess/amor/resolution.py
```diff
@@ -0,0 +1,68 @@
...
+    """
+    The resolution from the projected sample size, where is may be bigger
+    than the detector pixel resolution as described in Secion 4.3.3 of the Amor
+    publication (doi: 10.1016/j.nima.2016.03.007).
+
+    :param pixel_position: Positions for detector pixels.
+    :param sample_size: Size of sample.
+    :return: Standard deviation of contribution from the sample size.
+    """
+    return fwhm_to_std(sc.to_unit(sample_size, 'm') / pixel_position.fields.z)
```
nvaytet: Should we also convert the `pixel_position.fields.z` to `m` ? to make sure they have the same unit?

src/ess/amor/resolution.py
```diff
@@ -0,0 +1,68 @@
...
+    """
+    Determine the angular resolution as described in Secion 4.3.3 of the Amor
+    publication (doi: 10.1016/j.nima.2016.03.007).
+
+    :param pixel_position: Positions for detector pixels.
+    :param theta: Theta values for events.
+    :param detector_spatial_resolution: FWHM of detector pixel resolution.
+    """
+    return fwhm_to_std(sc.atan(
+        detector_spatial_resolution / pixel_position.fields.z)) / theta
```
nvaytet: This will throw an error if ` detector_spatial_resolution` and `pixel_position.fields.z` don't have the same unit. e.g. `cm` and `m`. Should we convert both to e.g. `m` as you've done for `sample_size` above?
Should we also convert the result of the `atan` to the unit of `theta`? Since we recently discovered that dividing `deg` by `rad` does weird things with the unit...

src/ess/reflectometry/__init__.py
```diff
@@ -3,4 +3,4 @@
 # flake8 : noqa: F401
 from . import corrections
 from . import conversions
-from . import resolution
+from . import save
```
nvaytet: Should we make an `io` module that would also contain loaders? Would you ever need to read from an orso file to get info into a notebook? (not suggesting to implement an orso reader now, but might be good to think about where it would go if we need one in the future?)

src/ess/reflectometry/conversions.py
```diff
@@ -61,3 +96,145 @@ def specular_reflection() -> dict:
...
+        # angstrom in the ORSO object.
+        if unit == 'angstrom':
+            unit = 'angstrom'
+        data_array_wav.attrs[
+            'orso'].value.data_source.measurement.instrument_settings.wavelength = (
+                fileio.base.ValueRange(data_array_wav.coords['wavelength'].min().value,
+                                       data_array_wav.coords['wavelength'].max().value,
+                                       unit))
+    except ImportError:
+        raise UserWarning("For metadata to be logged in the data array, "
```
nvaytet: replace `raise` with `warnings.warn`

src/ess/reflectometry/conversions.py
```diff
@@ -61,3 +96,145 @@ def specular_reflection() -> dict:
...
+
+    Returns
+    -------
+    :
+        New data array with theta coordinate.
+    """
+    graph = graph if graph is not None else specular_reflection()
+    data_array_q = data_array.transform_coords(["Q"], graph=graph)
+    if q_edges is not None:
+        data_array_q = sc.bin(data_array_q, edges=[q_edges])
```
nvaytet, line 217: You are not logging to orso here. Is it because the Q values are the final data form you are saving to the orso file anyway, so you don't need to log the range of values?

src/ess/reflectometry/corrections.py
```diff
@@ -1,35 +1,80 @@
...
+                                            data_array.bins.coords['theta'])
+    footprint_scale = sc.erf(
+        fwhm_to_std(data_array.coords['sample_size'] / size_of_beam_on_sample))
+    data_array_fp_correction = data_array / footprint_scale.squeeze()
+    try:
+        data_array_fp_correction.attrs['orso'].value.reduction.corrections += [
+            'footprint correction'
+        ]
+    except KeyError:
+        warnings.warn(
```
nvaytet: This comes up a lot, can you make it a single function somewhere and call that?
Something like 
```Py
def check_orso():
    try:
        from orsopy import fileio
    except ImportError:
        warnings.warn("To store information about corrections it is "
            "necessary to install the orsopy package.", UserWarning)
        return False
    return True

...
if check_orso():
     data_array_fp_correction.attrs['orso'].value.reduction.corrections += [
            'footprint correction'
        ]
```
I think `check_orso()` should then probably also be used when you are first creating the orso attr while loading the files?

src/ess/reflectometry/corrections.py
```diff
@@ -1,35 +1,80 @@
...
+
+
+def normalize_by_counts(data_array: sc.DataArray) -> sc.DataArray:
     """
-    beam_on_sample = beam_size / sc.sin(theta)
-    fwhm_to_std = 2 * np.sqrt(2 * np.log(2))
-    return sc.erf(sample_size / beam_on_sample * fwhm_to_std)
+    Normalize the bin-summed data by the total number of counts.
 
+    Parmaeters
```
nvaytet:
```suggestion
    Parameters
```

src/ess/reflectometry/save.py
```diff
@@ -0,0 +1,31 @@
...
+    Parameters
+    ----------
+    data_array:
+        Scipp-data array to save.
+    filename:
+        Filename.
+    """
+    from orsopy import fileio
+    if filename[:-4] == '.ort':
+        raise UserWarning("The expected output file ending is .ort.")
```
nvaytet:
```suggestion
        raise ValueError("The expected output file ending is .ort.")
```
?

src/ess/reflectometry/save.py
```diff
@@ -0,0 +1,31 @@
...
+    q = data_array.mean('detector_id').coords['Q']
```
nvaytet: Will the dimension always be `detector_id`? Or should we make it a function arg that defaults to `detector_id`?
`detector_id` could be Amor-specific, while this lives in the `reflectometry` module.

docs/instruments/amor/amor_reduction.ipynb
```diff
@@ -311,82 +321,118 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "149c05ea-b5b5-44f1-9b1b-177d79ad1b70",
    "metadata": {},
+   "outputs": [],
    "source": [
-    "This footprint scale factor accounts for the fact that the illuminated area of the sample depends on the angle of incidence (which as we noted previously may be different for the sample and the reference)."
+    "sample_theta.coords['wavelength_resolution'] = amor.resolution.wavelength_resolution(\n",
+    "    sample.coords['source_chopper_1'].value['position'], sample.coords['source_chopper_2'].value['position'], sample.coords['position'])"
```
nvaytet: Use keyword args?

docs/instruments/amor/amor_reduction.ipynb
```diff
@@ -311,82 +321,118 @@
...
    "cell_type": "code",
    "execution_count": null,
-   "id": "3583f74f-b439-4944-9094-55643ad5eb4e",
+   "id": "7fb63c73-3507-4e87-b2f4-6882d49f43ca",
    "metadata": {},
    "outputs": [],
    "source": [
-    "sample_theta /= footprint_scale.squeeze()"
+    "sample_theta.bins.coords['angular_resolution'] = amor.resolution.angular_resolution(\n",
+    "    sample.coords['position'], sample_theta.bins.coords['theta'], sample_theta.coords['detector_spatial_resolution'])"
```
nvaytet: kwargs?

docs/instruments/amor/amor_reduction.ipynb
```diff
@@ -311,82 +321,118 @@
...
-   "id": "4e00af16-f26f-4df0-b478-d4ef0cc51b18",
+   "id": "87829723-a265-4a4d-ba8e-ecd903eb4197",
    "metadata": {},
    "outputs": [],
    "source": [
-    "reference_theta = reference_wav.transform_coords(['theta'], graph=graph)\n",
-    "beam_on_sample = reference_theta.coords['beam_size'] / sc.sin(reference_theta.bins.coords['theta'])\n",
-    "footprint_scale = sc.erf(amor.tools.fwhm_to_std(reference_theta.coords['sample_size'] / beam_on_sample))\n",
-    "reference_theta /= footprint_scale.squeeze()"
+    "sample_theta.coords['sample_size_resolution'] = amor.resolution.sample_size_resolution(sample.coords['position'], sample.coords['sample_size'])"
```
nvaytet: kwargs and split the long line?

nvaytet: So are the resolution computations Amor specific or could they live in the `reflectometry` module?

nvaytet (changes requested):
I think the notebook looks much cleaner now. Nice job! :+1: 
Some more comments, should be the final round of changes.

conda/meta.yaml
nvaytet: In this file https://github.com/scipp/ess/blob/main/docs/environments/ess.yml, add a section with
```
  - pip
  - pip:
    - orsopy
```

src/ess/reflectometry/corrections.py
nvaytet: Now that I read the code again, this wouldn't be so nice in the cases where you are doing
```Py
    try:
        from orsopy import fileio
        data_array_theta.attrs[
            'orso'].value.data_source.measurement.instrument_settings.incident_angle = (
                fileio.base.ValueRange(......
```
because changing it to
```Py
    if check_orso():
        from orsopy import fileio
        data_array_theta.attrs[
            'orso'].value.data_source.measurement.instrument_settings.incident_angle = (
                fileio.base.ValueRange(
```
would mean importing `orsopy` twice...

You could:
1. Simply make a function that makes the common error message, and keep the rest as is
```Py
def orso_not_found_error():
    warnings.warn(
            "To store information about corrections it is "
            "necessary to install the orsopy package.", UserWarning)

    try:
    ...
    except ImportError:
        orso_not_found_error()
```

2. Something like this, but I find it kind of dodgy/hacky
```Py
def check_orso():
    try:
        from orsopy import fileio
        return fileio
    except ImportError:
        warnings.warn("To store information about corrections it is "
            "necessary to install the orsopy package.", UserWarning)
        return False

...
if (orso_fileio := check_orso()):
     data_array_theta.attrs[
            'orso'].value.data_source.measurement.instrument_settings.incident_angle = (
                orso_fileio.base.ValueRange(
```

src/ess/amor/beamline.py
arm61: This chopper is an optically blind system achieved with two phase matched choppers that are 1 m apart.

src/ess/amor/normalize.py
arm61: Yes. I started correcting this but is it still in progress cause I got distracted by shiny things probably.

docs/instruments/amor/amor_reduction.ipynb
arm61: I expect them to be Amor specific.

src/ess/amor/beamline.py
arm61: I didn't really like `_1` and `_2` but I agree it is difficult to have something perfect. I am happy to go with a standard that we invent?

src/ess/reflectometry/conversions.py
arm61: Essentially yes, there it nothing in this conversion that is an essential or optional input in the ORSO object.

src/ess/reflectometry/__init__.py
arm61: This is a good idea cause in future there might be a function to read in ort as you say.

src/ess/amor/load.py
arm61: There are all mandatory in the ORSO format.

src/ess/amor/load.py
arm61: Do you mean like passing a dict to `load`, cause that seems weird if all the inputs are requirements.

nvaytet: No i meant 
```Py
populate_orso(data=data, filename=filename, owner=owner, sample=sample, creator=creator, reduction_script=reduction_script))
```

This prevents someone in the future just swapping round to arguments by mistake and the error going unnoticed until you inspect the metadata in the file and you find out the owner has not become the sample...

arm61: Ah gotcha

conda/meta.yaml
arm61: Yeah I tried that in b5088c9e95aac765e45011fb4b9912484adf2ab6 and the build failed because `ValueError: Dictionaries are not supported as values in requirements sections.  Note that pip requirements as used in conda-env environment.yml files are not supported by conda-build.`
[31](https://github.com/scipp/ess/runs/5828761443?check_suite_focus=true#step:5:31)

src/ess/amor/resolution.py
```diff
@@ -28,41 +36,73 @@ def wavelength_resolution(chopper_1_position: sc.Variable,
...
+    :
+        Angular resolution standard deviation
     """
-    return fwhm_to_std(sc.atan(
-        detector_spatial_resolution / pixel_position.fields.z)) / theta
+    return fwhm_to_std(
+        sc.to_unit(
+            sc.atan(
+                sc.to_unit(detector_spatial_resolution, 'm') /
+                sc.to_unit(pixel_position.fields.z, 'm')), theta.bins.unit)) / theta
```
nvaytet: With `theta.bins.unit`, this function can only be used for event data. Maybe someone would want to use it on histogrammed data?
We could do something like
```Py
angle_unit = theta.bins.unit if theta.bins is not None else theta.unit
```
?

In addition, we should probably add a `copy=False` to all our `to_unit` calls, because calling this on event data will make many copies of potentially very large data.

src/ess/amor/calibrations.py
```diff
@@ -1,52 +1,78 @@
...
+        Supermirror alpha value.
 
-    :return: Calibrated supermirror measurement.
+    Returns
+    -------
+    :
+        Calibrated supermirror measurement.
     """
+    if m_value is None:
+        m_value = sc.scalar(5, unit=sc.units.dimensionless)
```
nvaytet, line 31: Now that we don't have the default values in the function call, maybe it would be worth mentioning the default values in the docstring?

src/ess/amor/resolution.py
arm61: Also added `copy=False` to the `to_unit` that is on the pixels but not for the sample size or detector resolution cause these will be single values.

nvaytet (approved):

PR #117, opened 2022-04-11
Prepare v0.5.0

nvaytet:
Some changes to the amor notebook to make `orsopy` a soft dependency

### DIFF
--- a/docs/about/release-notes.rst
+++ b/docs/about/release-notes.rst
@@ -9,7 +9,8 @@ v0.5.0 (April 2022)
 Features
 ~~~~~~~~
 
-* Added resolution function for Amor and support for Orso file format in the ``reflectometry`` module `#115 <https://github.com/scipp/ess/pull/115>`_.
+* Added resolution function for Amor and support for Orso file format in the ``reflectometry`` module `#115 <https://github.com/scipp/ess/pull/115>`_ (*Reflectometry*).
+* Added footprint correction, super-mirror calibration, and pixel-dependent normalization in the Amor workflow `#97 <https://github.com/scipp/ess/pull/97>`_ (*Reflectometry*).
 
 Contributors
 ~~~~~~~~~~~~
--- a/docs/about/release-notes.rst
+++ b/docs/about/release-notes.rst
@@ -3,6 +3,22 @@
 Release Notes
 =============
 
+v0.5.0 (April 2022)
+-------------------
+
+Features
+~~~~~~~~
+
+* Added resolution function for Amor and support for Orso file format in the ``reflectometry`` module `#115 <https://github.com/scipp/ess/pull/115>`_.
+
+Contributors
+~~~~~~~~~~~~
+
+Simon Heybrock :sup:`a`\ ,
+Andrew McCluskey :sup:`a`\ ,
+Neil Vaytet :sup:`a`\ ,
+and Jan-Lukas Wynen :sup:`a`
+
 v0.4.0 (February 2022)
 ----------------------
 
--- a/docs/about/release-notes.rst
+++ b/docs/about/release-notes.rst
@@ -10,7 +10,7 @@ Features
 ~~~~~~~~
 
 * Added resolution function for Amor and support for Orso file format in the ``reflectometry`` module `#115 <https://github.com/scipp/ess/pull/115>`_ (*Reflectometry*).
-* Added footprint correction, super-mirror calibration, and pixel-dependent normalization in the Amor workflow `#97 <https://github.com/scipp/ess/pull/97>`_ (*Reflectometry*).
+* Added footprint correction, super-mirror calibration, and normalisation between sample and supermirror measurements on a per-pixel & per-Q-bin level in the Amor workflow `#97 <https://github.com/scipp/ess/pull/97>`_ (*Reflectometry*).
 
 Contributors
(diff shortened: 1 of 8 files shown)


src/ess/amor/load.py
```diff
@@ -46,10 +45,7 @@ def _tof_correction(data: sc.DataArray, dim: str = 'tof') -> sc.DataArray:
 
 
 def load(filename,
-         owner: fileio.base.Person,
-         sample: fileio.data_source.Sample,
-         creator: fileio.base.Person,
-         reduction_script: str,
+         orso: Any = None,
```
jl-wynen:
```suggestion
         orso: Optional[Any] = None,
```

It could also be
```.py
try:
  from orsopy import fileio
  Orso = fileio.orso.Orso
except ImportError:
  Orso = None

# ...
         orso: Optional[Orso] = None,
```

Not sure how far you want to go with supporting type hints...

docs/about/release-notes.rst
```diff
@@ -3,6 +3,22 @@
 Release Notes
 =============
 
+v0.5.0 (April 2022)
+-------------------
+
+Features
+~~~~~~~~
+
+* Added resolution function for Amor and support for Orso file format in the ``reflectometry`` module `#115 <https://github.com/scipp/ess/pull/115>`_.
```
arm61: Should this not also mention the normalisation changes from https://github.com/scipp/ess/pull/97?

src/ess/amor/orso.py
```diff
@@ -0,0 +1,53 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2022 Scipp contributors (https://github.com/scipp)
+from .. import __version__
+
+from typing import Any
+import platform
+from datetime import datetime
+
+
+def make_orso(owner: Any, sample: Any, creator: Any, reduction_script: str) -> Any:
```
arm61: I don't think these can be `Any` I think they need to be the specific Orso classes.

src/ess/amor/orso.py
```diff
@@ -0,0 +1,53 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2022 Scipp contributors (https://github.com/scipp)
+from .. import __version__
+
+from typing import Any
+import platform
+from datetime import datetime
+
+
```
jl-wynen, line 9: This file depends directly on `orsopy`. So I would say that it should import it at module scope. Users have to make sure not to import `amor.orso` if they don't have it.

This would also allow for concrete type hints.

docs/instruments/amor/amor_reduction.ipynb
```diff
@@ -85,6 +85,34 @@
...
    ]
   },
+  {
+   "cell_type": "markdown",
+   "id": "fdee84e5-f3c9-41f9-8bbe-d7e271c59af5",
+   "metadata": {},
+   "source": [
+    "## Setting the experiment metadata\n",
+    "\n",
+    "We use the [Orso](https://www.reflectometry.org/) reflectometry standard and it's Python interface\n",
```
jl-wynen:
```suggestion
    "We use the [Orso](https://www.reflectometry.org/) reflectometry standard and its Python interface\n",
```

docs/about/release-notes.rst
nvaytet: Yes, can you write a sentence that I can add?

src/ess/amor/orso.py
nvaytet: I'm using `Any` because I don't want to import `orsopy` just for the type hints, as this would make `orsopy` a hard dependency.

src/ess/amor/orso.py
nvaytet: You mean you would then do
```Py
import ess.amor.orso as orso
orso_obj = orso.make_orso()
```
?

docs/about/release-notes.rst
arm61:
```suggestion
* Added resolution function for Amor and support for Orso file format in the ``reflectometry`` module `#115 <https://github.com/scipp/ess/pull/115>`_.
* Enabled normalisation between sample and supermirror measurements on a per-pixel & per-Q-bin level, footprint correction and supermirror calibration `#97 <https://github.com/scipp/ess/pull/97>`_
```

src/ess/amor/orso.py
arm61: Fair enough.

arm61: What about using a Forward Reference? https://stackoverflow.com/questions/39740632/python-type-hinting-without-cyclic-imports

nvaytet: I think it would still make orsopy a hard-dependency? The issue wasn't a cyclic import but that the package may not be installed at all? In any case, as you saw @jl-wynen suggested we do not import this by default, but the user has to manually import. This is ok for now I think

arm61: Cool 😎

src/ess/amor/orso.py
jl-wynen: Yes

nvaytet:
@arm61 @jl-wynen if you think this is now ok, I can make a release

arm61 (approved):

PR #119, opened 2022-04-12
Build docs in separate step

nvaytet:
Having docs building as part of the conda-build tests is not good when `pip` packages are required because those packages end up in the conda package.

We build the docs in a separate step, as is done in scipp, to avoid this.

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -3,19 +3,22 @@ name: ess-docs-build
 channels:
   - conda-forge
   - mantid
+  - scipp
   - nodefaults
 
 dependencies:
   - ipympl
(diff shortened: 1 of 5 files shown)


docs/environment.yml
```diff
@@ -0,0 +1,25 @@
+name: ess-docs-build
+
+channels:
+  - conda-forge
+  - mantid
+  - scipp
+  - nodefaults
+
+dependencies:
+  - ipympl<0.9
```
SimonHeybrock: Add a comment why this is limited?

SimonHeybrock (approved):

PR #128, opened 2022-05-31
Address some TODO items, use scippneutron 0.7, pin packages in CI, skip conda-build for PRs

nvaytet:
**TODOs**: they were left over from when we were using older versions of scipp with missing features
- Some simple deletions
- Removed the old and very outdated peak-finding method for finding frame edges in WFM
- Use `sc.midpoints` and remove unused `wfm.tools` module.

**New Scipp/Scippneutron:**
- Update to scippneutron 0.7 and scipp 0.14

**CI dependency pinning:**
- Due to more build failures, we are now pinning all packages for the env that runs the tests and builds the docs.
- The env pinning was created using
```bash
conda create -f environment.yml
conda activate ess-developer
conda env export --no-builds --file .buildconfig/ci.yml
````
- The `environment.yml` should now be used for developers, it replaces the `developer-extra.yml` and `metatoenv.py` files.

**Skip conda build for PRs:**
- Running a `conda-build` step in PRs increases build times significantly
- We remove that step (keep it only for releases) and manually run `pip install -e .` instead before running unit tests and building docs

### CODE REVIEW

File: src/ess/sans/i_of_q.py [hunk 1]
```diff
@@ -81,10 +81,7 @@ def denoise_and_rebin_monitors(monitors: Union[dict, sc.DataArray],
             dim = non_background_range.dim
             below = monitors[dim, :non_background_range[0]]
             above = monitors[dim, non_background_range[1]:]
-            # TODO: if we implement `ones_like` for data arrays, we could use that here
-            # instead of dividing the below and above pieces by themselves
-            divisor = sc.nansum(below / below).data + sc.nansum(above / above).data
-            background = (below.sum().data + above.sum().data) / divisor
+            background = sc.concat([below.data, above.data], dim=dim).mean()
```
jl-wynen, line 84: I don't think this is exactly the same. The old code counted only non-zero elements in `divisor`. `mean` counts all elements.

src/ess/wfm/plot.py
```diff
@@ -190,10 +176,6 @@ def frames_before_stitching(data: sc.DataArray,
                           data=summed,
                           dim=dim,
                           bins=bins_per_frame)
-        # TODO: is this still required with the new plotting in scipp 0.8?
-        if out[key].bins is not None:
-            out[key] = _bin_event_data_for_ploting(out[key], frames['frame', i],
-                                                   bins_per_frame)
```
jl-wynen, line 196: Does the code now rely on automatic histogramming in the plot?

src/ess/sans/i_of_q.py
nvaytet: I think the old code was wrong ;-)

src/ess/wfm/plot.py
nvaytet: Yes. This will have to change again with the new plotting.

src/ess/sans/i_of_q.py
jl-wynen: OK

jl-wynen (approved):

jl-wynen (approved):

nvaytet:
@jl-wynen I updated the CI (pinned package versions and removed the conda-build step from the PR builds). Can you have another look? Thanks!

jl-wynen (approved):

PR #129, opened 2022-06-21
First submission for NUrF at LoKI

gudlot:
This is my first submission for the NUrF module. Work in progress. `nurf.py` contains all functions and `scipp_947.ipynb` contains a Jupyter notebook. 
This is a PR to open the discussion on code for the NUrF module. `New_fluo_plot_function` is my current feature development branch.


docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1922 @@
...
+from scipy.signal import medfilt
+from scipy.optimize import leastsq  # needed for fitting of turbidity
+
+# local application imports
+import scippneutron as scn
+import scippnexus as snx
+import scipp as sc
+print("This is scippneutron ", scn.__version__)
+print("This is scippnexus ", snx.__version__)
+print("This is scipp ", sc.__version__)
```
SimonHeybrock: We should not have `print` statements at module-level. If output is needed we should use the logging facilities. I would not advocate logging the versions though.

docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1922 @@
...
+
+
+def complete_fname(scan_numbers):
+    """Converts a list of input numbers to a filename uses at ILL.
```
SimonHeybrock: Is this relevant for ESS? If it is just for testing now while you have sample files from ILL I think we should consider moving it elsewhere?

docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1922 @@
...
+
+    Parameters
+    ----------
+    scan_numbers: list of int or a single int
```
SimonHeybrock: We are using a Sphinx extension so the type information here is handled automatically based on type-hints. That is, here we should just have:
```suggestion
    scan_numbers:
```
and the corresponding type hints above needs to be written.

For reference, see https://github.com/scipp/ess/blob/d08178fa5442b3701a806d293585211162c5e838/docs/conf.py#L45 for the extension.

docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1922 @@
...
+        List of filenumbers or one filenumnber.
+
+    Returns:
+    ----------
+    flist_num: list of str or  one  str
```
SimonHeybrock: See above, do not type info include in the docstring, use type hints and our extension does the rest.
```suggestion
    flist_num:
```
See also https://scipp.github.io/reference/developer/coding-conventions.html#docstrings.

docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1922 @@
...
+    """
+
+    assert isinstance(da, sc.DataArray)
+
+    # We have already a condition variable 'is_dark' and we separate the incoming
+    #  sc.DataArray accordingly
+    dark = da[da.coords["is_dark"]].squeeze()
+    ref = da[da.coords["is_reference"]].squeeze()
+    #TODO: wpotrzebowski points out that dark, reference are also data. Should we convert the entry into the proposed LoKI.nxs to 'is_sample'?
+    data = da[da.coords["is_data"]]
```
SimonHeybrock: `is_sample` sounds ok to me.

docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1922 @@
...
+        flist_num = f"{str(scan_numbers).zfill(6)}.nxs"
+
+    if isinstance(scan_numbers, list):
+        # convert a list of input numbers to real filename
+        flist_num = [str(i).zfill(6) + ".nxs" for i in scan_numbers]
+
+    return flist_num
+
+
+def single_contributions(da):
```
SimonHeybrock: This function name is not very clear.

docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1922 @@
...
+    data = da[da.coords["is_data"]]    
+    # Why do we have to use .sequeeze() here?
+    # GL: Dark and ref have only one spectrum, but data could have multiple in it. So
+    # we need to squeeze dark, ref, but not data.
+    # SH: It allows for the subsequent operation to broadcast this input, i.e., use the
+    #  same dark spectrum for all signal spectra.
+    # SH: You cannot combine things in operations that have different length along a
+    # dimensions
+
+    # gather single DataArrays in a dictionary
```
SimonHeybrock: Remove comments?

docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1922 @@
...
+    da_dict = {"data": data, "reference": ref, "dark": dark}
+
+    return da_dict
```
SimonHeybrock:
```suggestion
    return {"data": data, "reference": ref, "dark": dark}
```
?

But why not return a `sc.Dataset`?

docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1922 @@
...
+
+    # separation
+    uv_dict = single_contributions(uv)
+
+    return uv_dict
+
+
+def normalize_uv(
+    *, data: sc.DataArray, reference: sc.DataArray, dark: sc.DataArray
+) -> sc.DataArray:  # <- here the star means that the function accepts only keyword arguments. It eliminates the risk of messing up the order of arguments
```
SimonHeybrock: Remove the comment.

docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1922 @@
...
+       UV spectrum.
+
+    Parameters
+    ----------
+    name: str
+        Filename for a LoKI.nxs file containting UV entry.
+
+     Returns
+    ----------
+    normalized.mean('spectrum'): sc.DataArray
```
SimonHeybrock: The part before the `:` is the "name" of the output variable. I don't think it should contain code?

docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1922 @@
...
+        One averaged UV spectrum. Averaged over all UV spectra contained in the file
+        under UV entry data.
+
+    """
+
+    uv_dict = load_uv(name)
+    normalized = normalize_uv(**uv_dict)  # results in DataArrays with multiple spectra
```
SimonHeybrock: Are these comments necessary?

docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1922 @@
...
+    legend_props = {"show": True, "loc": 1}
+    figs, axs = plt.subplots(1, 4, figsize=(16, 3))
+
+    # How to plot raw data spectra in each file?
+    out1 = sc.plot(
+        sc.collapse(uv_dict["data"], keep="wavelength"),
+        linestyle="dashed",
+        marker=".",
+        grid=True,
+        legend=legend_props,
```
SimonHeybrock, line 177: You could put all the common options (of the 4 plots) into a dict, and then use `**options` here, do avoid the duplication and make changes easier.

docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1922 @@
...
+    uv_dict: dict
+        Dictionary that contains UV data signal (data) from the sample, the reference,
+        and the dark measurement.
+        Keys: data, reference, dark
+
+    """
+
+    # load the nexus and extract the uv entry
+    with snx.File(name) as f:
+        uv = f["entry/instrument/uv"][()]
```
SimonHeybrock, line 69: I think we may have to avoid hard-coded names as much as possible. `scippnexus` may currently lack good options to find the group you are looking for. It could be done by hand (finding `NXentry` and `NXinstrument`), or using the undocumented `by_nx_class()` method (of the file root), but maybe addressing this comment has to wait until we have better tools available.

docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1922 @@
...
+    )
+    if vmin is not None and vmax is not None:
+        fig_all_spectra.ax.set_ylim(vmin, vmax)
+    if lambda_min is not None and lambda_max is not None:
+        fig_all_spectra.ax.set_xlim(lambda_min, lambda_max)
+    display(fig_all_spectra)
+    return fig_all_spectra
+
+
+def load_fluo(name):
```
SimonHeybrock, line 279: This function is duplicate with `load_uv`, apart from replacing `uv` with `fluorescence`, I suggest removing it, and pass the name of the `NXdata` to load as a function parameter.

docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1922 @@
...
+
+    Returns
+    ----------
+    final_fluo: sc.DataArray
+        DataArray that contains the calculated fluo signal, one spectrum or mulitple spectra.
+
+    """
+
+    # all spectra in this file are converted to final_fluo
+    final_fluo = (data - dark) - (reference - dark)
```
SimonHeybrock: This looks wrong. Should it be a division? Also, there is no `log` (as in the UV function). Is that intentional?

docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1922 @@
...
+    uv_dict = load_uv(name)  # dict of scipp.DataArrays
+    normalized = normalize_uv(
+        **uv_dict
+    )  # results in one DataArray, all uv spectra normalized
+
+    # In case you want to visualize the normalized sc.DataArray
+    # display(normalized)
+
+    normalized_avg = normalized.mean("spectrum")
+
```
SimonHeybrock: Wasn't there a function above that does examply this?

docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1922 @@
...
+            normalized.coords["wavelength"].values.transpose(),
+            normalized_avg.data.values.transpose(),
+            normalized.data["spectrum", :].values.transpose(),
+        )
+    )
+    path_to_save = os.path.join(path_output, output_filename)
+
+    # dump the content
+    with open(path_to_save, "w") as f:
+        np.savetxt(f, data_to_save, fmt="%.10f", delimiter="\t", header=final_header)
```
SimonHeybrock: Maybe we should use `pandas` for saving? We can easily convert a 1-D DataArray or Dataset to a `pandas.DataFrame`.

docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1922 @@
...
+        ]
+    )
+    hdrtxt = "wavelength [nm]\t uv_spectra_avg\t"
+    final_header = hdrtxt + l
+
+    data_to_save = np.column_stack(
+        (
+            normalized.coords["wavelength"].values.transpose(),
+            normalized_avg.data.values.transpose(),
+            normalized.data["spectrum", :].values.transpose(),
```
SimonHeybrock:
```suggestion
            normalized.values.transpose(),
```

docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1922 @@
...
+    )
+    path_to_save = os.path.join(path_output, output_filename)
+
+    # dump the content
+    with open(path_to_save, "w") as f:
+        np.savetxt(f, data_to_save, fmt="%.10f", delimiter="\t", header=final_header)
+
+
+def export_fluo(name, path_output):
+    """Export corrected fluo data contained in a LoKI.nxs file to .dat file
```
SimonHeybrock, line 583: What is the application of these "export" functions? Can we use NeXus instead? Here all the metadata (such as units) is lost, so this is not really an ideal solution.

docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1922 @@
...
+
+    from collections import defaultdict
+
+    # I need the help of the defaultdic function.
+    fluo_int_dict = defaultdict(dict)
+
+    for name in flist_num:
+        fluo_dict = load_fluo(name)
+        fluo_da = normalize_fluo(**fluo_dict)
+        print(f'Number of fluo spectra in {name}: {fluo_da.sizes["spectrum"]}')
```
SimonHeybrock: We should not `print` in production code. If necessary use logging.

docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1922 @@
...
+    ----------
+    fluo_int_dict: dict
+        A dictionary of nested dictionaries. For each found monowavelength, there are nested dictionaries
+        for each file containing the maximum intensity "intensity_max" and the corresponding wavelength "wavelength_max"
+
+    """
+
+    from collections import defaultdict
+
+    # I need the help of the defaultdic function.
```
SimonHeybrock: That is not a useful comment?

docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1922 @@
...
+            fmt="".join(["%.5f\t"] + ["%.5e\t"] * final_fluo.sizes["spectrum"]),
+            delimiter="\t",
+            header=final_header,
+        )
+
+
+def fluo_maxint_max_wavelen(
+    flist_num,
+    wllim=300,
+    wulim=400,
```
SimonHeybrock, line 662: For safety and convenience we should always provide such parameters as scalar (0-D) variables with the correct unit. This will avoid a lot of bugs in the future, and also makes the code more readable.

docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1922 @@
...
+        print(f"This is the fluo filt max intensity and wavelength dataset for {name}.")
+        display(fluo_filt_max)
+
+        # print(f'{name} Unique mwl:', np.unique(fluo_filt_max.coords['monowavelengths'].values))
+        unique_monowavelen = np.unique(fluo_filt_max.coords["monowavelengths"].values)
+
+        # create entries into dict
+        for mwl in unique_monowavelen:
+            fluo_int_max = fluo_filt_max["intensity_max"][
+                fluo_filt_max.coords["monowavelengths"] == mwl * sc.Unit("nm")
```
SimonHeybrock: What if `fluo_filt_max.coords["monowavelengths"]` is not in `nm`?

docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1922 @@
...
+            ].values  # returns a numpy array
+
+            # I collect the values in a dict with nested dicts, separated by wavelength
+            fluo_int_dict[f"{mwl}{sc.Unit('nm')}"][f"{name}"] = {
+                "intensity_max": fluo_int_max,
+                "wavelength_max": fluo_wavelen_max,
+            }
+
+    return fluo_int_dict
```
SimonHeybrock: I found the above very confusing and have trouble understanding what this is doing. Maybe a different data layout would be clearer?

docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1922 @@
...
+        ax[j, 1].set_xticks(
+            np.arange(0, len(max_int_wavelen_dict.keys()), 1),
+            labels=[f"{name}" for name in max_int_wavelen_dict],
+            rotation=90,
+        )
+
+
+def apply_medfilter(
+    da: sc.DataArray, kernel_size: Optional[int] = None
+) -> sc.DataArray:  # https://docs.python.org/3/library/typing.html#typing.Optional
```
SimonHeybrock:
```suggestion
) -> sc.DataArray:
```

docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1922 @@
...
+    out2 = sc.plot(
+        sc.collapse(da_mfilt, keep="wavelength"),
+        linestyle="dashed",
+        marker=".",
+        grid=True,
+        legend=legend_props,
+        title=f"After median filter - kernel_size: {kernel_size}",
+        ax=axs[1],
+    )
+    # display(figs)
```
SimonHeybrock, line 890: I suppose this is leftover from debugging? -> remove

docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1922 @@
...
+            labels=[f"{name}" for name in max_int_wavelen_dict],
+            rotation=90,
+        )
+
+
+def apply_medfilter(
+    da: sc.DataArray, kernel_size: Optional[int] = None
+) -> sc.DataArray:  # https://docs.python.org/3/library/typing.html#typing.Optional
+    """Applies a median filter to a sc.DataArray that contains fluo or uv spectra to remove spikes
+    Filter used: from scipy.ndimage import median_filter. This filter function is maybe subject to change for another scipy function.
```
SimonHeybrock, line 822: Can we now just remove the entire function and switch to using `scipp.ndimage.median_filter`?

docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1922 @@
...
+        legend=legend_props,
+        title=f"After median filter - kernel_size: {kernel_size}",
+        ax=axs[1],
+    )
+    # display(figs)
+
+    return da_mfilt
+
+
+def uv_peak_int(uv_da: sc.DataArray, wavelength=None, wl_unit=None, tol=None):
```
SimonHeybrock, line 895: See above, use a 0-D variable as a parameter, do not pass unit separately.

docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1922 @@
...
+    x2 = sc.linspace(
+        dim="wavelength", start=wavelength, stop=wavelength, num=1, unit=wl_unit
+    )
+    # alternative
+    # x2 = sc.array(dims=['wavelength'], values=[wavelength], unit=wl_unit)
+    # this gives us the intensity value for one specific wavelength
+    uv_int_one_wl = uv_interp(x2)
+
+    # now we want to have as well the wavelength for the interval
+    uv_int_mean_interval = uv_da_filt.mean(dim="wavelength")
```
SimonHeybrock, line 960: We should not have such experimental code (including commented code) here.

docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1922 @@
...
+        uv_peak_int_dict["wl_interval"]["spectrum", :],
+        linestyle="none",
+        grid=True,
+        title=f"UV peak intensity for intervall [{uv_peak_int_dict['wavelength']-uv_peak_int_dict['tol']} , {uv_peak_int_dict['wavelength']+uv_peak_int_dict['tol']}]{uv_peak_int_dict['unit']} in {name} ",
+        ax=ax1
+    )
+    ax1.set_ylabel("UV int")
+
+
+def uv_quick_data_check(
```
SimonHeybrock, line 1054: We have to figure out which functions the `nurf` module should provide to the user. It seems like we have quite a few things here that are more for debugging problems? If we have a huge list of public functions the user may have trouble finding/understanding what they need.

docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1922 @@
...
+
+    Parameters
+    ----------
+    b: np.ndarray
+        Offset, baseline
+
+    m: np.ndarray
+        Slope
+
+    wl: np.ndarray
```
SimonHeybrock, line 1253: Why using numpy directly here?

docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1922 @@
...
+    wl_unit: sc.Unit
+        Wavelength unit
+    fit_llim: int
+        Lower wavelength limit of fit range for turbidity
+    fit_ulim: int
+        Upper wavelength limit of fit range for turbidity
+    b_llim: int
+        Lower wavelength limit of fit range for b
+    b_ulim: int
+        Upper wavelength limit of fit range for b
```
SimonHeybrock, line 1319: See above, use 0-D variables.

docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1922 @@
...
+        # start parameters
+        p0 = [b0[i], m0[i]]
+        popt, pcov = leastsq(
+            residual,
+            p0,
+            args=(
+                uv_da_filt.coords["wavelength"].values,
+                uv_da_filt["spectrum", i].values,
+            ),
+        )
```
SimonHeybrock, line 1385: Can we use `scipp.optimize.curve_fit` (which uses scipy) instead?

docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1922 @@
...
+        dims=["spectrum"], values=res_popt[:, 0]
+    )
+    uv_da_turbcorr.attrs["fit-pcov"] = sc.array(
+        dims=["spectrum"], values=res_pcov[:, 0]
+    )
+
+    # display(uv_da_turbcorr)
+
+    # Switch on plots for verification
+    if plot_corrections:
```
SimonHeybrock, line 1415: Having such options on a function is often not a good design choice. Is this necessary? Can it be moved to a separate function that can be called on the return value if this function?

docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1922 @@
...
+    legend_props = {"show": True, "loc": 1}
+    num_spectra = multi_uv_turb_corr_da.sizes["spectrum"]
+
+    out = sc.plot(
+        sc.collapse(multi_uv_turb_corr_da, keep="wavelength"),
+        grid=True,
+        ax=ax[0],
+        legend=legend_props,
+        title=f"All turbidity corrected UV spectra for {num_spectra} spectra",
+    )
```
SimonHeybrock, line 1558: As mentioned above, I think we should avoid mixing the "business logic" (such as fitting) from visualization/plotting. Keep in mind that we may want to run the processing code in script instead of a notebook, and we don't want to create plots then (or plot to files). Can you find a way to disentangle all the code that creates outputs (HTML or plots) from the rest? This applies to many functions here, not just this one.

docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1922 @@
...
+        wllim < wulim
+    ), "Lower wavelength limit needs to be smaller than upper wavelength limit"
+
+    # let's go and filter
+    # filter spectrum values for the specified interval, filtered along the wavelength dimension
+    fluo_filt = fluo_da["wavelength", wllim * wl_unit : wulim * wl_unit]
+
+    # there is no function in scipp similar to xr.argmax
+    # access to data np.ndarray in fluo_filt
+    fluo_filt_data = fluo_filt.data.values
```
SimonHeybrock:
```suggestion
    fluo_filt_data = fluo_filt.values
```
is equivalent

docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1922 @@
...
+
+    # set defaul values for wllim and wulim for wavelength interval where to search for the maximum
+    if wllim is None:
+        wllim = 300
+    if wulim is None:
+        wulim = 400
+
+    assert (
+        wllim < wulim
+    ), "Lower wavelength limit needs to be smaller than upper wavelength limit"
```
SimonHeybrock, line 1660: If this error is meant for the user, raise an exception instead of using `assert`.

SimonHeybrock (commented):
Thanks for the contribution, you have obviously put a lot of effort into this 👍 

See comments and questions below. If have at this point not looked into all the details of the more specific functions since I anticipate that they may change significantly if/once some of my more general comments are addressed.

SimonHeybrock:
General suggestion: Have you considering submitting the main "logic part" of this as a separate PR? That is, exclude all the plotting and debugging stuff, just keep the actual code for loading, normalization, filtering, fitting, etc.. This would make it easier to bring it into a state everyone is happy with to include in the package. You can then, as a second step, work out which plotting utilities and helpers the user/scientist may need and do all that in follow-up work.

docs/instruments/loki/nurf/nurf.py
jl-wynen: The versions get logged when you configure the `ess` logger for a workflow. Have a look at cell 2 in https://scipp.github.io/ess/instruments/amor/amor_reduction.html

But this should be done in a workflow, not in a module like `nurf.py`.

gudlot: Ok. I remove them. It was more for myself. I am not familiar with logging facilities, but would like to see them.

docs/instruments/loki/nurf/nurf.py
gudlot: No. That is a function I wrote for my ILL test data. I could put all ILL related functions in a different .py file.

docs/instruments/loki/nurf/nurf.py
gudlot: Ok. All data types will go into the function call and an empty line (apart from the explanation) remains in the docstring.

docs/instruments/loki/nurf/nurf.py
gudlot: Ok. It will be `is_sample` in the next iteration.

docs/instruments/loki/nurf/nurf.py
gudlot: `split_contributions` maybe? I am not creative here. Needs more thinking.

docs/instruments/loki/nurf/nurf.py
gudlot: Done.

docs/instruments/loki/nurf/nurf.py
gudlot: Why is line 85 better than 87, please? How would you write this correctly then in a docstring as the return variable does not have a name? 
In our first iteration, you suggested a `dict`. What is the advantage of a dataset over a dict here? 
A dataset would be more in line with scipp philosophy?

SimonHeybrock: It avoids introducing a named variable (which hurts readability). Not sure what it has to do with the docstring?

A dataset should be used where possible. It indicates that the contents are all related (sharing dimensions of same length, etc.), instead of a random assortment of stuff (a `dict` could contain anything).

docs/instruments/loki/nurf/nurf.py
SimonHeybrock: `split_sample_dark_reference`? "contributions" could be anything.

docs/instruments/loki/nurf/nurf.py
gudlot: My docstring for this function looks like that:
```
Returns:
    ----------
    da_dict: dict
            Dictionary that contains spectroscopy data signal (data) from the sample,
            the reference, and the dark measurement.
            Keys: sample, reference, dark

```

Is it okay, if it will look like that:
```
Returns:
    ----------
    :
            Dictionary that contains spectroscopy data signal (data) from the sample,
            the reference, and the dark measurement.
            Keys: sample, reference, dark

```
No variable name and no type hin according to your docstring convention.

docs/instruments/loki/nurf/nurf.py
gudlot: Done. I assume such a long func name does not violate yet anything.

docs/instruments/loki/nurf/nurf.py
gudlot: Done.

docs/instruments/loki/nurf/nurf.py
gudlot: No, it should not. The name is `normalized`. Adjusted the docstring.

docs/instruments/loki/nurf/nurf.py
gudlot: No. I have the tendency to write comments mainly for myself. They were meant for me. I will remove them. This is gone.

docs/instruments/loki/nurf/nurf.py
gudlot: Thank you. Or maybe looping through all keys and search for `uv`, `fluo`, and later for `perturbation`. I just want to mention the latter, so it is not forgotten.

docs/instruments/loki/nurf/nurf.py
gudlot: I am unsure how to do it. But also isn't there more of a difference in comparison to uv? Example
```
fluo = f["entry/instrument/fluorescence"][()]
```
Both load functions contain the path to the Nexus entry.

docs/instruments/loki/nurf/nurf.py
gudlot: No, I confirmed it with my collaboration partner Cedric. It is correct, even with the dark. 
A comment from my discussion with him about this topic
>I agree to keep it consistent as it is written. 
The dark measurement is necessary for this type of detector. Sometimes, one can use the fluorescence emission without the reference. In that case having the dark is important.
In most cases the reference should have no fluorescence emission, basically flat. In more complex solvent the reference may have some intrinsic fluorescence that would need to be substracted.
For consistency, we should keep the dark and reference for Fluo and Abs

I would like to keep this code line.

docs/instruments/loki/nurf/nurf.py
jl-wynen: Remove the indentation of the `----` line and you got it.
See here for details: https://scipp.github.io/reference/developer/coding-conventions.html#docstrings
We won't enforce this convention super strictly in `ess` but it would be great if you stuck to it. Especially to make sure that the HTML pages are rendered properly. (It is quite picky about spaces...)

docs/instruments/loki/nurf/nurf.py
gudlot: Yes and no. This export function aims for two things: 1. Export all UV spectra in a Loki.nxs as a table and in the average over all spectra in an additional column. This is why I have `normalized` and `normalized_avg` here. The function `process_uv` returns only the averaged spectra and the function `load_and_normalize_uv` returns all UV spectra, but no average. 
If I do it in the current way, I avoid a file being loaded twice.

docs/instruments/loki/nurf/nurf.py
gudlot: That could be probably as well an option. The user wished here for a simple ASCII file. That is the only reason why I wrote this function in first place. There are users out there, that have no Python knowledge and prefer ASCII files or wish to extract the data from the nxs files and use other software to look at the data.
I wrote last year Python code with `pandas`, where whole Loki files were manipulated and the results were stored in multicolumn `pandas.DataFrame`.

docs/instruments/loki/nurf/nurf.py
gudlot: Done

docs/instruments/loki/nurf/nurf.py
gudlot: As for the `export_uv` function, the client expressed the wish to have all data extracted from the NeXus files as ASCII files. There are still users out that prefer ASCII and import the data into Origin, for example, or as ASCII into MATLAB (I guess). I don't know if this is a supported function or should be left to the user in the future. Also not all users seem to be ready or familiar with Python or a Jupyter.

docs/instruments/loki/nurf/nurf.py
gudlot: I can remove it. The reason why I kept is that my data is not ideal. The number of spectra in my Nexus file varies. My current data is not super "perfect". I used it as a visual debugging option in Jupyter. 
I wanted to inspect multiple Dataarrays here and if I do `display(fluo_da)` there is currently no information on the dataarray showing from which file it originates. I could attach maybe an attribute to the dataarray?

docs/instruments/loki/nurf/nurf.py
gudlot: :-) True, but I really needed here the help of this module. I remove the comment. Done.

docs/instruments/loki/nurf/nurf.py
gudlot: Could you please show me an example with such a parameter as scalar variable with the correct unit? 
Do you mean something like `sc.Variable(dim='wavelength', value=300, unit=sc.Unit('nm'))` ?
In that case, I could unify value and unit in my code and remove some checks.

docs/instruments/loki/nurf/nurf.py
gudlot: Thank you. You are right. Well spotted. I made a mistake. The function has a key argument `wl_unit`. Although  the function `fluo_peak_int` checks for the unit
```
if wl_unit is None:
        wl_unit = fluo_da.coords["wavelength"].unit
 ```
 this check has to be executed as well in the function `fluo_maxint_max_wavelen`. My suggestion is for the moment  to add the code above in the function, until I have figured out how to combine `wl_unit` and `wulim` for example (see your comment above) or until I have found a better strategy.

File: docs/instruments/loki/nurf/nurf.py (hunk 24 above)
Comment (gudlot) [reply docs/instruments/loki/nurf/nurf.py]: This is maybe easier to solve in a meeting. But I try anyway to explain here. I needed to arrange the fluo content of several .nxs files into a new structure because this function aims to give an overview of what happens in a series of measurements.
The challenge is here that my test data is messy. The number of fluo spectra varies, but also the incident monowavelength order varies. It is a bit unpredictable. 

In each fluo part in a .nxs I can have multiple different incident monowavelengths, for example 5x 295nm and 6x 320nm. In the next file I have 4x 295nm and 4x 320nm. For each incident monowavelenght a fluo spectrum was recorded. 
Each nxs file is a dataarray. Now the idea is that I extract from all dataarrays all spectra that were measured with an incident monowavelength of 295nm and I am interested here in the max peak position and the corresponding wavelength of this maximum. I repeat this for 320nm.
My code is flexible enough to find any monowavelength and bundle the spectra. 
This is why I aimed for nested dictionaries.

File: docs/instruments/loki/nurf/nurf.py (hunk 25 above)
Comment (gudlot) [reply docs/instruments/loki/nurf/nurf.py]: Done
`apply_medfilter` needs to be rewritten to use the scipp median filters.

File: docs/instruments/loki/nurf/nurf.py (hunk 33 above)
Comment (gudlot) [reply docs/instruments/loki/nurf/nurf.py:1385]: I am not sure. I had a look at the `scipy.optimize.curve_fit` function while writing my function. But I need here to minimise the difference between the measured spectral data and the turbidity that is fitted.  I was not able to realize this with the `curve_fit` function, hence I chose `scipy.optimize.leastsq`.

File: docs/instruments/loki/nurf/nurf.py (hunk 34 above)
Comment (gudlot) [reply docs/instruments/loki/nurf/nurf.py:1415]: I chose it to declutter the output in the notebook. Sometimes it is useful to see the in-between plotting steps, sometimes not.  How would you do this, Simon, please? Would you have a dummy code for a design pattern by chance?
And yes, I agree with your other comment - to separate processing tools from plotting workflows.

File: docs/instruments/loki/nurf/nurf.py (hunk 35 above)
Comment (gudlot) [reply docs/instruments/loki/nurf/nurf.py:1558]: Yes, I agree to remove the business (fitting, reducing, normalising, regrouping) from the plotting business. I might have to repat some code.
Yes, I know it applies to many functions I wrote. The plots were helpful for visual inspection as well.

File: docs/instruments/loki/nurf/nurf.py (hunk 36 above)
Comment (gudlot) [reply docs/instruments/loki/nurf/nurf.py]: Done. It occurs in other locations as well, need to check if nothing breaks when I apply there this suggestion.

File: docs/instruments/loki/nurf/nurf.py (hunk 37 above)
Comment (gudlot) [reply docs/instruments/loki/nurf/nurf.py:1660]: I am not sure, it can be if I want. At least, it makes the user aware of his wrong choice. 
What is the rule, please? `assert` has the possibility to break the code and I was taught that if the code breaks in a controlled way, it is a good thing.
I don't know when to go for `raise ValueError` or `assert`.

**Comment by gudlot** [comment]
>General suggestion: Have you considering submitting the main "logic part" of this as a separate PR? That is, exclude all the plotting and debugging stuff, just keep the actual code for loading, normalization, filtering, fitting, etc.. This would make it easier to bring it into a state everyone is happy with to include in the package. You can then, as a second step, work out which plotting utilities and helpers the user/scientist may need and do all that in follow-up work.

Yes, it occurred to me while writing the functions. I am constantly torn between writing code "quick and dirty" (to quote somebody else) and writing better code following good programming standards, at least I am trying. The idea was first to show the functionality/workflow and get an idea on the data. Please don't forget my data comes from an experiment where not everything has worked technically in a good way. That affected the datasets (=multiple nxs files). Priority was on the beamtime, and resources to support the implementation of the spectrometer were limited. That will be in the future a task for ESS maybe to ensure that data from the spectrometers are correctly recorded and handles errors. You and the others commentators are the first people, I can discuss my code with in terms of code design/strategy. 
My idea would be to do a process function, return a dataarray or dataset (or a function layer in between that gathers the values from multiple files) and build around it the corresponding plot function. I would name all plot functions with `plot_` in the beginning.  
Thank you very much for the discussion, the helpful comments, and your patience. I try my best to improve and learn.

**Comment by wpotrzebowski** [comment]
> used at ILL

@gudlot Yes, I know it is ILL data but I think there was typo in the doc string ("at" missing), which I think has been corrected now.

**Comment by wpotrzebowski** [comment]
I think it is mostly the question for @nvaytet. Sorry if I missed this in conversation but is there an intention to keep all files (including python files) in docs/instruments/loki/nurf/? As far as I know NURF is avaialable both at ISIS and ILL and I think in the future they may also be interested in this scripts in data processing. It can always be addressed later but It may worth to think about it now.

File: docs/instruments/loki/nurf/nurf.py (hunk 22 above)
Comment (nvaytet) [reply docs/instruments/loki/nurf/nurf.py:662]: `sc.scalar(300, unit='nm')`

Comment (gudlot) [reply docs/instruments/loki/nurf/nurf.py:662]: Thank you. Is it worth having the `unit` as well as the handover parameter, `unit=wl_unit`? It will be probably by default `nm`.

File: docs/instruments/loki/nurf/ill_auxilliary_funcs.py [hunk 38]
```diff
@@ -0,0 +1,393 @@
...
+
+        # print(data_spectro_file)
+
+    # file_handle is returned as np.ndarray and its an np.array, elements correspond to row indices
+    return data
+
+
+def nurf_file_creator(loki_file, path_to_loki_file, data):
+    """
+    Appends NUrF group to LOKI NeXus file for ESS
```
nvaytet, line 129: Is this the way NUrF data will always be added to the LoKI files? Or is this a temporary way to do things until all the data is properly stored in the Nexus files?
If the latter, I am not sure this should become part of the `ess` package. To me it looks like a (still experimental?) utility script that patches the data?

docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1892 @@
...
+            header=final_header,
+        )
+
+
+def fluo_maxint_max_wavelen(
+    flist_num,
+    wllim=300,
+    wulim=400,
+    wl_unit=None,
+    medfilter=True,
```
nvaytet, line 664: You seem to have several functions that say they do one thing, "keep the data between min and max wavelength" in this case, but are actually doing more under the hood, since you can also apply a mean filter.
This is usually a sign that you should either make two functions and call both one after the other, or change the function name.
In this case, I feel that selecting a wavelength range is quite different from applying a filter. So I would vote for two functions.

Try to have a look at other places where you are doing something similar, and change accordingly.

docs/instruments/loki/nurf/nurf.py
```diff
@@ -0,0 +1,1892 @@
+# standard library imports
```
nvaytet, line 1: As Simon suggested, we need to split things up before we can move forward here.
Splitting up the processing from the plotting is a good start. I think I would like to break it up even further.

You seem to have two different workflows, UV and fluo. Are the two related (i.e. you cannot process one without the other?), or could you run them independently? Basically, could you submit changes for just one of these first (e.g. just the UV?), and once we nail the design, the general layout could be copied for processing the other one?

My suggestion:

1. Identify the minimal list of steps you need to do for a workflow (I think for now we should just start with processing and leave exporting data to ascii out of this PR? Maybe things like fitting can also be added later?)
2. Ideally, your notebook should just read like a list of function calls from the `ess.loki.nurf` module.
3. Move the `nurf.py` file out of the docs and into its own submodule in `ess.loki` (i.e. in a folder called `src/ess/loki/nurf`)
4. Strip out all the plotting code for now, and just have the list of processing steps in a notebook that you put in the docs
5. Add some final simple plotting in the notebook for now (we will add plotting of intermediate steps later)
6. Note that if a user wants a highly customized figure made for a paper, it will not be stored here. We will just have simple plots

Don't do all this in one go. Start with item 1. here, once we have a list, we can look at the other items.

docs/instruments/loki/nurf/ill_auxilliary_funcs.py
gudlot: When I started this project, we discussed how the Nurf data could be part of a Loki file. I appended my content to an example Loki file. At the moment it is more of a temporary way and my current nxs files contain only the uv and fluo relevant parts, nothing else. But for Kenan I append my data to the Loki file. 
No, it does not have to become part of any package. Please consider all functions in this file as helper functions that I needed to convert my data into a better format.

docs/instruments/loki/nurf/nurf.py
gudlot: Thank you. My thoughts were that using a medfilter is optional, depending on the data quality. Sometimes the data has a lot of outliners. With the medfilter I smooth it first and then I select the maximum intensity in a given wavelength range. If there would be an outliner, that could be misinterpreted.
I could separate them into functions, but probably would have a third to bring the two together.

docs/instruments/loki/nurf/nurf.py
gudlot: Thank you very much.

>You seem to have two different workflows, UV and fluo. Are the two related (i.e. you cannot process one without the other?), or could you run them independently? 

Yes, there are two workflows, one for the UV spectroscopy one for the fluo spectroscopy. I can process each without the other. Only later in the data interpretation, both will be together as they are complementary. 
>Basically, could you submit changes for just one of these first (e.g. just the UV?), 

Does it mean in two or even three separated python files? UV, fluo, medfilter (because this is a function that can be applied to both workflows?)
>and once we nail the design, 

Should this be graphically outlined? 
>the general layout could be copied for processing the other one?

No, the workflows for UV and fluo are really different and extract different parts of each spectrum.

>1. Identify the minimal list of steps you need to do for a workflow (I think for now we should just start with processing and leave exporting data to ascii out of this PR? Maybe things like fitting can also be added later?)

Ok to the minimal list of steps. 
Re: export to ascii: This is a wish of the user. The idea behind was if the user needs to further process the data with another program that cannot handle hdf5. The users wishes to have access to all raw data and processed data. 
Re fitting: Turbidity is a standard correction in UV spectroscopy I understood

>2. Ideally, your notebook should just read like a list of function calls from the ess.loki.nurf module.

What will be part then of the nurf module? Will it be composed of UV, fluo, plotting?
Will the plot functions not be part of `ess.loki.nurf`?  They are also part of the data interpretation/visualisation and an important part during the experiment.

>4. Strip out all the plotting code for now, and just have the list of processing steps in a notebook that you put in the docs

Okay, that can be done.

>5. Add some final simple plotting in the notebook for now (we will add plotting of intermediate steps later)

What would be considered a final simple plot? UV signal as a function of wavelength, maybe processed? Same for fluo spectra?

>6. Note that if a user wants a highly customized figure made for a paper, it will not be stored here. We will just have simple plots

The user does not request highly customized figures. The plots that I generated are standard plots for UV, fluo spectroscopy. They are just a bit more complicated than an I(q) plot for example.  The idea is to measure a perturbation, record SANS, fluo, UV along with it and show this during the experiment to the user.

>Don't do all this in one go. Start with item 1. here, once we have a list, we can look at the other items.

Okay. I could provide a workflow list.

nvaytet: A few comments.

For the file/module organisation, I think separating uv, fluo and plotting is good.
So something like:
```
src/ess/loki/
        |--> nurf/
              |--> __init__.py
              |--> fluo.py
              |--> uv.py
              |--> tools.py  # or utils.py, or common.py, however you want to name it
              |--> plot.py
```
?

We would start with either `fluo.py` or `uv.py`, probably with the `tools.py` as well, and then add the rest later.

> Should this be graphically outlined?

No need to draw things, just outlining a list of steps or function calls should be fine.

docs/instruments/loki/nurf/nurf.py
SimonHeybrock: Then why do you not replace this with `data - reference`? I doubt the difference in floating-point rounding-errors is relevant?

SimonHeybrock: My point is, this line of code anyway does not work without `reference` (so Cedric's comment does not seem to apply), so why is it written in this weird way?

docs/instruments/loki/nurf/nurf.py
SimonHeybrock: I mean, use pandas to save to ASCII.

docs/instruments/loki/nurf/nurf.py
SimonHeybrock: No, our point is that there should not be a default unit (unless the value is also default). Otherwise you risk bugs, when the user assumes the default is something else, or worse, when someone ever decides to change the default all user code is silently broken.

docs/instruments/loki/nurf/nurf.py
gudlot: I tried this, but then I got into trouble while loading the file. It suddenly asked for other parameters not applicable to my case. For the moment, I will keep `is_data`

docs/instruments/loki/nurf/nurf.py
gudlot: @jl-wynen  Thank you for the explanation. I checked, and for the moment there should not be any indentation for any dashed line present. I will stick to your docstring conventions.

gudlot: > It avoids introducing a named variable (which hurts readability). Not sure what it has to do with the docstring?

Probably a more simple question on my side: If there is no direct return variable, what would one write then into the docstring under Returns, please? 


> A dataset should be used where possible. It indicates that the contents are all related (sharing dimensions of same length, etc.), instead of a random assortment of stuff (a `dict` could contain anything).
I thought `dict` is the way to go with key:value pairs as a unique combination.
I tested to put `dark`, `reference`, `sample` into a dataset (dummy code that works with my current code):
```py
ds=sc.Dataset()
ds['sample']=uv_dict['sample']
ds['dark']=uv_dict['dark']
ds['reference']=uv_dict['reference']
```
Under `coordinates` I have the `integration_time` of the sample measurements. Integration time for `dark`, `reference` appear now under `attributes` in the corresponding entries under `Data`. Are they ranked now second-classed because they appear as attributes and not as coordinates?
We could opt for a dataset here, but I think this needs a lot of code changes down the way. We should be sure we are happy with it.
And what happens if I have multiple .nxs files in a series? Would be the idea to stack only normalized/corrected files together to a dataset? In that case, dark and reference would be dropped.

docs/instruments/loki/nurf/nurf.py
gudlot: True. And where would you store this dict? Still in the same functions or maybe put it into a different function and call it, similar to the marker function maybe? Thank you for your advice.

docs/instruments/loki/nurf/nurf.py
gudlot: Replaced by 
```
load_nurfloki_file(name: str, exp_meth: str )
```

docs/instruments/loki/nurf/nurf.py
gudlot: I think this is solved with `load_nurfloki_file(name: str, exp_meth: str )`

docs/instruments/loki/nurf/nurf.py
gudlot: I think Cedric gave the answer. In the case where the same dark is used, it simplifies to `sample-reference`, but where no reference is used, it would be `sample-dark`. Sure, the code in its current version does not foresee this.

docs/instruments/loki/nurf/nurf.py
gudlot: Don't you need a Pandas dataframe in first place to do so? In the meantime, I updated the export function to the user's wish. But these export function are not my highest priority at the moment because there is currently no other program that requires this output. It is a could have maybe.

docs/instruments/loki/nurf/nurf.py
gudlot: I am still unsure. How would you provide something like this then, e.g.:
```py
fluo_maxint_max_wavelen(
    flist_num,
    wllim=300,
    wulim=400,
    wl_unit=None,
    medfilter=True,
    kernel_size=15,
```
The function should look for the maximum intensity value in a range of 300-400nm. If the user does not want it, he can change it.
Or is it simply: `wllim=sc.scalar(300, unit='nm')` in the function at the top as @nvaytet has suggested, i.e. don't split value+unit.

SimonHeybrock: > How would you provide something like this then, e.g.:
> ...
> Or is it simply: wllim=sc.scalar(300, unit='nm') in the function at the top as @nvaytet has suggested, i.e. don't split value+unit.

Yes, I think so.

docs/instruments/loki/nurf/nurf.py
gudlot: I kept it to show the influence of the median filter for the user. I like to visualise changes.

docs/instruments/loki/nurf/nurf.py
gudlot: Yes, it is also on my list to rewrite this with more scipp functionality.

docs/instruments/loki/nurf/nurf.py
gudlot: Yes. Now it is clearer. Will adjust accordingly.

docs/instruments/loki/nurf/nurf.py
gudlot: There were two ways of how to get access to the data and I followed the clients wish. I like to comment my code, but can remove this.

docs/instruments/loki/nurf/nurf.py
gudlot: Sure. But the current data I work with is not clean. Clean data requires also better hardware control on the Nurf to feed clean data into the `nurf` module, some functions could and will probably disappear. Please keep in mind, if a user is at Loki, he wants to see quickly spectroscopy data. This particular function is left from another plotting function, and it shows me a graph of how many data points I have per file in a series. 
I suggest giving a user Jupyter notebooks to demonstrate the workflow.
I can get rid of this function.

docs/instruments/loki/nurf/nurf.py
gudlot: It worked. I had a deadline and so I went in the first iteration with numpy to finish on time. If it should be done in scipp, code needs to change because of some unit mismatch. I did not have experience writing fit algorithms with scipp. In addition, there is an assumption in spectroscopy to get rid of the unit for `b`, but the value was supposed to be one. Would need to check again with the expert user.

docs/instruments/loki/nurf/nurf.py
gudlot: Okay, I will bring units and values together.

docs/instruments/loki/nurf/nurf.py
gudlot: I wrote now for `utils.py` a little wrapper `nurf_median_filter(da,kernel_size)` to apply the `scipp.ndimage.median_filter` in direction of the the `wavelength`.

nvaytet:
@gudlot What is the status on this? Can I help you with something to get this moving again?

PR #132, opened 2022-07-04
Remove cap on scippneutron version pinning and prepare release 0.7.1

nvaytet:
For some background about why we are removing the upper limit on the version number, see https://iscinumpy.dev/post/bound-version-constraints/

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -15,7 +15,7 @@ requirements:
     - ipywidgets
     - matplotlib
     - pooch
-    - scippneutron=0.7.*
+    - scippneutron>=0.7
 
 test:
   imports:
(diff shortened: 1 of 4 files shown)


conda/meta.yaml
```diff
@@ -15,7 +15,7 @@ requirements:
     - ipywidgets
     - matplotlib
     - pooch
-    - scippneutron=0.7.*
+    - scippneutron>=0.7
```
SimonHeybrock, line 18: This unpins the runtime-requirement, but to be consistent what we have learned and implemented for the other subprojects we should pin our *test* requirement? Is this possible for the CI workflow?

nvaytet: This should be pinned in https://github.com/scipp/ess/blob/main/.buildconfig/ci.yml

SimonHeybrock (approved):

PR #134, opened 2022-07-14
Configure Mantid's logger

jl-wynen:
With Mantid 6.4, we can now capture Mantid's log messages and display them along our logs. This should fix the problem of messages not showing sometimes.

If an older version of Mantid is installed, the configuration will work as before and not touch Mantid.

### DIFF
--- a/docs/instruments/loki/sans2d_to_I_of_Q.ipynb
+++ b/docs/instruments/loki/sans2d_to_I_of_Q.ipynb
@@ -60,7 +60,7 @@
    },
    "outputs": [],
    "source": [
-    "_ = configure_workflow('sans2d_I_of_Q', filename='sans2d.log')"
+    "logger = configure_workflow('sans2d_I_of_Q', filename='sans2d.log')"
    ]
   },
   {
--- a/docs/instruments/loki/sans2d_to_I_of_Q.ipynb
+++ b/docs/instruments/loki/sans2d_to_I_of_Q.ipynb
@@ -32,12 +32,35 @@
    "cell_type": "code",
    "execution_count": null,
    "id": "8c7f7cf7-0582-4953-a772-a0f87d1cf0e2",
-   "metadata": {},
+   "metadata": {
+    "tags": []
+   },
    "outputs": [],
    "source": [
     "import scipp as sc\n",
     "from ess import loki, sans\n",
+    "from ess.logging import configure_workflow\n",
     "import scippneutron as scn"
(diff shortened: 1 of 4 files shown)


nvaytet:
Can you add a logging configuration step to the LOKI (sans2d) workflows so that it also catches the Mantid output there?
Screenshot at 2022-07-14 12-18-05

Also, I remember we have a similar problem in `scippneutron`, e.g. the instrument view page. Do we need a similar fix there?

jl-wynen:
> Can you add a logging configuration step to the LOKI (sans2d) workflows so that it also catches the Mantid output there?

Done.

I postponed doing anything in scippneutron until we are happy with the setup in ess. We didn't like having to call a config function manually. And I don't think anything has changed.

nvaytet:
Do i remember correctly that getting rid of the config step would probably be solved by a workflow builder/manager or something like it?
So we would need a different solution for scippneutron anyway?

docs/instruments/loki/sans2d_to_I_of_Q.ipynb
```diff
@@ -32,14 +32,37 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "760adfde-01e6-4d25-bc8d-b8bbd20c0467",
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "_ = configure_workflow('sans2d_I_of_Q', filename='sans2d.log')"
```
nvaytet:
```suggestion
    "logger = configure_workflow('sans2d_I_of_Q', filename='sans2d.log')"
```
Would be a bit less cryptic? (This is how it's done in the Amor notebook)

jl-wynen:
> Do i remember correctly that getting rid of the config step would probably be solved by a workflow builder/manager or something like it?
> So we would need a different solution for scippneutron anyway?

If we end up implementing a manager, yes.
Since scippneutron does not have any 'application' components, I think the best we can do is provide a similar configure function and call that at the top of a notebook. Unless, that is, we implement a workflow manager in scippneutron, there is nothing inherently ess-specific about it.

docs/instruments/loki/sans2d_to_I_of_Q.ipynb
jl-wynen: Ok. It is unused, though.

nvaytet (approved):

PR #136, opened 2022-08-02
Update to scippneutron=0.8

nvaytet:


docs/conf.py
```diff
@@ -92,6 +92,9 @@
 # The master toctree document.
 master_doc = 'index'
 
+# Turn warnings into errors
+warning_is_error = True
```
SimonHeybrock, line 96: So also this in the pytest config: https://github.com/scipp/scipp/blob/c776d04616a36480968e9f6e3905aa62d1b59719/pyproject.toml#L25-L28

tests/wfm/stitch_test.py
```diff
@@ -95,13 +95,8 @@ def _do_stitching_on_beamline(wavelengths, dim, event_mode=False):
...
     else:
-        out = sc.rebin(wav,
-                       dim='wavelength',
-                       bins=sc.linspace(dim='wavelength',
-                                        start=1.0,
-                                        stop=10.0,
-                                        num=1001,
-                                        unit='angstrom'))
+        out = wav.rebin(wavelength=sc.linspace(
+            dim='wavelength', start=1.0, stop=10.0, num=1001, unit='angstrom'))
```
SimonHeybrock, line 99:
```suggestion
        out = wav.rebin(wavelength=1000)
```
may be good enough for a test?

tests/wfm/stitch_test.py
```diff
@@ -121,13 +116,10 @@ def _check_lambda_inside_resolution(lam,
...
-                              edges=[
-                                  sc.array(dims=['wavelength'],
-                                           values=[(lam - dlam).value,
-                                                   (lam + dlam).value],
-                                           unit=lam.unit)
-                              ]).bins.sum().data['wavelength', 0]
+        sum_in_range = data.bin(
+            wavelength=sc.array(dims=['wavelength'],
+                                values=[(lam - dlam).value, (lam + dlam).value],
+                                unit=lam.unit)).hist().data['wavelength', 0]
```
SimonHeybrock, line 122: Something like
```suggestion
        sum_in_range = data.hist(wavelength=sc.concat([lam - dlam, lam + dlam], 'wavelength']))data['wavelength', 0]
```
unrelated, so ignore this is you like.

docs/conf.py
nvaytet: hmm we need to be using `tox` for this (I think?), which `ess` is currently still not using.

That said, would this raise errors only in the python tests?

I don't really understand why the warnings raised by the notebooks did not fail the docs build...

tests/wfm/stitch_test.py
nvaytet: Seems like it isn't. Test fails with your suggestion :-(

docs/conf.py
SimonHeybrock: Why tox? This is just a pytest option (in can also be configured in other config files).

docs/instruments/amor/amor_reduction.ipynb
```diff
@@ -762,11 +762,10 @@
    "outputs": [],
    "source": [
     "nbins = 165\n",
-    "theta_edges = sc.linspace(dim='theta', start=0.0, stop=1.2, num=nbins, unit='deg')\n",
+    "theta_edges = sc.linspace(dim='theta', start=0.0, stop=1.2, num=nbins, unit='deg').to_unit(theta_edges, 'rad')\n",
```
SimonHeybrock: Something looks wrong here, mixing up function and method syntax for `to_unit`?

docs/conf.py
nvaytet: ok, my mistake, i thought the pyproject file was a `tox` thing.

SimonHeybrock (approved):

tests/reflectometry/corrections_test.py
```diff
@@ -18,11 +18,7 @@ def test_normalize_by_counts():
                        values=values,
                        variances=values)
     array = sc.DataArray(data=data)
-    with warnings.catch_warnings(record=True) as w:
-        array_normalized = corrections.normalize_by_counts(array)
-        assert len(w) == 1
-        assert issubclass(w[-1].category, UserWarning)
-        assert 'orsopy' in str(w[-1].message)
+    array_normalized = corrections.normalize_by_counts(array)
```
SimonHeybrock, line 20: Why remove the asserts?

nvaytet: Because there was no more warning, it got converted to an error and the length of `w` was now 0.
I didn't find an option in pytest to make this still work. So i decided that the check here is not that important, and the test is more about checking that the normalization is correct.

SimonHeybrock (approved):

PR #138, opened 2022-08-29
Update to scipp0.16

nvaytet:
These changes are mainly due to coordinate transformation graphs having been moved to a new location in `scippneutron`.


.buildconfig/ci.yml
```diff
@@ -231,23 +229,21 @@ dependencies:
...
-  - pyzmq=23.2.0
+  - pyzmq=23.2.1
   - rapidjson=1.1.0
   - readline=8.1.2
   - requests=2.28.1
-  - scipp=0.15.1
-  - scippneutron=0.8.0
-  - scippnexus=0.1.7
+  - scipp=0.16.2
+  - scippneutron=0.9.1
```
jl-wynen: Let's wait for scippneutron 0.9.2 (building now)

src/ess/reflectometry/conversions.py
```diff
@@ -2,8 +2,8 @@
 # Copyright (c) 2022 Scipp contributors (https://github.com/scipp)
 import scipp as sc
 from scipp.constants import m_n, h, pi
-from scippneutron.tof import conversions
-from scippneutron.core.conversions import _elem_dtype, _elem_unit
+from scippneutron.conversion.graph import beamline, tof
+from scippneutron._utils import elem_dtype, elem_unit
```
jl-wynen, line 6: We need a proper way of exposing these! (This is a comment about scipp...)

src/ess/sans/conversions.py
```diff
@@ -3,8 +3,11 @@
 
 import scipp as sc
 from scipp.constants import m_n, h
-from scippneutron.tof import conversions
-from scippneutron.core.conversions import _elem_unit
+from scippneutron.conversion.graph import beamline, tof
+from scippneutron._utils import elem_unit
+
+# from scippneutron.tof import conversions
+# from scippneutron.core.conversions import _elem_unit
```
jl-wynen: Remove

src/ess/reflectometry/conversions.py
SimonHeybrock: I am wondering once more whether `dtype` and `unit` should just forward to the elem dtype and unit? There are several questions:

- What to do for `Dataset` buffer? There is not unique `dtype` and `unit` in this case. `unit` could be `None`.
- How to get the buffer type, if not via `dtype`?
- Is there code that will break, and why?

jl-wynen (approved):

PR #139, opened 2022-09-08
Add `nexus` submodule with components for building customized NeXus file loaders

SimonHeybrock:


src/ess/nexus/_loader.py
```diff
@@ -0,0 +1,159 @@
...
+    loaded = {}
+    for field in fields(schema):
+        key = field.name
+        cls = field.type
+        field_kwargs = arg_dict.get(key, {})
+        loaded[key] = cls.from_nexus(group, **field_kwargs)
+    return schema(**loaded)
+
+
+T1 = TypeVar('T1', bound='InstrumentMixin')
```
nvaytet: Maybe better names for the types? Something that references Instrument and Entry instead of 1 and 2 ?

src/ess/nexus/_loader.py
```diff
@@ -0,0 +1,159 @@
...
+
+def _load_multi(load: Callable, items: Dict[str, snx.NXobject], /,
+                **kwargs) -> Dict[str, sc.DataArray]:
+    return {
+        key: loaded
+        for key in items if (loaded := load(items[key], **kwargs)) is not None
+    }
+
+
+def _select_events_and_load(detector, pulse_min=None, pulse_max=None, **kwargs):
```
nvaytet, line 76: Why is there a special function for pulses but not any of the other things you might want to filter by? (e.g. detector id)

Also, I don't see any test for `pulse_min` and `pulse_max`.

src/ess/nexus/_loader.py
```diff
@@ -0,0 +1,159 @@
...
+        c = group[key] if isinstance(key, str) else group.__getattr__(key.__name__[2:])
+        data = load(c, **kwargs)
+        return cls(data) if isinstance(data, dict) else data
+
+    return type(name, (dict, ), dict(from_nexus=classmethod(from_nexus)))
+
+
+Fields = make_section("Fields", [snx.Field, snx.NXlog])
+Choppers = make_section("Choppers", [snx.NXdisk_chopper, snx.NXfermi_chopper])
+Detectors = make_section("Detectors", snx.NXdetector, _select_events_and_load)
```
nvaytet, line 128: If you select pulses for the detector data, shouldn't you do the same for event-mode monitor data?

src/ess/nexus/_loader.py
```diff
@@ -0,0 +1,159 @@
...
+    Basic loader component for NXentry.
+
+    Produces a basic subset of NXentry, loading NXinstrument, NXmonitor, and NXsample
+    children as well as fields (including NXlog). Subclass this in a new dataclass to
+    add sections for more NeXus classes.
+    """
+    fields: Fields
+    instrument: BasicInstrument
+    monitors: Monitors
+    sample: Sample
```
nvaytet, line 159: :+1:

tests/nexus/load_test.py
```diff
@@ -0,0 +1,132 @@
...
+    assert not hasattr(data, 'sample')
+    assert not hasattr(data.instrument, 'detectors')
+
+
+DetectorByPulse = nexus.make_section("DetectorPyPulse", snx.NXdetector,
+                                     lambda group: group.events[()])
+
+
+@dataclass
+class InstrumentWithDetectorsPyPulse(nexus.InstrumentMixin):
```
nvaytet:
```suggestion
class InstrumentWithDetectorsByPulse(nexus.InstrumentMixin):
```

tests/nexus/load_test.py
```diff
@@ -0,0 +1,132 @@
...
+    detectors: DetectorByPulse
+
+
+def test_load_instrument_without_binning_to_pixels(nxroot):
+    data = InstrumentWithDetectorsPyPulse.from_nexus(nxroot.entry)
```
nvaytet:
```suggestion
    data = InstrumentWithDetectorsByPulse.from_nexus(nxroot.entry)
```

src/ess/nexus/_loader.py
nvaytet, line 159: Not loading the source?

src/ess/nexus/_loader.py
```diff
@@ -0,0 +1,159 @@
...
+    Parameters
+    ----------
+    key:
+        Child name or child's NX_class (subclass of :py:class:`scippnexus.NXobject`).
+    load:
+        Callable to perform the loading the child group.
+    """
+
+    def from_nexus(cls, group: snx.NXobject, **kwargs):
+        c = group[key] if isinstance(key, str) else group.__getattr__(key.__name__[2:])
```
nvaytet, line 119: Is the `key.__name__[2:]` to remove the `NX` prefix? Could this be brittle logic?

tests/nexus/load_test.py
```diff
@@ -0,0 +1,132 @@
...
+    assert set(data.detectors) == {'det0', 'det1'}
+    assert 'event_time_zero' in data.detectors['det0'].coords
+
+
+NamedLeaf = nexus.make_leaf("NamedLeaf", "det1")
+
+
+def test_loaded_named_leaf(nxroot):
+    data = NamedLeaf.from_nexus(nxroot.entry.instrument)
```
nvaytet, line 130: What happens if you try to use a `make_leaf` type on something that is not a leaf? (i.e. there is still some data nested below)

SimonHeybrock: You will get an exception from scippnexus, complaining that this NX_class cannot be loaded.

src/ess/nexus/_loader.py
SimonHeybrock: This will need iterations. I don't know if `Source` always exists, or is needed?

src/ess/nexus/_loader.py
SimonHeybrock: I could not think of a good case for supporting this by default. Filtering by detector ID is extremely inefficient (it has to load all the data anyway). Beyond that data may not be large? This can be added later or by the user if they have a case for it.

src/ess/nexus/_loader.py
```diff
@@ -0,0 +1,159 @@
...
+    @classmethod
+    def from_nexus(cls: Type[T2], group: Union[snx.NXobject, PathLike], /,
+                   **kwargs: Any) -> T2:
+        """Load data from the NXentry group."""
+        if isinstance(group, snx.NXentry):
+            return _load_dataclass(group, cls, kwargs)
+        if isinstance(group, snx.NXroot):
+            return cls.from_nexus(group.entry, **kwargs)
+        with snx.File(group) as f:
+            return cls.from_nexus(f, **kwargs)
```
jl-wynen, line 53: Why can this Mixin open a file but the others can't? I think it makes more sense to require the user to open it themselves in all cases.

src/ess/nexus/_loader.py
```diff
@@ -0,0 +1,159 @@
...
+                 index=(),
+                 skip_errors=False) -> Optional[Union[sc.DataArray, dict]]:
+    # The list of caugth exceptions is copied from scippneutron.load_nexus. We will
+    # likely wanto to add better customization options.
+    try:
+        return group[index]
+    except (snx.NexusStructureError, KeyError, sc.DTypeError, ValueError,
+            IndexError) as e:
+        if not skip_errors:
+            raise e from None
```
jl-wynen: Just `raise` (without `e`) to get the full backtrace?

src/ess/nexus/_loader.py
```diff
@@ -0,0 +1,159 @@
...
+        ``[scippnexus.Field, scippnexus.NXlog]``.
+    load:
+        Callable to perform the loading of a single group. This will be called once for
+        each child matching ``key``.
+    """
+
+    def from_nexus(cls, group: snx.NXobject, **kwargs):
+        return cls(_load_multi(load, group[key], **kwargs))
+
+    return type(name, (dict, ), dict(from_nexus=classmethod(from_nexus)))
```
jl-wynen, line 102: Not sure what implications this has. Is subclassing `dict` supported? Would it make more sense to use a dataclass or NamedDict? Or would those require knowing the possible fields ahead of time?

jl-wynen (commented):
For my own understanding: Is the reason for implementing this interface that it makes assumptions about the structure of a file, e.g. an instrument knows that it contains fields and detectors as direct children?
Is this why this code is in `ess` and not `scippnexus`?

SimonHeybrock:
> For my own understanding: Is the reason for implementing this interface that it makes assumptions about the structure of a file, e.g. an instrument knows that it contains fields and detectors as direct children?
> Is this why this code is in `ess` and not `scippnexus`?

I think we may eventually decide to move this into `scippnexus`. I did it here initially, since I think we will likely need some iterations, and want to try how it integrates with implementation of data reduction workflows. Having it in `scippnexus` would tie us more to releases and complicate this.

src/ess/nexus/_loader.py
SimonHeybrock: > Is subclassing dict supported?
Supported by what?

> Would it make more sense to use a dataclass or NamedDict? Or would those require knowing the possible fields ahead of time?
Indeed, we do not know all the field names (e.g., the names of all NXdetector in the file).

jl-wynen:
But then I am unclear about what benefit this has over `group[NXInstrument]`.

src/ess/nexus/_loader.py
jl-wynen: > >Is subclassing dict supported?
> Supported by what?

By dict and the rest of the ecosystem.

SimonHeybrock: Yes, `dict` can be subclassed.

SimonHeybrock:
> But then I am unclear about what benefit this has over `group[NXInstrument]`.

You mean the additions in this PR? It avoids a significant amount of boilerplate code, I think.

jl-wynen:
> > But then I am unclear about what benefit this has over `group[NXInstrument]`.
> 
> You mean the additions in this PR? It avoids a significant amount of boilerplate code, I think.

But why not use `group[NXInstrument][()]` instead of `Instrument.from_nexus(group)`?

SimonHeybrock:
> > > But then I am unclear about what benefit this has over `group[NXInstrument]`.
> > 
> > 
> > You mean the additions in this PR? It avoids a significant amount of boilerplate code, I think.
> 
> But why not use `group[NXInstrument][()]` instead of `Instrument.from_nexus(group)`?

The former would necessarily load everything, whereas the approach implemented here allows for customization?

jl-wynen:
> The former would necessarily load everything, whereas the approach implemented here allows for customization?

Do you mean because the dataclasses here do not have to include all fields in a nexus group? If so, wouldn't this be specific to a use case?

SimonHeybrock:
> > The former would necessarily load everything, whereas the approach implemented here allows for customization?
> 
> Do you mean because the dataclasses here do not have to include all fields in a nexus group? If so, wouldn't this be specific to a use case?

Yes, the idea is that each instrument would customize this, e.g., to load (ore ignore) specific instrument components.

src/ess/nexus/_loader.py
SimonHeybrock: `scippnexus.NXmonitor` currently does not support the `events` property. Of course this could be added in the future.

src/ess/nexus/_loader.py
SimonHeybrock: Yes and no. The `NX` prefix is standardized in Nexus. But of course one could pass a Python class that does not follow this schema. One could require providing the name as a separate argument, but I feel that would be *more* likely to break in practice?

src/ess/nexus/_loader.py
SimonHeybrock: @nvaytet @jl-wynen What are your thoughts on this? Would you prefer implementing a wrapper function that opens the file?

SimonHeybrock:
I finally addressed some of the comments here. See answers and updates. As mentioned face-to-face, I am certain that this will require iterations. For now, we can avoid documenting and advertising it widely such that we can try it in practice for a small subset of our applications.

SimonHeybrock:
I cannot reproduce this error locally. Tried both the latest `scippnexus` and 0.3.2. Can one of you reproduce this?

jl-wynen:
No, the tests pass on my workstation

SimonHeybrock:
The CI fail comes from a NumPy Deprecation warning, triggered by and old h5py (h5py=3.1.0). This was fixed one and a half years ago in h5py=3.2.0. Can we update @nvaytet, or is something (Mantid?) holding us back?

nvaytet:
Yes, Mantid is forcing us to pin I think, although I have not tried in a while, but `hdf5=1.10` is still required by the mantid conda recipe. See the discussion in Slack:
Screenshot at 2022-11-14 12-23-54

SimonHeybrock:
So.... should we suppress the warning instead? Or can we avoid the Mantid dependency in the CI?

nvaytet:
Not sure, I've just asked on the Mantid slack if we could move `hdf5` to a more recent version. We'll see what happens.

SimonHeybrock:
I don't think a change there would happen on a timescale that works for us. I think we should begin considering dropping Mantid from the CI, or finding ways to split things.

nvaytet (approved):
At this point, I feel further discussion would not be the most productive. It looks ok to me, and I think we should just try it out in practise, this will tell us much quicker whether the approach needs substantial changes or not.

PR #140, opened 2022-09-28
Update to scipp 0.16.3 and scippneutron 0.9.3

jl-wynen:
Failing because of a bug identified in scipp and fixed in https://github.com/scipp/scipp/pull/2821


.buildconfig/ci.yml
```diff
@@ -223,24 +240,25 @@ dependencies:
...
+  - qt=5.12.9
   - rapidjson=1.1.0
   - readline=8.1.2
   - requests=2.28.1
-  - scipp=0.16.2
-  - scippneutron=0.9.2
-  - scippnexus=0.1.8
-  - scipy=1.9.0
-  - setuptools=65.3.0
+  - scipp=0.16.3
```
SimonHeybrock: Won't you need 0.16.4?

SimonHeybrock (commented):
Note: `pip-compile-multi` has an option to only update a specific package. We should generally use that when targeting a specific fix, I think?

.buildconfig/ci.yml
jl-wynen: Yes, once that actually exists.

jl-wynen:
> Note: `pip-compile-multi` has an option to only update a specific package. We should generally use that when targeting a specific fix, I think?

This uses conda, not pip-compile. Is there such an option, too?

SimonHeybrock:
> > Note: `pip-compile-multi` has an option to only update a specific package. We should generally use that when targeting a specific fix, I think?
> 
> This uses conda, not pip-compile. Is there such an option, too?

Hmm, probably not.

SimonHeybrock (approved):

nvaytet (approved):

PR #147, opened 2022-11-30
Py separation

gudlot:
Here is my latest version for the NURF project.

There are 4 .py files
- plot.py: Plot functions
- fluo.py: Fluo-related functions
- utils.py: General functions applicable to both exp. methods
- uv.py: UV-related functions
There are three .ipynb for demonstration, that need more cleaning, and the SANS version needs more TLC.
-fluo_demonstration.ipynb
-uv_demonstration.ipynb
-SANS_demonstration.ipynb

There is as well:
- ill_auxilliary_funcs.py: Contains `nurf_file_creator` function that outlines how NURF data should be organised in  Nexus files. 
- ill_to_loki.ipynb: Is the jupyter notebook based on ill_auxilliary_funcs.py
A future NURF module at ESS does not need these files because you will have your own Nexus files, but I provide them because the test files from ILL require them to convert the data to the new format.


wpotrzebowski:
@gudlot it seems that this pull request considerably overlaps with the older one (#129). Should #129 be closed or still need to be reviewed and merged?

gudlot:
Sorry, I have not seen your comment @wpotrzebowski earlier.
Here I implemented the changes that were suggested in #129 I would suggest continuing from here.  I don't know what is the current status of this project.

src/ess/loki/nurf/utils.py
```diff
@@ -0,0 +1,287 @@
...
+        scipp.ndimage 
+        This function filters only along the wavelength direction, not along the spectrum direction.
+        Kernel_size could be given as sc.scalar(value:float, unit='nm'), but only if data is equally spaced.
+        Default kernel_size in wavelength direction: 3
+        #TODO: Take care of this option, when hardware is ready for integration.
+        If not, and currently this is the case for the spectrometer, kernel_size has to be int, odd or even.
+        I don't check for int because scipp does it.
+    """
+
+    if not ('spectrum' and 'wavelength') in da.dims:
```
dmytrokyrychuk: This condition only checks for the presence of `'wavelength'`.

```python
>>> spectrum = {'spectrum': 1}
>>> wavelength = {'wavelength': 1}
>>> both = {'spectrum': 1, 'wavelength': 1}
>>> print(('spectrum' and 'wavelength') in spectrum)
False
>>> print(('spectrum' and 'wavelength') in wavelength)
True
>>> print(('spectrum' and 'wavelength') in both)
True
```

I think you meant to write `if 'spectrum' not in da.dims or 'wavelength' not in da.dims`

gudlot: Hi @dmytrokyrychuk 
Thank you very much for pointing this out. I fell for a syntax that Python does not support. I replaced your suggestion with the `issubset()`-method:

```py
if not {'spectrum', 'wavelength'}.issubset(da.dims):
        raise ValueError('Dimensions spectrum and wavelength expected.')
```

What do you think about this? 
You are correct. I want both `spectrum` and `wavelength` as dimensions in the scipp.DataArray.  Your suggestion would work, of course, too. 
I should have implemented tests, but my contract finished. 

In addition, I fixed changes in the API of scipp, in `utils.py`, `fluo.py`, 'uv.py'. But on the TODO list remain now some issues in the `plot.py`.

dmytrokyrychuk: Your suggestion is valid. I like my suggestion because it reads like a sentence.

Please note that I am not affiliated with this project. I was simply stumbled upon this PR and reported an issue I noticed.

PR #150, opened 2023-01-03
Make CI consistent with other Scipp projects

### LINKED ISSUE #149: Update project to use same setup as scippnexus / scippneutron
Now that scippneutron provides a PyPI package we should be able to use a tox-based setup and make PyPI packages as well. Mantid is required in some places so approach may be closer to what is implemented in scippneutron.

nvaytet:
- Use `tox` for static analysis
- Use `setup.cfg` and `pyproject.toml`, remove `setup.py`
- Build on ubuntu, macos & windows

Fixes #149


.github/workflows/pr_and_main.yml
```diff
@@ -19,44 +19,53 @@ defaults:
...
-    env:
-      DOCS_HTML_DIR: ${{ github.workspace }}/docs_html
+    name: build ${{ matrix.os }}, py${{ matrix.python-version }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ["ubuntu-22.04", "macos-11", "windows-2019"]
+        python-version: ["3.8"]
+    # env:
```
jl-wynen: Remove?

setup.cfg
```diff
@@ -1,3 +1,43 @@
+[metadata]
```
jl-wynen, line 1: Most of these can be moved to pyproject.toml. See, e.g., https://github.com/scipp/scippneutron/blob/main/pyproject.toml
 A few fields are not supported yet. But while you are at it, you could move everything you can at this point.

pyproject.toml
```diff
@@ -0,0 +1,23 @@
+[build-system]
+requires = [
+  "setuptools>=42",
+  "setuptools_scm[toml]>=6.2",
+]
+build-backend = "setuptools.build_meta"
+
+[tool.setuptools_scm]
+
+[tool.pytest.ini_options]
```
jl-wynen, line 10: scippneutron is a bit stricter now (https://github.com/scipp/scippneutron/blob/b717619590c29922852f4c15c291aa413e8f483e/pyproject.toml#L43) and I think we should use these options in all projects.

setup.cfg
nvaytet: In Plopp and scippnexus they are in the setup. Is it better to have in the pyproject?

jl-wynen: As I understand it, the plan (by the Python devs) is to move everything into pyproject.toml and remove setup.(cfg|py) completely. So eventually, we have to move. But we still have years left.

tox.ini
```diff
@@ -0,0 +1,18 @@
...
+[testenv:static]
+description = Code formatting and static analysis
+basepython = python3.8
+skip_install = true
+deps = -r requirements/static.txt
+allowlist_externals = find
+commands =
+  nbstripout --extra-keys 'metadata.language_info.version cell.metadata.jp-MarkdownHeadingCollapsed' --drop-empty-cells .
+  yapf --recursive --in-place src tests
+  flake8 .
```
SimonHeybrock, line 18: I have also added a `deps` env in some projects:

https://github.com/scipp/scipp/blob/31868a165bc52785551be2a22648d1f1368d6fa3/tox.ini#L72-L78

SimonHeybrock (approved):

PR #161, opened 2023-01-26
Remove deprecation warnings from docs notebooks

### LINKED ISSUE #148: Remove deprecation warnings in SANS workflows
Screenshot at 2022-12-01 10-10-20

nvaytet:
This was both deprecation warnings from using the old plotting and from using `histogram` instead of `hist`.

Fix #148


docs/instruments/loki/sans2d_reduction.ipynb
```diff
@@ -496,7 +495,8 @@
    "mimetype": "text/x-python",
    "name": "python",
    "nbconvert_exporter": "python",
-   "pygments_lexer": "ipython3"
+   "pygments_lexer": "ipython3",
+   "version": "3.8.12"
```
SimonHeybrock: Shouldn't our pre-commit or static CI strip the version? Is the config outdated here?

SimonHeybrock (approved):

docs/instruments/loki/sans2d_reduction.ipynb
nvaytet: Strange, it should be using [pre-commit](https://github.com/scipp/ess/blob/main/tox.ini#L18) via `tox`, and pre-commit is configured to use [nbstripout](https://github.com/scipp/ess/blob/main/.pre-commit-config.yaml#L28)...

PR #167, opened 2023-02-21
Drop monitor variances in SANS

### LINKED ISSUE #162: Update SANS workflow to discard normalization variances depending on threshold
Result of discussion with Judith and @wpotrzebowski:

- Normalization terms in SANS almost always have very good statistics.
- We should update the SANS workflows to set `variances = None` in the relevant places.
  - Before this, add a check for `alpha`, defined as in our paper, and raise an error if a threshold is exceeded.
  - Exact threshold is to be defined, but we can start with something on the safe side and then iterate.
  - May consider issuing a warning if we get close to the threshold (but not above)?

jl-wynen:
Fixes #162

I chose the threshold quite arbitrarily by looking at Fig 2 in out paper. Is it reasonable?

Compared to the issue, this will always go through, no matter the value of alpha. Does this make sense or should it better raise an exception if alpha is too large? What is too large?

Interestingly, there one figure changed in this, the plot in cell 25 of 'SANS2D: I(Q) workflow for a single run (sample)' is as shown below. It has visible error bars for high Q while the old plot does not. Hope this is a good sign...
Figure 2

### DIFF
--- a/docs/instruments/loki/sans2d_to_I_of_Q.ipynb
+++ b/docs/instruments/loki/sans2d_to_I_of_Q.ipynb
@@ -148,6 +148,31 @@
   },
   {
    "cell_type": "markdown",
+   "id": "ef04d51b-f6de-406b-85f6-f944ebc5f0c4",
+   "metadata": {},
+   "source": [
+    "Monitors will be broadcast to all pixels of the detectors and this introduces correlations if the monitors have variances.\n",
+    "So simply remove variances from the monitors.\n",
+    "This is only correct as long as the monitors have many more counts than the detectors."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "323d8ecc-82ff-43bd-989a-ad201d624e3d",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "monitors = {\n",
+    "    'sample': sans.i_of_q.drop_monitor_variances(\n",
+    "        monitors['sample'], data=sample),\n",
+    "    'direct': sans.i_of_q.drop_monitor_variances(\n",
+    "        monitors['direct'], data=direct),\n",
+    "}"
+   ]
+  },
+  {
+   "cell_type": "markdown",
    "id": "e3b0b991-85c4-40fe-a903-74b43716a155",
    "metadata": {},
    "source": [
@@ -769,7 +794,8 @@
    "mimetype": "text/x-python",
    "name": "python",
    "nbconvert_exporter": "python",
-   "pygments_lexer": "ipython3"
+   "pygments_lexer": "ipython3",
+   "version": "3.8.12"
   }
  },
  "nbformat": 4,
--- a/src/ess/sans/i_of_q.py
+++ b/src/ess/sans/i_of_q.py
@@ -1,13 +1,14 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
 
-from typing import Tuple, Union
+from typing import Tuple, Union,Dict
 
 import scipp as sc
 from scipp.scipy.interpolate import interp1d
 
 from . import conversions, normalization
 from .common import gravity_vector
+from ..uncertainty import drop_variances
 
 
 def make_coordinate_transform_graphs(gravity: bool) -> Tuple[dict, dict]:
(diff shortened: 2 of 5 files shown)


SimonHeybrock:
> Compared to the issue, this will always go through, no matter the value of alpha. Does this make sense or should it better raise an exception if alpha is too large? What is too large?

I think we should raise above some limit, otherwise this will just be systematically get ignored. Initially I would suggest to chose a rather "strict" limit. This can than be gradually increased if the scientist verify that it is still a safe region.

> Interestingly, there one figure changed in this, the plot in cell 25 of 'SANS2D: I(Q) workflow for a single run (sample)' is as shown below. It has visible error bars for high Q while the old plot does not. Hope this is a good sign...

This makes no sense. Are you sure this is not due to an unrelated change?

docs/instruments/loki/sans2d_to_I_of_Q.ipynb
```diff
@@ -146,6 +146,31 @@
...
    ]
   },
+  {
+   "cell_type": "markdown",
+   "id": "ef04d51b-f6de-406b-85f6-f944ebc5f0c4",
+   "metadata": {},
+   "source": [
+    "Monitors will be broadcast to all pixels of the detectors and this introduces correlations if the monitors have variances.\n",
+    "So simply remove variances from the monitors.\n",
+    "This is only correct as long as the monitors have many more counts than the detectors."
```
SimonHeybrock, line 156: correct -> a good approximation?

src/ess/uncertainty.py
```diff
@@ -0,0 +1,68 @@
...
+            'may be underestimated.', f"of '{name}'" if name else '', alpha)
+    return sc.values(da)
+
+
+def _compute_alpha(numerator: Union[sc.Variable, sc.DataArray],
+                   denominator: Union[sc.Variable, sc.DataArray]) -> float:
+    alpha = numerator.sum().data / denominator.sum().data
+    if alpha.unit != 'one':
+        raise sc.UnitError(
+            'Cannot compare counts, the reference has a different unit from the data.')
```
SimonHeybrock, line 67: I think the check here assumes that `variances = values`, i.e., raw counts. I think this is often the case, but not always. Can this condition and check be reformulated for other cases?

src/ess/uncertainty.py
```diff
@@ -0,0 +1,68 @@
...
+    """
+    alpha = _compute_alpha(numerator=reference, denominator=da)
+    logger = get_logger()
+    msg = 'Dropping variances%s. Total counts divided by reference: alpha = %f'
+    if alpha < threshold:
+        logger.info(msg, f" of '{name}'" if name else '', alpha)
+    else:
+        logger.warning(
+            msg + ' This value may be too large and the uncertainties of the result '
+            'may be underestimated.', f"of '{name}'" if name else '', alpha)
```
SimonHeybrock, line 58: I think we should have a second threshold above which we raise.

nvaytet:
> > Interestingly, there one figure changed in this, the plot in cell 25 of 'SANS2D: I(Q) workflow for a single run (sample)' is as shown below. It has visible error bars for high Q while the old plot does not. Hope this is a good sign...
> 
> This makes no sense. Are you sure this is not due to an unrelated change?

Yes, I think this is an issue with the plotting, where the old plotting was not showing the error bars (some bug), but when using Plopp we see the error bars again.

src/ess/sans/i_of_q.py
```diff
@@ -239,8 +271,9 @@ def to_I_of_Q(data: sc.DataArray,
         that contribute to different regions in Q space.
     """
 
-    monitors = _make_dict_of_monitors(data_monitors=data_monitors,
-                                      direct_monitors=direct_monitors)
+    monitors = _make_dict_of_monitors(
+        data_monitors=drop_monitor_variances(data_monitors, data=data),
```
nvaytet, line 275: In our paper, we define alpha as the ratio of data counts to `sample_transmission_monitor` counts. But actually, the term that gets broadcast to the pixel dimensions is the transmission fraction `T` times the `sample_incident_monitor`.

This equates to `sample_transmission_monitor * direct_incident_monitor / direct_transmission_monitor`.
Hence, shouldn't we, both here and in the paper, compute `alpha` using this whole term, instead of just the `sample_transmission_monitor` (in the paper), or an `alpha` for each monitor (as here in the code)?

src/ess/uncertainty.py
```diff
@@ -0,0 +1,68 @@
...
+from .logging import get_logger
+
+T = TypeVar("T", bound=Union[sc.Variable, sc.DataArray])
+
+
+def drop_variances(da: T,
+                   *,
+                   reference: Union[sc.Variable, sc.DataArray],
+                   name: str = '',
+                   threshold: float = 1.0) -> T:
```
nvaytet, line 18: In our paper we have the condition `alpha << 1`, so I would say the threshold should be more something like 0.1?
What does `alpha` evaluate to for this experimental data?

nvaytet:
Closing in favour of #170

PR #169, opened 2023-03-01
Beam centre finder

nvaytet:
This PR adds the `beam_center` function to find the beam center in a SANS experiment.
The prodecure to determine the precise location of the beam center is the following:

1. obtain an initial guess by computing the center-of-mass of the pixels, weighted by the counts on each pixel
2. from that initial guess, divide the panel into 4 quadrants
3. compute $I(Q)$ inside each quadrant and compute the residual difference between all 4 quadrants
4. iteratively move the centre position and repeat 2. and 3. until all 4 $I(Q)$ curves lie on top of each other

This is described in detail in a new notebook, and is used in the existing SANS notebooks.

I do not know how to add good unit tests for the beam center finder... Suggestions welcome.


nvaytet:
@wpotrzebowski the final $I(Q)$results have changed, because we are masking both numerator and denominator now (this was an oversight in the old code).
We should carefully go over the results to make sure I haven't messed up.

src/ess/sans/beam_center.py
```diff
@@ -0,0 +1,187 @@
...
+import scipp as sc
+
+from ..logging import get_logger
+from . import i_of_q
+from .common import gravity_vector
+from .conversions import sans_elastic
+from .normalization import normalize
+
+
+def _center_of_mass(data):
```
SimonHeybrock: Type hints missing everywhere

src/ess/sans/beam_center.py
```diff
@@ -0,0 +1,187 @@
...
+    """
+    Find the center of mass of the data counts.
+    """
+    summed = data.sum(list(set(data.dims) - set(data.meta['position'].dims)))
+    v = sc.values(summed.data)
+    com = sc.sum(summed.meta['position'] * v) / v.sum()
+    return com.fields.x, com.fields.y
```
SimonHeybrock: This assumes that the beam direction is `z`. Should this be checked somewhere?

src/ess/sans/beam_center.py
```diff
@@ -0,0 +1,187 @@
...
+
+def _cost(xy, data):
+    """
+    Cost function for determining how close the I(Q) curves are in all four quadrants.
+    """
+    ref = data['right']
+    c = ((data['top'] - ref)**2 + (data['left'] - ref)**2 +
+         (data['bottom'] - ref)**2) / ref**2
+    out = c.sum().value
+    print(xy, out)
```
SimonHeybrock:
```suggestion
```
?

src/ess/sans/beam_center.py
```diff
@@ -0,0 +1,187 @@
...
+        denominator.coords[c] = data.coords[c]
+    denominator.masks['circle'] = data.masks['circle']
+
+    pi = sc.constants.pi.value
+    phi_offset = sc.scalar(pi / 4, unit='rad')
+    phi = (sc.atan2(y=data.coords['position'].fields.y,
+                    x=data.coords['position'].fields.x) +
+           phi_offset) % (2 * (pi * sc.units.rad))
+    phi_bins = sc.linspace('phi', 0, pi * 2, 5, unit='rad')
+    quadrants = ['right', 'top', 'left', 'bottom']
```
SimonHeybrock: Not sure I follow entirely, but given the pi/4 and naming (right, top, ...), it looks like you make an `X` cut into quadrants? I imagine this is fine for square detector banks, but what if the bank is rectangular? Wouldn't you get very different pixel counts in the (left/right) vs (top/bottom) quadrants?

Can you make a `+` cut instead to avoid this?

src/ess/sans/beam_center.py
```diff
@@ -0,0 +1,187 @@
...
+            wavelength_bands):
+    """
+    Compute the intensity as a function of Q inside 4 quadrants in Phi.
+    Return the sum of the squares of the relative differences between the 4 quadrants.
+    """
+    # Make a copy of the original data
+    data = sc.DataArray(data=sample.data)
+    coord_list = ['position', 'sample_position', 'source_position']
+    for c in coord_list:
+        data.coords[c] = sample.meta[c].copy(deep=True)
```
SimonHeybrock: - Why are you not using `data=data.copy()`? Is it do avoid copying wrong wavelength/Q coords from previous iterations?
- `data = sc.DataArray(data=sample.data)` will not make a copy of the data values, is that intentional? Note that if you have event data then this will also affect event coords.
- Are there no prior masks (such as TOF, or dead pixels) that need to be copied?

src/ess/sans/conversions.py
```diff
@@ -41,17 +41,23 @@ def two_theta(gravity: sc.Variable, wavelength: sc.Variable, incident_beam: sc.V
     return out
 
 
-def sans_elastic(gravity: bool = False) -> dict:
+def phi(position: sc.Variable) -> sc.Variable:
```
SimonHeybrock: Should at the very least document that is for a beam direction along `z`. I do not know how precisely this is fulfilled in practice.

src/ess/sans/i_of_q.py
```diff
@@ -142,61 +144,145 @@ def convert_to_q_and_merge_spectra(data: sc.DataArray, graph: dict,
...
+    if da.bins is not None:
+        da = da.bin({mask.dim: mask.coords[mask.dim]})
+        da.masks[name] = mask.data
+    else:
+        lu = sc.lookup(mask, mask.dim)
+        if da.coords.is_edges(mask.dim):
+            sampling = sc.midpoints(da.coords[mask.dim])
+        else:
+            sampling = da.coords[mask.dim]
+        da.masks[name] = lu[sampling]
```
SimonHeybrock: Hmm, is considering the *midpoints* the correct solution? Shouldn't we mask all bins that have *overlap* with a masked region instead? Something like

```python
dim = mask.dim
edges = sc.lookup(mask, dim)[da.coords[dim]]
da.masks[name] = edges[dim, 1:] | edges[dim, :-1]
```

SimonHeybrock (commented):
Some comments for now, I did not look at everything in detail yet.

src/ess/sans/conversions.py
nvaytet: hmm nice catch, we are not using this anymore. I forgot to remove it.

SimonHeybrock: There are also other places where the same assumption is made in this code.

src/ess/sans/beam_center.py
nvaytet: Nice observations. I tested a `+` cut instead and I get nice results, so I will go with that. Thanks.

src/ess/sans/beam_center.py
nvaytet: This should go to the logger :+1:

src/ess/sans/beam_center.py
nvaytet: > Is it do avoid copying wrong wavelength/Q coords from previous iterations?

Yes

> will not make a copy of the data values, is that intentional?

Yes, I don't think we need a deep copy.

> Note that if you have event data then this will also affect event coords.

Not sure what you mean by `this will also affect event coords`.

> Are there no prior masks (such as TOF, or dead pixels) that need to be copied?

hmm yes, masks are missing! Thanks. I would like to change to use `da.copy()` and then `da.coords.clear()` once I can use the latest version of Scipp, but we first need to deal with the variances broadcast issue before we can do this.

wpotrzebowski:
Thanks @nvaytet. I will try to take a look at it tomorrow. I think I need to setup dev environment first to test it. Is this https://github.com/scipp/ess/blob/main/environment.yml file I should use?

src/ess/sans/beam_center.py
SimonHeybrock: > Not sure what you mean by this will also affect event coords

I mean, not copying the data implies not copying the event coords (since they are "within" the data values).

src/ess/sans/beam_center.py
nvaytet: fixed

src/ess/sans/i_of_q.py
nvaytet: I made a function `mask_range` which basically implements what you suggest, and works for both binned and dense data. Not sure if that should go somewhere more general than in the SANS submodule, as it's fully generic.

src/ess/sans/conversions.py
nvaytet: I left this in the end and generalized it so that we don't assume the beam is along `z`.

nvaytet:
@SimonHeybrock @wpotrzebowski i think this is ready for another round of review. Thanks!

src/ess/sans/beam_center.py
nvaytet: I tried to make it general in terms of normal to the direction of the beam. please check

src/ess/sans/common.py
```diff
@@ -11,3 +14,53 @@ def gravity_vector() -> sc.Variable:
...
+    The provided edges are used to define the ranges to be masked.
+
+    Parameters
+    ----------
+    da:
+        The data array to be masked.
+    edges:
+        The edges of the ranges to be masked.
+    mask:
+        The mask values to be applied within the ranges defined by ``edges``.
```
SimonHeybrock: How about using `DataArray`, with the mask as the values? This may anyway come out of upstream math operations.

src/ess/sans/common.py
```diff
@@ -11,3 +14,53 @@ def gravity_vector() -> sc.Variable:
...
+    Returns
+    -------
+    :
+        A copy of the input data array with the mask applied.
+    """
+    dim = edges.dim
+    if edges.sizes[dim] != mask.sizes[dim] + 1:
+        raise RuntimeError(
+            f"Size of {dim} in edges ({edges.sizes[dim]}) does not match "
+            f"size of {dim} in mask ({mask.sizes[dim]})")
```
SimonHeybrock: See above, can be removed if you use a `DataArray`, or replaced by bin-edge check.

src/ess/sans/common.py
```diff
@@ -11,3 +14,53 @@ def gravity_vector() -> sc.Variable:
...
+    if name is None:
+        name = uuid.uuid4().hex
+    lu = sc.DataArray(data=mask, coords={dim: edges})
+    if da.bins is not None:
+        if dim in da.coords:
+            new_bins = sc.sort(sc.concat([da.coords[dim], edges], dim=dim), key=dim)
```
SimonHeybrock: Will there be trouble with empty bins, of mask edges coincide with data edges?

src/ess/sans/conversions.py
```diff
@@ -7,15 +7,60 @@
...
+    """
     grav = sc.norm(gravity)
     L2 = sc.norm(scattered_beam)
-    y = sc.dot(scattered_beam, gravity) / grav
-    n = sc.cross(incident_beam, gravity)
-    n /= sc.norm(n)
-    x_term = sc.dot(scattered_beam, n)
-    x_term *= x_term
+
+    # TODO: Using cylindrical_x means we can no longer raise to the power of 2 in place
```
SimonHeybrock: Why is there a TODO here?

src/ess/sans/conversions.py
```diff
@@ -7,15 +7,60 @@
...
+                  scattered_beam: sc.Variable) -> sc.Variable:
+    """
+    Compute the vertical y coordinate perpendicular to the incident beam direction.
+    Note that it is assumed here that the incident beam is perpendicular to the gravity
+    vector.
+    """
+    return sc.dot(scattered_beam, cyl_y_unit_vector)
+
+
+def two_theta(cylindrical_x: sc.Variable, cylindrical_y: sc.Variable,
```
SimonHeybrock: A bit odd to not have `incident_beam` as input here. Is there a reason why you don't compute x and y internally?

src/ess/sans/i_of_q.py
```diff
@@ -1,90 +1,80 @@
...
+                                         non_background_range=non_background_range)
+            for key, mon in monitor.items()
+        })
+
+    if 'wavelength' not in monitor.dims:
+        monitor = monitor.transform_coords('wavelength',
+                                           graph=conversions.sans_monitor())
+
+    if monitor.variances is not None:
+        monitor.variances = None  # TODO: Hack to set variances to None
```
SimonHeybrock, line 61: Unaddressed TODO

src/ess/sans/i_of_q.py
```diff
@@ -142,134 +154,199 @@ def convert_to_q_and_merge_spectra(data: sc.DataArray, graph: dict,
...
     for i in range(wavelength_bands.sizes['wavelength'] - 1):
         band = data_q['wavelength', wavelength_bands[i]:wavelength_bands[i + 1]]
-        bands.append(band.hist({q_bins.dim: q_bins}).sum('spectrum'))
+        bands.append(band.hist({q_bins.dim: q_bins}).sum(sum_dims))
     q_summed = sc.concat(bands, 'wavelength')
     return q_summed
 
 
-def _make_dict_of_monitors(data_monitors, direct_monitors):
+def add_mask(da: sc.DataArray, mask: sc.DataArray, name: str) -> sc.DataArray:
```
SimonHeybrock: Isn't there a new function for this in another module now?

src/ess/sans/common.py
nvaytet: I did that at first, but I thought that was maybe a weird or unusual way of storing a mask?
By using separate keyword args, I thought it made using the function easier if you have not seen it before?

That said, by

> This may anyway come out of upstream math operations.

do you mean things like `mask = da1 > da2` would mean that `mask` is already a DataArray with a `wavelength` coord? But then you would have to specify in the function args which dimension/coord should be used to define the range?
We could allow a data array with only a single dim, and then assume that it has the correct dimension coord?
Not sure which I prefer, I'm not sure giving the edges and mask values separately is so much worse...

src/ess/sans/i_of_q.py
nvaytet: This one will be addressed in a future PR (#167), which checks the value of `alpha` before dropping the variances. For now, we drop them and I will adapt Jan-lukas's PR once we merge this one.

src/ess/sans/i_of_q.py
nvaytet: nice catch :+1:

src/ess/sans/conversions.py
nvaytet: Because I wanted to ask you / draw your attention to this part, to see if the slightly less efficient implementation was ok. As it turns out, your suggestion below allowed me to fix this (i think!).

src/ess/sans/conversions.py
nvaytet: I changed this, and it helped fix your question above.

src/ess/sans/common.py
nvaytet: Yes, they would be! Fixed it (I hope), thanks

src/ess/sans/common.py
SimonHeybrock: > I did that at first, but I thought that was maybe a weird or unusual way of storing a mask?

You are right that we have not done it much, but I do not think it is weird at all. It is actually safer, since it avoids bugs from defining the mask with different binning, etc.

src/ess/sans/common.py
```diff
@@ -50,7 +51,13 @@ def mask_range(da: sc.DataArray,
...
     if da.bins is not None:
         if dim in da.coords:
-            new_bins = sc.sort(sc.concat([da.coords[dim], edges], dim=dim), key=dim)
+            not_in_coord = np.setdiff1d(edges.values, da.coords[dim].values)
+            new_bins = sc.sort(sc.concat([
+                da.coords[dim],
+                sc.array(dims=edges.dims, values=not_in_coord, unit=edges.unit)
+            ],
+                                         dim=dim),
+                               key=dim)
```
SimonHeybrock: Might be simpler with https://numpy.org/doc/stable/reference/generated/numpy.unique.html?

src/ess/sans/conversions.py
```diff
@@ -59,8 +58,8 @@ def two_theta(cylindrical_x: sc.Variable, cylindrical_y: sc.Variable,
     grav = sc.norm(gravity)
     L2 = sc.norm(scattered_beam)
 
-    # TODO: Using cylindrical_x means we can no longer raise to the power of 2 in place
-    x_term = cylindrical_x * cylindrical_x
+    x_term = cylindrical_x(cyl_x_unit_vector(gravity, incident_beam), scattered_beam)
```
SimonHeybrock, line 61: A bit odd to have `cylindrical_x` which does the same as `cylindrical_y`... I suppose that is just because you want to use them in `transform_coords`?

wpotrzebowski (commented):
Few comments about functionality 
- We don't need BCF for every data set. As I understand this notebooks serves as an example of existing functionality rather than something that serves as a user workflow
- Tof bins are no longer used (at least explicetly) - I guess it is because we don't use DataSet any more? 
- Pixel height and width are now incorporated into files is it the same true for sample_pos_z_offset  and monitor4_pos_z_offset? 

Comment to code:
In sans2d_reduction.ipynb masking is  done on sample directly (sample.masks['holder_mask'] = holder_mask). Shouldn't be dg['sample']?

src/ess/sans/common.py
nvaytet: Actually there is `union1d` which does exactly what I want ;-)

nvaytet:
> Few comments about functionality
> 
> We don't need BCF for every data set. As I understand this notebooks serves as an example of existing functionality rather than something that serves as a user workflow

Yes

> Tof bins are no longer used (at least explicetly) - I guess it is because we don't use DataSet any more?

Correct (and we will use `DataGroup` in the very near future, once we address the issue of the variances being broadcast).

> Pixel height and width are now incorporated into files is it the same true for sample_pos_z_offset and monitor4_pos_z_offset?

Yes

> 
> Comment to code:
> In sans2d_reduction.ipynb masking is done on sample directly (sample.masks['holder_mask'] = holder_mask). Shouldn't be dg['sample']?

They are the same python object, so it should not matter. But I need to check about maybe also masking the background run...

src/ess/sans/common.py
nvaytet: I changed it to use a `DataArray`. See latest commits.

src/ess/sans/beam_center.py
```diff
@@ -0,0 +1,209 @@
...
+    return sc.sum(summed.meta['position'] * v) / v.sum()
+
+
+def _cost(data: Dict[str, sc.DataArray]) -> float:
+    """
+    Cost function for determining how close the I(Q) curves are in all four quadrants.
+    """
+    ref = data['north-east']
+    c = ((data['north-west'] - ref)**2 + (data['south-west'] - ref)**2 +
+         (data['south-east'] - ref)**2) / ref**2
```
SimonHeybrock: This it biased and not symmetric. One would expect a result like
```
AB
AA
```
to have the same cost as others such as
```
BA
AA
```
but this is not the case in this implementation.

src/ess/sans/beam_center.py
```diff
@@ -0,0 +1,209 @@
...
+        run.
+    direct_monitors:
+        The data arrays for the incident and transmission monitors for the direct
+        run.
+    wavelength_bins:
+        The binning in the wavelength dimension to be used.
+    q_bins:
+        The binning in the Q dimension to be used.
+    masking_radius:
+        The radius of the circular mask to apply to the data while iterating.
```
SimonHeybrock: Which circular mask? To mask the beam stop?

src/ess/sans/common.py
```diff
@@ -11,3 +15,59 @@ def gravity_vector() -> sc.Variable:
...
+    """
+    dim = mask.dim
+    edges = mask.coords[dim]
+    if not mask.coords.is_edges(dim):
+        raise sc.DimensionError(
+            f'Coordinate {dim} must be bin-edges to mask a range, found midpoints.')
+    if (dim in da.coords) and (da.coords[dim].ndim > 1):
+        raise sc.DimensionError(
+            'Cannot mask range on data with multi-dimensional coordinate. '
+            f'Found dimensions {da.coords[dim].dims} for coordinate {dim}.')
```
SimonHeybrock, line 57: Is it worth checking if there is an existing mask of the same name? Or else we risk dropping it silently.

docs/techniques/sans/sans-beam-center-finder.ipynb
```diff
@@ -0,0 +1,705 @@
...
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "sample = loki.io.load_sans2d(filename=loki.data.get_path('SANS2D00063114.hdf5'))\n",
+    "direct = loki.io.load_sans2d(filename=loki.data.get_path('SANS2D00063091.hdf5'))\n",
+    "# Add X, Y coordinates\n",
+    "sample.coords['x'] = sample.coords['position'].fields.x\n",
+    "sample.coords['y'] = sample.coords['position'].fields.y\n",
```
SimonHeybrock, line 97: Why is this necessary?

docs/techniques/sans/sans-beam-center-finder.ipynb
```diff
@@ -0,0 +1,705 @@
...
+    "p.ax.plot(xc.value, yc.value, 'o', color='red', ms=5)\n",
+    "p"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "afbf9a8c-4fc7-4eb3-a477-111baa047b18",
+   "metadata": {},
+   "source": [
+    "## Making 4 quadrants\n",
```
SimonHeybrock, line 252: Is all of this repeating the code from the Python module?

src/ess/sans/beam_center.py
```diff
@@ -0,0 +1,209 @@
...
+    Description of the procedure:
+
+    #. obtain an initial guess by computing the center-of-mass of the pixels,
+       weighted by the counts on each pixel
+    #. from that initial guess, divide the panel into 4 quadrants
+    #. compute :math:`I(Q)` inside each quadrant and compute the residual difference
+       between all 4 quadrants
+    #. iteratively move the centre position and repeat 2. and 3. until all 4
+       :math:`I(Q)` curves lie on top of each other
+
```
SimonHeybrock: We had talked about documenting all of the things that may be done wrong, i.e., all the things you learning during implementation. Shouldn't this be documented here somewhere?

src/ess/sans/beam_center.py
nvaytet: What would you suggest? Compute the mean of all 4 and use that as `ref`?

SimonHeybrock: That could work, yes.

src/ess/sans/beam_center.py
nvaytet: It's this mask:
Screenshot at 2023-03-23 13-10-51

I added more details in the docstring.

docs/techniques/sans/sans-beam-center-finder.ipynb
nvaytet: Convenience. We hist in x and y to make images of the detector panel, instead of using the instrument view which increases size bloat of the docs pages.

docs/techniques/sans/sans-beam-center-finder.ipynb
nvaytet: Yes, it was. I was never happy with the duplication.
I now tried to do what I did in the other notebooks, which is to break up the BCF code into smaller functions, and use those inside the notebook.

src/ess/sans/beam_center.py
nvaytet: As discussed, I added some notes in the docstring of the `beam_center` function.

src/ess/sans/beam_center_finder.py
```diff
@@ -0,0 +1,392 @@
...
+    Returns
+    -------
+    :
+        The sum of the residuals for :math:`I(Q)` in the 4 quadrants, with respect to
+        the mean :math:`I(Q)` in all quadrants.
+    """  # noqa: E501
+    iofq = iofq_in_quadrants(xy, *args)
+    all_q = sc.concat(list(iofq.values()), dim='quadrant')
+    ref = sc.values(all_q.mean('quadrant'))
+    c = sc.abs(all_q - ref) / ref
```
SimonHeybrock: Didn't you have squares in the old implementation?

src/ess/sans/beam_center_finder.py
```diff
@@ -0,0 +1,392 @@
...
+    #. obtain an initial guess by computing the center-of-mass of the pixels,
+       weighted by the counts on each pixel
+    #. from that initial guess, divide the panel into 4 quadrants
+    #. compute :math:`I(Q)` inside each quadrant and compute the residual difference
+       between all 4 quadrants
+    #. iteratively move the centre position and repeat 2. and 3. until all 4
+       :math:`I(Q)` curves lie on top of each other
+
+    Notes
+    -----
```
SimonHeybrock, line 307: 👍 

Maybe place under the parameter section, since users will often want to read the params, but not the notes?

src/ess/sans/beam_center_finder.py
nvaytet: I did, I am not sure it matters?

SimonHeybrock: Does it not affect convergence speed, maybe?

nvaytet: It doesn't seem to, the number of iterations is the same as before.

SimonHeybrock (approved):

src/ess/sans/beam_center_finder.py
```diff
@@ -187,18 +187,38 @@ def cost(xy: List[float], *args) -> float:
...
     """  # noqa: E501
     iofq = iofq_in_quadrants(xy, *args)
-    all_q = sc.concat(list(iofq.values()), dim='quadrant')
-    ref = sc.values(all_q.mean('quadrant'))
-    c = sc.abs(all_q - ref) / ref
-    out = c.sum().value
+    all_q = sc.concat([sc.values(da) for da in iofq.values()], dim='quadrant')
+    ref = all_q.mean('quadrant')
+    c = (all_q - ref)**2
+    out = (sc.sum(ref * c) / sc.sum(ref)).value
```
SimonHeybrock, line 210: You mention above that you now use a weighted mean, but I cannot really see this here. I expected that the variances would be used as weights?

nvaytet: I thought you meant that the signal intensity should be used as weights, to give less importance to the noisy parts with low signal?

nvaytet: I'm slightly reluctant to use variances as weights, as we are not computing them accurately?

SimonHeybrock (approved):

PR #170, opened 2023-04-11
Drop normalization variances in sans and reflectometry workflows

### LINKED ISSUE #162: Update SANS workflow to discard normalization variances depending on threshold

nvaytet:
Fixes #162

### DIFF
--- a/docs/techniques/wfm/reducing-wfm-data.ipynb
+++ b/docs/techniques/wfm/reducing-wfm-data.ipynb
@@ -975,8 +975,12 @@
    },
    "outputs": [],
    "source": [
-    "# Normalizing binned data is done using the sc.lookup helper\n",
+    "# In a normalization operation, the denominator must be histogrammed (it cannot be events)\n",
+    "# Note: we drop the variances with `sc.values` because the event normalization introduces correlations\n",
+    "# which are not handled by Scipp's uncertainty propagation. This is ok here as the size of the\n",
+    "# standard deviations is not the focus of this notebook.\n",
     "hist = sc.values(converted_event[\"vanadium\"].hist(wavelength=converted[\"sample\"].coords['wavelength']))\n",
+    "# Normalizing binned data is done using the sc.lookup helper\n",
     "normalized_event = converted_event[\"sample\"].bins / sc.lookup(func=hist, dim='wavelength')"
    ]
   },
--- a/docs/techniques/wfm/reducing-wfm-data.ipynb
+++ b/docs/techniques/wfm/reducing-wfm-data.ipynb
@@ -19,7 +19,9 @@
    "cell_type": "code",
    "execution_count": null,
    "id": "f5716e61",
-   "metadata": {},
+   "metadata": {
+    "tags": []
+   },
    "outputs": [],
    "source": [
     "import numpy as np\n",
@@ -56,7 +58,9 @@
    "cell_type": "code",
    "execution_count": null,
    "id": "1a7055c5",
-   "metadata": {},
+   "metadata": {
+    "tags": []
+   },
    "outputs": [],
    "source": [
     "coords = wfm.make_fake_beamline(nframes=6)\n",
@@ -79,7 +83,9 @@
    "cell_type": "code",
    "execution_count": null,
    "id": "51ad4053",
-   "metadata": {},
+   "metadata": {
+    "tags": []
+   },
    "outputs": [],
    "source": [
     "x = np.linspace(1, 10.0, 100000)\n",
@@ -111,7 +117,9 @@
    "cell_type": "code",
    "execution_count": null,
    "id": "f326c6c9",
-   "metadata": {},
+   "metadata": {
+    "tags": []
+   },
    "outputs": [],
    "source": [
     "nevents = 1_000_000\n",
@@ -151,7 +159,9 @@
    "cell_type": "code",
    "execution_count": null,
    "id": "b0e727e8",
-   "metadata": {},
+   "metadata": {
+    "tags": []
+   },
    "outputs": [],
    "source": [
     "# Histogram and plot the event data\n",
@@ -176,7 +186,9 @@
    "cell_type": "code",
    "execution_count": null,
    "id": "a2e6fce2",
-   "metadata": {},
+   "metadata": {
+    "tags": []
+   },
    "outputs": [],
    "source": [
     "for key, item in events.items():\n",
@@ -197,7 +209,9 @@
    "cell_type": "code",
    "execution_count": null,
    "id": "1bca7897",
-   "metadata": {},
+   "metadata": {
+    "tags": []
+   },
    "outputs": [],
    "source": [
     "# The ratio of neutron mass to the Planck constant\n",
@@ -223,7 +237,9 @@
    "cell_type": "code",
    "execution_count": null,
    "id": "35e2d1d9",
-   "metadata": {},
+   "metadata": {
+    "tags": []
+   },
    "outputs": [],
    "source": [
     "ds = sc.Dataset(coords=coords)\n",
@@ -243,7 +259,9 @@
    "cell_type": "code",
    "execution_count": null,
    "id": "ec0ac640",
-   "metadata": {},
+   "metadata": {
+    "tags": []
+   },
    "outputs": [],
    "source": [
     "f = wfm.plot.time_distance_diagram(ds)"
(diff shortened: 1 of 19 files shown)


src/ess/sans/i_of_q.py
```diff
@@ -73,7 +71,13 @@ def preprocess_monitor_data(
...
             monitor = monitor.rebin(wavelength=wavelength_bins)
     if background is not None:
-        monitor -= background
+        # TODO: reference Heybrock et al. (2023) paper
+        # For subtracting the background from the monitors, we need to remove the
+        # variances because the broadcasting operation will fail.
+        # We then need to have a check, once the sample data has been converted to
+        # wavelength that this approximation is valid, by comparing the sample data
+        # counts to the monitor counts. See :func:`to_I_of_Q`.
+        monitor = monitor - sc.values(background)
```
SimonHeybrock: Where is that check?

src/ess/sans/i_of_q.py
```diff
@@ -349,6 +357,17 @@ def to_I_of_Q(data: sc.DataArray,
     for coord in ['position', 'sample_position', 'source_position']:
         denominator.coords[coord] = data.meta[coord]
 
+    # Verify that the ratio of sample detector counts to monitor counts is small, so
+    # we can safely drop the variances of the monitor to avoid broadcasting issues.
+    # See Heybrock et al. (2023).
+    alpha = alpha_ratio(data.hist(wavelength=wavelength_bins),
+                        data_monitors['transmission'].rebin(wavelength=wavelength_bins))
```
SimonHeybrock: Shouldn't we also check the other factors in the denominator, such as the direct beam?

src/ess/sans/i_of_q.py
```diff
@@ -349,6 +357,17 @@ def to_I_of_Q(data: sc.DataArray,
...
+    if alpha > alpha_threshold:
+        raise ValueError(
+            f'alpha = {alpha} > {alpha_threshold}! This means that the ratio of '
+            'detector counts to monitor counts is too high, and the variances of the '
+            'monitor data cannot be safely dropped.')
```
SimonHeybrock: Can we add logging of warnings if we are close to the threshold?

src/ess/sans/normalization.py
```diff
@@ -103,8 +103,11 @@ def compute_denominator(direct_beam: sc.DataArray, data_incident_monitor: sc.Dat
     :
         The denominator for the SANS I(Q) normalization.
     """
-    denominator = (solid_angle * direct_beam * data_incident_monitor *
-                   transmission_fraction)
+    # TODO: reference Heybrock et al. (2023) paper
+    # We need to remove the variances because the broadcasting operation between
+    # solid_angle (pixel-dependent) and monitors (wavelength-dependent) will fail.
+    denominator = sc.values(solid_angle) * sc.values(
+        direct_beam * data_incident_monitor * transmission_fraction)
```
SimonHeybrock: Is this relying on being called via the code in `i_of_q.py`? It seems dangerous to drop the variances here, but not check alpha here. Sooner or later someone will use this code but not the one where the check happened.

src/ess/uncertainty.py
```diff
@@ -0,0 +1,74 @@
...
+from typing import TypeVar, Union
+
+import scipp as sc
+
+from .logging import get_logger
+
+T = TypeVar("T", bound=Union[sc.Variable, sc.DataArray])
+
+
+def drop_variances(da: T, *, reference: T, name: str = '', threshold: float = 0.1) -> T:
```
SimonHeybrock: Unused?

src/ess/sans/i_of_q.py
nvaytet: I added a check comparing the background counts to the total monitor counts. I don't know if that is appropriate. I think it's difficult to estimate using the equations in the paper because the background is a sum over some random range of the input, and getting a correlation for that is not obvious.

Do we need a more detailed analysis?

(Also the text about checking once the sample data has been converted to wavelength was nonsense, so I removed it)

src/ess/sans/normalization.py
nvaytet: I moved the alpha check in here :+1:

src/ess/uncertainty.py
nvaytet: Yes, i'll remove.

docs/techniques/wfm/reducing-wfm-data.ipynb
```diff
@@ -900,11 +970,13 @@
...
    "id": "8b98d524",
-   "metadata": {},
+   "metadata": {
+    "tags": []
+   },
    "outputs": [],
    "source": [
     "# Normalizing binned data is done using the sc.lookup helper\n",
-    "hist = converted_event[\"vanadium\"].hist(wavelength=converted[\"sample\"].coords['wavelength'])\n",
+    "hist = sc.values(converted_event[\"vanadium\"].hist(wavelength=converted[\"sample\"].coords['wavelength']))\n",
```
SimonHeybrock, line 982: Dropping variances without comment?

src/ess/reflectometry/conversions.py
```diff
@@ -136,9 +136,9 @@ def tof_to_wavelength(data_array: sc.DataArray,
             unit = 'angstrom'
         data_array_wav.attrs[
             'orso'].value.data_source.measurement.instrument_settings.wavelength = (
-                fileio.base.ValueRange(data_array_wav.coords['wavelength'].min().value,
-                                       data_array_wav.coords['wavelength'].max().value,
-                                       unit))
+                fileio.base.ValueRange(
+                    float(data_array_wav.coords['wavelength'].min().value),
+                    float(data_array_wav.coords['wavelength'].max().value), unit))
```
SimonHeybrock, line 141: I think the `.value` may be unnecessary, iirc Jan-Lukas added support that conversion to float some time last year.

src/ess/sans/i_of_q.py
```diff
@@ -73,7 +70,18 @@ def preprocess_monitor_data(
         else:
             monitor = monitor.rebin(wavelength=wavelength_bins)
     if background is not None:
-        monitor -= background
+        # TODO: reference Heybrock et al. (2023) paper
+        # For subtracting the background from the monitors, we need to remove the
+        # variances because the broadcasting operation will fail.
+        # We add a simple check comparing the background level to the total number
+        # of counts. TODO: is this check good enough?
```
SimonHeybrock: Are there open issues for these TODOs?

src/ess/sans/i_of_q.py
```diff
@@ -316,6 +270,9 @@ def to_I_of_Q(data: sc.DataArray,
         If defined, return the data as a set of bands in the wavelength dimension. This
         is useful for separating different wavelength ranges that contribute to
         different regions in Q space.
+    alpha_threshold:
```
SimonHeybrock: I would give a more explicit name. `alpha` was used in the paper, but is meaningless in any other context.

src/ess/uncertainty.py
```diff
@@ -0,0 +1,31 @@
...
+"""Tools for handling statistical uncertainties."""
+
+from typing import TypeVar, Union
+
+import scipp as sc
+
+T = TypeVar("T", bound=Union[sc.Variable, sc.DataArray])
+
+
+def alpha_ratio(numerator: T, denominator: T) -> float:
```
SimonHeybrock: Naming... see above

docs/techniques/wfm/reducing-wfm-data.ipynb
nvaytet: Added a comment.

src/ess/reflectometry/conversions.py
nvaytet: Doesn't this work only if the unit is `dimensionless` or `None`?

src/ess/sans/i_of_q.py
nvaytet: I'll make some.

src/ess/reflectometry/conversions.py
SimonHeybrock: Yes, you are probably right.

SimonHeybrock (approved):

Issue #176, opened 2023-04-18
Masking tools


SimonHeybrock:
This is required by DREAM and LOKI (and probably more):

In a notebook:

- Create plots that visualize counts for various logical/physical detector components, e.g., total counts in wires, modules, ...
- Provide Python function to detector broken components (such as wires), e.g., if they have no counts, or counts below a threshold.
- Plot the masks in the diagnostics plot. This lets the user verify the masks.
- All for manual (Python based) manipulation of the masks, e.g., be listing additional wires to mask by hand.
- Further plots may depend on the instrument, e.g., some may want to show the instrument view with masking in place.
- ~Masks will need to be saved to a file. Format to be decided.~ done, see #180.

Notebooks will likely be instrument-specific, since we need, e.g., different plots for different detector banks, and dimensionality depends on the instrument as well.

Potential example for DREAM:

![image](https://user-images.githubusercontent.com/12912489/232732148-0fc78178-5bef-443c-8e43-ca4536c7c901.png)

See also https://confluence.esss.lu.se/display/DAM/DREAM+-+Save+mask+to+file

SimonHeybrock:
@nvaytet Is this done / replaced? If not, reopen please.

nvaytet:
A tool for masking was added in https://github.com/scipp/scippneutron/pull/521.
General re-shaping and slicing of detectors will just be done with Scipp, as detectors are to be loaded with logical dimensions.
So yes, I would say it's done.

PR #180, opened 2023-05-01
Save detector masks

SimonHeybrock:
Related to https://confluence.esss.lu.se/display/DAM/DREAM+-+Save+mask+to+file. #176 will later use this to save masks created by users.

Note that this depends on ScippNexus features that have not been released yet, expect tests to fail on CI for now.

### DIFF
--- a/src/ess/masking.py
+++ b/src/ess/masking.py
@@ -1,7 +1,8 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
 from pathlib import Path
-from typing import Union, Mapping, Optional
+from typing import Mapping, Optional, Union
+
 import h5py
 import scipp as sc
 
--- a/src/ess/masking.py
+++ b/src/ess/masking.py
@@ -1,14 +1,16 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
 from pathlib import Path
-from typing import Mapping, Union
-
+from typing import Union, Mapping, Optional
 import h5py
 import scipp as sc
 
 
 def save_detector_masks(
-    filename: Union[str, Path], detectors: Mapping[str, sc.DataArray]
+    filename: Union[str, Path],
+    detectors: Mapping[str, sc.DataArray],
+    *,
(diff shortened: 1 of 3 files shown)


src/ess/masking.py
```diff
@@ -0,0 +1,54 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
+from pathlib import Path
+from typing import Mapping, Optional, Union
+
+import h5py
+import scipp as sc
+
+
+def save_detector_masks(
```
nvaytet, line 10: So is saving masks on detector pixels the main application for this?
I guess you usually don't have a semi-permanent and re-usable mask on a time log because that is most often very run-dependent.
Could you have a wavelength mask that is always the same on some intrument? (e.g. masking some Bragg peaks from the guide or sample environment?)
I'm just asking if we need a different function name, or additional functions to mask other things than detector pixels.

SimonHeybrock: This function should work fine for saving tof (or wavelength) masks of detectors. One thing that people may want to do is saving *monitor* masks, which would work in the same way (but set `NX_class='NXmonitor'`). If we had `attrs` we could even do that automatically. Until then we will need to make a separate function.

src/ess/masking.py
```diff
@@ -0,0 +1,85 @@
...
+      >>>     masks = f['entry'][()]
+    """
+    with h5py.File(filename, "w-") as f:
+        entry = f.require_group("entry")
+        entry.attrs["NX_class"] = "NXentry"
+        if entry_metadata:
+            for key, value in entry_metadata.items():
+                entry[key] = value
+        instrument = entry.require_group("instrument")
+        instrument.attrs["NX_class"] = "NXinstrument"
```
nvaytet, line 70: Can we use `scippnexus` to do some of the work for us instead of using `h5py` directly?

SimonHeybrock: Not really, no. In such simple cases it would actually be more code.

nvaytet (approved):
Is it still true that the required `scippnexus` features are not released yet? CI appears to be passing now.

SimonHeybrock:
> Is it still true that the required `scippnexus` features are not released yet? CI appears to be passing now.

It has been released last week.

PR #181, opened 2023-05-16
Use ScippNexus for Amor files

jl-wynen:
As discussed it is unclear how we want to propagate metadata and store data in a workflow. This will likely need support by a future workflow framework. So for now, this PR only replaces `load_nexus` with using ScippNexus directly and puts the data into (almost) the same structure as before. At least, this will avoid the upcoming deprecation warning.

### DIFF
--- a/src/ess/amor/load.py
+++ b/src/ess/amor/load.py
@@ -5,7 +5,7 @@ from pathlib import Path
 from typing import Any, Optional, Union
 
 import scipp as sc
-import scippnexus as snx
+import scippnexus.v2 as snx
 
 from ..logging import get_logger
 from .beamline import make_beamline
--- a/src/ess/amor/load.py
+++ b/src/ess/amor/load.py
@@ -64,7 +64,6 @@ def _assemble_event_data(dg: sc.DataGroup) -> sc.DataArray:
     """
     events = dg['instrument']['multiblade_detector']
     events.bins.coords['tof'] = events.bins.coords.pop('event_time_offset')
-    del events.bins.coords['event_time_zero']
     events.coords['position'] = sc.spatial.as_vectors(
         events.coords.pop('x_pixel_offset'),
         events.coords.pop('y_pixel_offset'),
--- a/src/ess/amor/load.py
+++ b/src/ess/amor/load.py
@@ -50,6 +50,18 @@ def _tof_correction(data: sc.DataArray, dim: str = 'tof') -> sc.DataArray:
 
 
 def _assemble_event_data(dg: sc.DataGroup) -> sc.DataArray:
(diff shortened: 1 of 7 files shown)


arm61 (approved):
Looks good to me (massive pinch of salt).

src/ess/amor/load.py
```diff
@@ -31,15 +31,16 @@ def _tof_correction(data: sc.DataArray, dim: str = 'tof') -> sc.DataArray:
     """
     if 'orso' in data.attrs:
         data.attrs['orso'].value.reduction.corrections += ['chopper ToF correction']
+    tof_unit = data.bins.constituents['data'].coords[dim].unit
```
SimonHeybrock:
```suggestion
    tof_unit = data.bins.coords[dim].bins.unit
```
avoid the `constituents` access.

src/ess/amor/load.py
```diff
@@ -92,27 +120,24 @@ def load(
         ].data.values
 
     # Convert tof nanoseconds to microseconds for convenience
-    # TODO: is it safe to assume that the dtype of the binned wrapper coordinate is
-    # the same as the dtype of the underlying event coordinate?
-    data.bins.coords['tof'] = data.bins.coords['tof'].astype('float64', copy=False)
-    data.coords['tof'] = data.coords['tof'].astype('float64', copy=False)
-    data.bins.coords['tof'] = sc.to_unit(data.bins.coords['tof'], 'us', copy=False)
-    data.coords['tof'] = sc.to_unit(data.coords['tof'], 'us', copy=False)
+    data.bins.coords['tof'] = data.bins.coords['tof'].to(
```
nvaytet, line 122: Is also converting `data.coords['tof']` no longer needed? (Was it ever needed?)

jl-wynen: There no longer is a tof bin coord. This was a quirk of `load_nexus` which added a single tof bin.

src/ess/amor/load.py
```diff
@@ -48,14 +49,47 @@ def _tof_correction(data: sc.DataArray, dim: str = 'tof') -> sc.DataArray:
...
+        A data group with the structure of an Amor NeXus file.
+
+    Returns
+    -------
+    :
+        A data array with the events extracted from ``dg``.
+    """
+    events = dg['instrument']['multiblade_detector']
+    events.bins.coords['tof'] = events.bins.coords.pop('event_time_offset')
+    del events.bins.coords['event_time_zero']
```
SimonHeybrock: Will be required for pulse-skipping modes and event filtering, why remove? To save space?

SimonHeybrock (approved):

PR #184, opened 2023-06-12
Dream instrument view

nvaytet:
~- Add some helper functions to manipulate DREAM data in logical dimensions~
~- Add a `dream.instrument_view` that can have a `tof` slider.~
~- Add notebook with dream diagnostics plots and instrument views~

This adds a `dream.instrument_view` that can have a `tof` slider.
It also has some checkboxes to hide/show the separate detector modules.
Screenshot at 2023-11-17 09-08-30

Also reduce duplicate code in data registries.


src/ess/dream/diagnostics.py
```diff
@@ -0,0 +1,127 @@
...
+def to_logical_dims(da: sc.DataArray) -> sc.DataArray:
+    """
+    Reshape the input data to logical dimensions.
+
+    Parameters
+    ----------
+    da:
+        DataArray to reshape.
+    """
+    return da.group('module', 'segment', 'counter', 'wire', 'strip')
```
SimonHeybrock: Are those specific to the mantle? The endcaps and high-res detectors may be different?

src/ess/dream/diagnostics.py
```diff
@@ -0,0 +1,127 @@
...
+        DataArray to group.
+    dims:
+        Coordinates to use for grouping.
+    """
+    in_dims = set(da.dims)
+    out_dims = set(dims)
+    if (not dims) or (in_dims == out_dims):
+        return da
+    grouped = da.group(*list(out_dims - in_dims))
+    return grouped.bins.concat(list(set(grouped.dims) - out_dims))
```
SimonHeybrock: `set` is unordered, I think this may be a problem here?

src/ess/dream/diagnostics.py
```diff
@@ -0,0 +1,127 @@
...
+    da:
+        DataArray to group.
+    dims:
+        Coordinates to use for grouping.
+    """
+    in_dims = set(da.dims)
+    out_dims = set(dims)
+    if (not dims) or (in_dims == out_dims):
+        return da
+    grouped = da.group(*list(out_dims - in_dims))
```
SimonHeybrock: This won't work, I think: If you `group` with a dim-name then you will only get those groups that exist. So if you have, e.g., a pixel without any neutrons it will not be present in the output.

src/ess/dream/diagnostics.py
nvaytet: It seems all of them have the same components? (none are empty in the datagroup)
Screenshot at 2023-06-13 12-37-05

SimonHeybrock: 👍 (even though the NeXus file I have does not have the `sumo*` groups, but uses other names)

src/ess/dream/diagnostics.py
nvaytet: I guess this is works for the Geant4 csv files where you are getting the events, which don't list the zeros anyway.
But as we have often said, the zeros are important.
I'll ask Celine for that Nexus file so I can see how it is stored there and get the data into a form that is equivalent to the Nexus file.

SimonHeybrock: I don't think it should work for the CSV files either: If there is no event you will not get a group. Try loading a shorter section of the file, I think it will fail here (when reshaping).

src/ess/dream/instrument_view.py
```diff
@@ -0,0 +1,139 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
+
+from typing import List, Optional, Union
+
+import plopp as pp
+import scipp as sc
+
+DREAM_DETECTOR_DIMENSIONS = ('module', 'segment', 'counter', 'wire', 'strip')
```
SimonHeybrock: We should define this in a common place, also used by loaders, see my as well as @jl-wynen's PRs.

src/ess/dream/instrument_view.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+def instrument_view(
+    data: Union[sc.DataArray, sc.DataGroup],
+    x: Optional[str] = None,
+    y: Optional[str] = None,
+    z: Optional[str] = None,
+    pos: Optional[str] = None,
+    dim: Optional[str] = None,
+    pixel_size: Union[sc.Variable, float] = DREAM_PIXEL_SIZE,
```
SimonHeybrock: Note that we are violating the mutable-default-arg rule. We have the same problem in other places. Not sure if we should outright avoid it, or allow it by having the readonly flag set on a variable?

src/ess/dream/instrument_view.py
```diff
@@ -0,0 +1,139 @@
...
+    dims = [d for d in data.dims if (d in DREAM_DETECTOR_DIMENSIONS) and (d != dim)]
+    to = 'pixel'
+    if not isinstance(data, sc.DataArray):
+        data = sc.concat(
+            [
+                _preprocess_data(data=da, to_be_flattened=dims, dim=dim, to=to)
+                for da in data.values()
+            ],
+            dim=to,
+        )
```
SimonHeybrock: Didn't pythreejs support multiple point clouds? Could different detector banks be added separately, potentially with a check box to toggle them on and off?

src/ess/dream/instrument_view.py
```diff
@@ -0,0 +1,139 @@
...
+        data = _preprocess_data(data=data, to_be_flattened=dims, dim=dim, to=to)
+
+    if pos is not None:
+        if any((x, y, z)):
+            raise ValueError(
+                f'If pos ({pos}) is defined, all of '
+                f'x ({x}), y ({y}), and z ({z}) must be None.'
+            )
+        coords = {
+            (x := f'{pos}.x'): data.coords[pos].fields.x,
```
SimonHeybrock: Why walrus? You are not using `x`?

src/ess/dream/instrument_view.py
nvaytet: Yes, you're right. Having toggle boxes may be a nice addition.
To be honest, I am not so happy about how the implementation is shaping here. I think I'd rather first work on https://github.com/scipp/plopp/issues/247 and then it should be easier to do this one.

docs/instruments/dream/dream-instrument-view.ipynb
celinedurniak, line 1: The instrument view works correctly. But instead of displaying the SUMOs, I think that for consistency, the 4 tick boxes for the SUMOs should be replaced by 2 tick boxes for the endcap backward and forward detectors.
Note that for the endcap detectors, SUMOs are numbered starting from 3.

celinedurniak (approved):

docs/instruments/dream/dream-instrument-view.ipynb
```diff
@@ -0,0 +1,237 @@
...
+    "    {\n",
+    "        key: da.group('module', 'segment', 'counter', 'wire', 'strip')\n",
+    "        for key, da in dg.items()\n",
+    "    }\n",
+    ")\n",
+    "for da in dg.values():\n",
+    "    da.coords['x'] = da.bins.coords.pop('x_pos').bins.mean()\n",
+    "    da.coords['y'] = da.bins.coords.pop('y_pos').bins.mean()\n",
+    "    da.coords['z'] = da.bins.coords.pop('z_pos').bins.mean()\n",
+    "dg"
```
SimonHeybrock: At least some of this should now be done using `dream.load_geant4_csv` added in #203, I think?

docs/instruments/dream/dream-instrument-view.ipynb
```diff
@@ -0,0 +1,237 @@
...
+   "metadata": {
+    "editable": true,
+    "slideshow": {
+     "slide_type": ""
+    },
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "full_view = dream.instrument_view(data, dim='tof')\n",
```
SimonHeybrock, line 139: Is the intention here that one could pass `dim='wavelength'` after transforming the coords? Might be worth mentioning here, since it may not be obvious to readers.

src/ess/amor/data.py
```diff
@@ -1,33 +1,16 @@
...
 
-    return pooch.create(
-        path=pooch.os_cache('ess/amor'),
-        env='ESS_AMOR_DATA_DIR',
-        base_url='https://public.esss.dk/groups/scipp/ess/amor/{version}/',
-        version=_version,
-        registry={
-            "reference.nxs": "md5:56d493c8051e1c5c86fb7a95f8ec643b",
-            "sample.nxs": "md5:4e07ccc87b5c6549e190bc372c298e83",
-        },
```
SimonHeybrock, line 19: This code has been moved already, the refactor will get lost (but that is ok, I guess).

src/ess/dream/instrument_view.py
```diff
@@ -0,0 +1,135 @@
...
+        data = sc.DataGroup({data.name or 'data': data})
+    elif isinstance(data, dict):
+        data = sc.DataGroup(data)
+    return data
+
+
+@pp.node
+def slice_range(da, trunc_range):
+    min_tr, max_tr = trunc_range
+    return da['tof', min_tr:max_tr].sum('tof')
```
SimonHeybrock: This seems to assume that data is binned (or histogrammed) with a potentially fine resolution? I believe DREAM has some high-resolution settings, where we may need thousands of TOF bins. Combined with the large number of pixels this may use too much memory. We can work around this using something event filtering.

@celinedurniak How many (how small) TOF (or wavelength) bins will you need to visualize?

src/ess/dream/instrument_view.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+@pp.node
+def pre_process(da, dim):
+    dims = list(da.dims)
+    if dim is not None:
+        dims.remove(dim)
+    out = da.flatten(dims=dims, to='pixel')
+    sel = sc.isfinite(out.coords['x'])
```
SimonHeybrock: What is `'x'`? The literal X-coord? Why do we need to select based on it?

src/ess/dream/instrument_view.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+def instrument_view(data, dim=None, pixel_size=None, **kwargs):
+    """
+    Three-dimensional visualization of the DREAM instrument.
+
+    Parameters
+    ----------
+    data:
+        Data to visualize.
```
SimonHeybrock: Misses some info about the expected structure (datagroup, subgroups, dims, coords) of the input.

src/ess/dream/instrument_view.py
```diff
@@ -0,0 +1,135 @@
...
+            key = self.artist_mapping[name]
+            val = ch.value
+            self.fig.artists[key].points.visible = val
+            for c in "xyz":
+                cut_nodes = getattr(self.cutting_tool, f'cut_{c}').select_nodes
+                if key in cut_nodes:
+                    self.fig.artists[cut_nodes[key].id].points.visible = val
+
+
+def instrument_view(data, dim=None, pixel_size=None, **kwargs):
```
SimonHeybrock: Type hints?

docs/instruments/dream/dream-instrument-view.ipynb
nvaytet: Right now, I am just using the keys in the data group that is supplied as input, to keep it flexible.
If you want to replace the sumos with end-caps, it's probably at the data loading that we should do that.

nvaytet: Actually, using `dream.load_geant4_csv` should make this go away.

src/ess/dream/instrument_view.py
nvaytet: Actually, the `slice_range` function is no longer used. In a previous iteration, I used to have a range slider which allowed you to change the tof range to sum over, like the Mantid instrument view.

However, I now switched to just slicing single bins, because it is much faster.

That said, I think the issue you raise is still valid?

src/ess/dream/instrument_view.py
nvaytet: This is because of the way we compute the coordinates for the pixels. We do a `mean` on the positions inside the bins, but some of the bins are empty (no events reached that pixel).
So some positions end up being NaN, and the 3D view complains about nans (nothing is rendered).

So I need some way to filter out NaNs. I agree that maybe the instrument view shouldn't silently drop NaNs, and maybe this should be done explicitly in the notebook?

Filtering on NaNs here was easy because I was internally flattening the data. Using advanced indexing only works using a 1D mask. But I think it makes sense to keep the data with logical dimensions in the notebook.
Which is why I wanted to filter here...

I don't know what the best solution is.

nvaytet: I thought about filling the empty pixels with a single event with zero weight, which would then appear as a dark pixel in the instrument view, but I then do not know what the positions of those pixels should be, because they are absent from the original table.

src/ess/dream/instrument_view.py
SimonHeybrock: Yes, it is still a problem.

src/ess/dream/instrument_view.py
SimonHeybrock: Makes sense.

From NeXus files we will (according to the current plan) get a `position` coord instead. Will you handle this as well?

nvaytet: I think that if I use the loader Jan-Lukas wrote, I get a position coord, so I will probably just filter on that.

docs/instruments/dream/dream-instrument-view.ipynb
nvaytet: Most of this is now gone after switching to `load_geant4`.

docs/instruments/dream/dream-instrument-view.ipynb
nvaytet: Added a sentence.

SimonHeybrock (approved):

PR #193, opened 2023-08-08
Update mamba actions in CI

nvaytet:


jl-wynen (approved):
:+1:

Can you do this in Scipp and ScippNeutron, too? It would be great if we could handle this via copier...

nvaytet:
> It would be great if we could handle this via copier...

Yes, I guess this is what copier is for. Would be good to set this up in all repos soon.

PR #194, opened 2023-08-11
Offspec data reduction

arm61:
Here is the OFFSPEC data reduction example.


arm61:
That final fix might break the Amor notebook (to fix it is only necessary to have a `dimension='dectector_id'` in the `save_ort` call).

docs/instruments/external/offspec/data.py
```diff
@@ -0,0 +1,90 @@
...
+            "reduced_mantid.xye": "md5:1f372f51d2cefb8dee302cf0093b684f"
+        })
+
+
+_pooch = _make_pooch()
+
+
+def get_path(name: str) -> str:
+    """
+    Return the path to a data file bundled with scippneutron.
```
jl-wynen: Copy mistake?

docs/instruments/external/offspec/data.py
```diff
@@ -0,0 +1,90 @@
...
+        description="Raw OFFSPEC data from the direct beam.",
+        type="raw",
+
+        owner_group="ess",
+        access_groups=["offspec"],
+
+        owner="Joshanial F. K. Cooper",
+        principal_investigator="Joshanial F. K. Cooper",
+        contact_email="jos.cooper@stfc.ac.uk",
+        end_time=datetime.now(),
```
jl-wynen: Can you use a more realistic value? There should be something in the file.

docs/instruments/external/offspec/data.py
```diff
@@ -0,0 +1,90 @@
...
+
+        data_format="ISIS NeXus file",
+        creation_location = "ISIS Neutron and Muon Source",
```
jl-wynen: According to the docs, this field should be site/facility/beamline. But I don't think Max has a clear spec for ESS. But I'd still change it to
```suggestion
        creation_location = "RAL/ISIS/OFFSPEC",
```

docs/instruments/external/offspec/data.py
```diff
@@ -0,0 +1,90 @@
...
+        instrument_id = "OFFSPEC"
```
jl-wynen: Ok. This should be the id in the catalogue. But since there is no catalogue...

docs/instruments/external/offspec/data.py
```diff
@@ -0,0 +1,90 @@
...
+    )
+    file = Path(get_path('direct_beam.nxs'))
+    direct_beam_dataset.add_local_files(str(file), base_path=str(file.parents[0]))
```
jl-wynen: Why use `str` here?

docs/instruments/external/offspec/data.py
```diff
@@ -0,0 +1,90 @@
...
+        access_groups=["offspec"],
+
+        owner="Joshanial F. K. Cooper",
+        principal_investigator="Joshanial F. K. Cooper",
+        contact_email="jos.cooper@stfc.ac.uk",
+        end_time=datetime.now(),
+
+        data_format="ISIS NeXus file",
+        creation_location = "ISIS Neutron and Muon Source",
+        instrument_id = "OFFSPEC"
```
jl-wynen: See above

src/ess/reflectometry/io.py
```diff
@@ -5,7 +5,7 @@
 import scipp as sc
 
 
-def save_ort(data_array: sc.DataArray, filename: str, dimension: str = None):
+def save_ort(data_array: sc.DataArray, filename: str, dimension: str):
```
jl-wynen:
```suggestion
def save_ort(data_array: sc.DataArray, filename: str, dimension: Optional[str]):
```
And explain that it doesn't compute an average if `dimension is None`.

docs/instruments/external/offspec/offspec_mantid.ipynb
```diff
@@ -0,0 +1,787 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "e4b8eee9-2fa7-4493-8c85-8b166d5bfcec",
+   "metadata": {},
+   "source": [
+    "# Set up the SciCat database locally\n",
+    "\n",
+    "First, we will set up a fake SciCat database and add our raw data to this. \n",
+    "This is achieved with the `scitacean` package, which is also compatible with the real ESS SciCat instance. \n",
+    "The direct beam and sample (which is data from a system of quartz and copper at the air-silicon interface) NeXus files are added to a dataset that is \"uploaded\" to the fake SciCat instance. "
```
jl-wynen: Can you add a link to https://scicatproject.github.io/scitacean/user-guide/downloading.html in order to explain how to set up a real client.

jl-wynen: And can you explain the helper functions and that this would not be used in a real situation?

docs/instruments/external/offspec/offspec_mantid.ipynb
```diff
@@ -0,0 +1,787 @@
...
+   "metadata": {},
+   "source": [
+    "## Populating the ORSO header\n",
+    "\n",
+    "We will write the reduced data file following the [ORSO `.ort` standard](https://www.reflectometry.org/file_format/specification), to enable a metadata rich header. \n",
+    "We will create an empty header and then populate this. \n",
+    "\n",
+    "### The data source information\n",
+    "\n",
+    "Many aspects of the ORSO header can be populate directly from the raw data file (in particular within the `data_source` section). "
```
jl-wynen:
```suggestion
    "Many aspects of the ORSO header can be populated directly from the raw data file (in particular within the `data_source` section). "
```

docs/instruments/external/offspec/offspec_mantid.ipynb
```diff
@@ -0,0 +1,787 @@
...
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "header.reduction.software = fileio.reduction.Software(name='ess', version=ess.__version__, platform=platform.platform())\n",
+    "header.reduction.timestamp = datetime.now()\n",
+    "header.reduction.creator = fileio.base.Person(name='Andrew R. McCluskey', \n",
+    "                                              affiliation='European Spallation Source', \n",
+    "                                              contact='andrew.mccluskey@ess.eu')\n",
```
jl-wynen: Don't you rather want to use a fake name? People copy code from documentation and forget to change it...

docs/instruments/external/offspec/offspec_mantid.ipynb
```diff
@@ -0,0 +1,787 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "16562ab9-0f20-47fd-b9fd-e5d3ca8019da",
+   "metadata": {
+    "tags": []
+   },
+   "source": [
+    "### Correcting the position of detector pixels\n",
+    "\n",
+    "The pixel positions in sample file must be modified to account for the transformation on the detector by rotating it around the sample. \n",
```
jl-wynen:
```suggestion
    "The pixel positions in the sample file must be modified to account for the transformation on the detector by rotating it around the sample. \n",
```

docs/instruments/external/offspec/offspec_mantid.ipynb
```diff
@@ -0,0 +1,787 @@
...
+   "source": [
+    "norm_q.plot(norm='log')"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "7636a6b0-6d9e-40e6-9169-a29fa18e4b21",
+   "metadata": {},
+   "source": [
+    "## Comparison with Mantid\n",
```
jl-wynen: I don't think this should be part of the docs. We want to show how to do things with Scipp and not with Mantid. That being said, the comparison is valuable.
You can leave it here for now. We'll hopefully move this to a better place.

docs/instruments/external/offspec/offspec_mantid.ipynb
```diff
@@ -0,0 +1,787 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "c1f70104-e237-4afa-bddd-3e0ade61fdd3",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "reduced_mantid_remote = client.get_dataset(reduced_mantid_up.pid)\n",
+    "reduced_mantid_local = client.download_files(reduced_mantid_remote, target=\".\")\n",
```
jl-wynen: Is this really necessary? For one, it pollutes the file system with the mantid file. And also, we can use the file directly and don't have to make a roundtrip through SciCat.

docs/instruments/external/offspec/offspec_mantid.ipynb
```diff
@@ -0,0 +1,787 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "reduced = sample_local.derive()\n",
+    "\n",
+    "reduced.add_local_files('./offspec.ort', base_path='.')\n",
+    "reduced.contact_email = 'andrew.mccluskey@ess.eu'\n",
+    "reduced.input_datasets.append(direct_beam_local.pid)\n",
+    "reduced.name = 'Reduced reflectometry data from QCS sample'\n",
+    "reduced.used_software = ['scipp-23.01.1']\n",
```
jl-wynen:
```suggestion
    "reduced.used_software = [f'scipp-{sc.__version__}']\n",
```

docs/instruments/external/offspec/data.py
arm61: It is the same as in the amor one, which I agree isn't correct for either file.

docs/instruments/external/offspec/offspec_mantid.ipynb
arm61: What part do you mean here? Specifically the downloading it?

docs/instruments/external/offspec/offspec_mantid.ipynb
arm61: The aim was to show that the reduction is as good as the mantid version.

src/ess/reflectometry/io.py
arm61: Does this function work with `dimension=None`?

jl-wynen: There is an explicit check here: https://github.com/scipp/ess/blob/cf91c69298abfac942e87a091648bfdc3c71bc2b/src/ess/reflectometry/io.py#L25 So I assumed it supports None.

docs/instruments/external/offspec/offspec_mantid.ipynb
jl-wynen: I don't see this as the purpose of the documentation. We want to show people how to use it. Investigations about how well it works are separate.

docs/instruments/external/offspec/offspec_mantid.ipynb
jl-wynen: > Specifically the downloading it?

Yes, in particular the file. But there is also no reason, IMHO, to download the dataset.

docs/instruments/external/offspec/data.py
```diff
@@ -48,13 +48,13 @@ def get_direct_beam(client: scitacean.Client):
         owner="Joshanial F. K. Cooper",
         principal_investigator="Joshanial F. K. Cooper",
         contact_email="jos.cooper@stfc.ac.uk",
-        end_time=datetime.now(),
+        end_time='2020-11-26T18:19:24',
```
jl-wynen: Mainly for future reference, scitacean parses this as UTC unless you specify a timezone. If this is the time at ISIS, then this is coincidentally (colonialism?) correct.

docs/instruments/external/offspec/offspec_mantid.ipynb
```diff
@@ -755,7 +754,7 @@
     "reduced.contact_email = 'andrew.mccluskey@ess.eu'\n",
     "reduced.input_datasets.append(direct_beam_local.pid)\n",
     "reduced.name = 'Reduced reflectometry data from QCS sample'\n",
-    "reduced.used_software = [f'scipp-{sc.__version__}']\n",
+    "reduced.used_software = ['scipp-23.01.1']\n",
```
jl-wynen: Why did you revert this again?

src/ess/reflectometry/io.py
arm61: Ah, I didn’t see that it was used again

jl-wynen:
I removed the SciCat stuff and comparison with Mantid. We will handle SciCat through Sciline when migrating the workflow. And the Mantid comparison is useful but shouldn't be part of the workflow example. We have a it in the history now and can recreate it later when we have a Sciline pipeline.

environment.yml
```diff
@@ -35,3 +35,4 @@ dependencies:
   - pip:
     - orsopy==1.0.1
     - sphinx-autodoc-typehints==1.18.3
+    - scitacean==23.7.0
```
nvaytet: remove?

docs/instruments/external/offspec/offspec_mantid.ipynb
```diff
@@ -0,0 +1,687 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "2f94ef8d-1dcb-4c68-9153-ae5a245cc86d",
+   "metadata": {},
+   "source": [
+    "## Loading the data\n",
+    "\n",
+    "In this example, we load some test data provided by the `offspec` package.\n",
+    "We need a sample measurement (the sample is `Air | Si(790 A) | Cu(300 A) | SiO2`) and a direct beam measurement.\n",
+    "The latter was obtained by positioning the detector directly in the beam of incident neutrons and moving the sample out of the way and gives an estimate for the ISIS pulse structure as a function of time-of-flight. "
```
nvaytet:
```suggestion
    "The latter was obtained by positioning the detector directly in the beam of incident neutrons and moving the sample out of the way.\n",
    "It gives an estimate for the ISIS pulse structure as a function of time-of-flight. "
```

docs/instruments/external/offspec/offspec_mantid.ipynb
```diff
@@ -0,0 +1,687 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "875e5233-95f8-4556-a564-e9fbece9eb42",
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "sample_full = scn.load_with_mantid(offspec.data.sample(), mantid_args={\"LoadMonitors\": True})\n",
```
nvaytet: We should move away from loading with Mantid (cf #208)

docs/instruments/external/offspec/offspec_mantid.ipynb
```diff
@@ -0,0 +1,687 @@
...
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "header = fileio.orso.Orso.empty()\n",
+    "\n",
+    "header.data_source.owner = fileio.base.Person(\n",
+    "    name=\"Joshanial F. K. Cooper\",\n",
+    "    affiliation=\"ISIS Neutron and Muon Source\",\n",
+    "    contact=\"jos.cooper@stfc.ac.uk\",\n",
```
nvaytet: Maybe we should have a fake name and email for our online docs?

docs/instruments/external/offspec/offspec_mantid.ipynb
```diff
@@ -0,0 +1,687 @@
...
+   "cell_type": "markdown",
+   "id": "16562ab9-0f20-47fd-b9fd-e5d3ca8019da",
+   "metadata": {
+    "tags": []
+   },
+   "source": [
+    "### Correcting the position of detector pixels\n",
+    "\n",
+    "The pixel positions in the sample data must be modified to account for the transformation on the detector by rotating it around the sample. \n",
+    "We can achieve this by understanding that the sample has been rotated by some amount (as discussed above) and that sample measurement has the specular peak at the same pixel as the direct beam measurement has the direct beam. \n",
```
nvaytet: > as discussed above

Not sure which discussion you are refering to?

docs/instruments/external/offspec/offspec_mantid.ipynb
```diff
@@ -0,0 +1,687 @@
+{
```
nvaytet: Maybe we should change the notebook name `offspec_mantid`?

docs/instruments/external/offspec/offspec_mantid.ipynb
```diff
@@ -0,0 +1,687 @@
...
+ "cells": [
```
nvaytet: Also, it is missing from the toc and does not show up in the built docs.

docs/instruments/external/offspec/offspec_mantid.ipynb
```diff
@@ -0,0 +1,687 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "3034804e-be92-4a62-b7e6-ce87bb1482a6",
+   "metadata": {},
+   "source": [
+    "A region of interest is then defined for the detector. \n",
+    "This is defined as twenty-five pixels around the specular peak or the direct beam. \n",
+    "The `scipp.DataArray` is concatenated along the `'spectrum'` coordinate at this stage, essentially collaposing all of the events onto a single pixel."
```
nvaytet:
```suggestion
    "The `scipp.DataArray` is concatenated along the `'spectrum'` coordinate at this stage, essentially collapsing all of the events onto a single pixel."
```

docs/instruments/external/offspec/offspec_mantid.ipynb
```diff
@@ -0,0 +1,687 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "c70eea4a-a996-4376-9cf7-79a12f6558ef",
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "sample_roi = sample['spectrum', 389:415].bins.concat('spectrum')\n",
+    "direct_beam_roi = direct_beam['spectrum', 389:415].bins.concat('spectrum')\n",
```
nvaytet: Should we be using spectrum numbers or use slicing in space with scalar variables?

docs/instruments/external/offspec/offspec_mantid.ipynb
```diff
@@ -0,0 +1,687 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "4390197d-b626-4398-a6e3-e1a7fff2482c",
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "sample_roi.coords['position'] = sample.coords['position'][401]\n",
```
nvaytet: Should we use some sort of weighted mean instead of a spectrum number? (Difficult to say when I have not seen  what the data looks like)

docs/instruments/external/offspec/offspec_mantid.ipynb
```diff
@@ -0,0 +1,687 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "73cf74fb-e2da-40b9-87df-66f1a36b719a",
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "sample_mon_wav -= sc.values(sample_mon_wav['wavelength', 15 * sc.units.angstrom].mean())\n",
```
nvaytet:
```suggestion
    "sample_mon_wav -= sc.values(sample_mon_wav['wavelength', 15 * sc.units.angstrom:].mean())\n",
```

docs/instruments/external/offspec/offspec_mantid.ipynb
```diff
@@ -0,0 +1,687 @@
...
+    "direct_beam_mon_wav -= sc.values(direct_beam_mon_wav['wavelength', 15 * sc.units.angstrom].mean())\n",
```
nvaytet:
```suggestion
    "direct_beam_mon_wav -= sc.values(direct_beam_mon_wav['wavelength', 15 * sc.units.angstrom:].mean())\n",
```

docs/instruments/external/offspec/offspec_mantid.ipynb
```diff
@@ -0,0 +1,687 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "011129a4-c67c-40c8-b651-52c5e36932a4",
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "sample_mon_wav_sum = sample_mon_wav['wavelength', 2 * sc.units.angstrom:15 * sc.units.angstrom].sum()\n",
```
nvaytet: Can we define some `wav_min` and `wav_max` to avoid repeating `15 * sc.units.angstrom` many times?

docs/instruments/external/offspec/offspec_mantid.ipynb
```diff
@@ -0,0 +1,687 @@
...
+    "direct_beam_mon_wav_sum = direct_beam_mon_wav['wavelength', 2 * sc.units.angstrom:15 * sc.units.angstrom].sum()\n",
+    "sample_norm = sample_wav / sc.values(sample_mon_wav_sum)\n",
```
nvaytet: Are we using `sc.values` because of the variance broadcast issue? If so, we should add a note here, so that we don't forget to update it to something better in the future (e.g. Simon's "upper bound" method)

docs/instruments/external/offspec/offspec_mantid.ipynb
```diff
@@ -0,0 +1,687 @@
...
+   "source": [
+    "sample_norm_hist = sample_norm.hist(\n",
+    "    wavelength=sc.geomspace(\n",
+    "        dim=\"wavelength\", start=2, stop=14, num=100, unit=sc.units.angstrom\n",
+    "    )\n",
+    ")\n",
+    "sample_norm_hist.coords.set_aligned('theta', False)\n",
+    "direct_beam_norm_hist = direct_beam_norm.hist(\n",
+    "    wavelength=sc.geomspace(\n",
+    "        dim=\"wavelength\", start=2, stop=14, num=100, unit=sc.units.angstrom\n",
```
nvaytet: Define edges once and store in a variable?

nvaytet: We could also consider putting sample and direct beam in a `DataGroup`?

docs/instruments/external/offspec/offspec_mantid.ipynb
jl-wynen: You are probably right for data protection reasons. But on the other hand, we need to somehow acknowledge the owner of the data. I'm not sure how to do both.

docs/instruments/external/offspec/offspec_mantid.ipynb
jl-wynen: Seems to have been removed. Do we need a discussion of is this comment (minus the ref to a discussion) enough?

docs/instruments/external/offspec/offspec_mantid.ipynb
jl-wynen: What should we change it to?

docs/instruments/external/offspec/offspec_mantid.ipynb
jl-wynen: Don't know. Probably the latter in the long run. This was chosen to reflect the exact choice used in Mantid.

docs/instruments/external/offspec/offspec_mantid.ipynb
jl-wynen: The specular peak is smaller than the detector resolution so there is little point to averaging in this case. If anything, we would probably have to do the opposite and count only part of the pixel. But I don't know the resolution of the data...

docs/instruments/external/offspec/offspec_mantid.ipynb
nvaytet: You can add a line in the notebooks (at the start or the end) saying something like "The data used in this notebook was kindly provided by Joshanial F. K. Cooper (ISIS, STFC)."?

docs/instruments/external/offspec/offspec_mantid.ipynb
nvaytet: Not sure, I would say removing "(as discussed above)" would be ok?

docs/instruments/external/offspec/offspec_mantid.ipynb
nvaytet: We are not using Mantid anywhere in the notebook, so something like `offspec_reduction`?

docs/instruments/external/offspec/offspec_mantid.ipynb
jl-wynen: Then we can keep the name here as well. Are you concerned about the email address? I can try removing it if ORSO lets me or use a dummy otherwise.

nvaytet: Yeah, it was mostly the email.

nvaytet (approved):

PR #199, opened 2023-08-28
Add `dream.load_nexus` for latest NeXus files

SimonHeybrock:
Not intended for merging, as this represents a loader for a file that is subject to change.

@celinedurniak Is this useful? I reverse-engineered what I think is the correct voxel order, and reordered voxels to something more sensible. I presume (hope?) that the order in the file will be fixed?

### DIFF
--- a/src/ess/dream/io.py
+++ b/src/ess/dream/io.py
@@ -1,0 +1,69 @@
+# Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
+import os
+from typing import Union, Dict, Any
+import scipp as sc
+import scippnexus as snx
+
+
+def load_nexus(path: Union[str, os.PathLike]):
+    definitions = snx.base_definitions()
+    definitions["NXdetector"] = FilteredDetector
+
+    with snx.File(path, definitions=definitions) as f:
+        dg = f['entry'][()]
+    if 'mantel' in dg['instrument']:
+        dg['instrument']['mantle'] = dg['instrument'].pop('mantel')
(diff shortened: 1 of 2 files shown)


src/ess/dream/io.py
```diff
@@ -0,0 +1,76 @@
...
+    """Load dream_mantle_test.nxs file."""
+    definitions = snx.base_definitions()
+    if not load_pixel_shape:
+        definitions["NXdetector"] = FilteredDetector
+
+    with snx.File(path, definitions=definitions) as f:
+        dg = f['entry'][()]
+    if 'mantel' in dg['instrument']:
+        dg['instrument']['mantle'] = dg['instrument'].pop('mantel')
+    # The file has (counter:2, hemisphere:2, module:5, segment:6, wire:32, strip:128)
```
celinedurniak, line 19: - What is "hemisphere"? 
- Comment: The number of strips, modules, wires depends on the detectors (endcap backward, forward, high resolution) and on the configuration (Full scope or Day 1)

SimonHeybrock: - Hemisphere is negative or positive Z. The file you shared seems to split wires into two parts, so there are only 128 strips per wire.
- Yes, this is processing only the mantle. I hope we can fix the order in the file such that the actual shape will be easier to guess (guessing is all we can do as it was decided that we cannot get this information from the NeXus file).

SimonHeybrock:
I will update this for the latest file, but move it into `essdiffraction`.

PR #203, opened 2023-09-11
Add dream.load_geant4_csv

### LINKED ISSUE #177: Load DREAM Geant4 csv files
https://confluence.esss.lu.se/display/DAM/DREAM+-+Load+csv+file

In my opinion, we should load the CSV file into a "schema" mimicking what we get from loading an DREAM NeXus file, to ensure that downstream code (after loading) has to be rewritten as little as possible.

jl-wynen:
Fixes #177

Mostly mimics the structure produced by #199 for NeXus files.

I tested it with `data_dream_HF_mil_closed_alldets_1e9.csv` provided as part of the requirements page on Confluence. The older file used as part of the tutorial in Scipp [here](https://scipp.github.io/tutorials/from-tabular-data-to-binned-data.html) does not work as it does not contain a `det ID` coordinate. This is required to split the data into the separate instrument components.

Here is what the loaded data looks like:

load_geant4_csv

There are some open questions:

- Can the files contain event weight? If the test file is meant to contain events, then the detector is illuminated completely uniformly.
- Can we publish the test data? If not, how do we test the loader?
- Should this combine module, segment, and counter into a 'subsegment' as in #199?
- Should we compute voxel positions from the event positions (average over events per voxel). As it stands, there is no voxel coordinate.
- The NeXus loader produces a `detector_number` coord. We cannot do this here. Is this ok?
- The loader in this PR makes a `position` coord for each event because that is needed by the instrument view and coord transforms. Should we drop `{x,y,z}_pos`? Or should we not produce `position`?

### DIFF
--- a/src/ess/dream/io/geant4.py
+++ b/src/ess/dream/io/geant4.py
@@ -2,7 +2,7 @@
 # Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
 
 import os
-from typing import Dict, Union
+from typing import Dict, Optional, Union
 
 import scipp as sc
 
@@ -54,18 +54,36 @@ def _split_detectors(
     data: sc.DataArray, detector_id_name: str = 'det ID'
 ) -> Dict[str, sc.DataArray]:
     groups = data.group(detector_id_name)
-    mantle = groups[detector_id_name, MANTLE_DETECTOR_ID].value
-    high_res = groups[detector_id_name, HIGH_RES_DETECTOR_ID].value
+    mantle = _extract_detector(groups, detector_id_name, MANTLE_DETECTOR_ID)
+    high_res = _extract_detector(groups, detector_id_name, HIGH_RES_DETECTOR_ID)
 
-    endcaps = sc.concat(
-        [groups[detector_id_name, i].value for i in ENDCAPS_DETECTOR_IDS], data.dim
-    )
-    endcap_forward = endcaps[endcaps.coords['z_pos'] > sc.scalar(0, unit='mm')]
-    endcap_backward = endcaps[endcaps.coords['z_pos'] < sc.scalar(0, unit='mm')]
+    endcaps_list = [
+        det
+        for i in ENDCAPS_DETECTOR_IDS
+        if (det := _extract_detector(groups, detector_id_name, i)) is not None
+    ]
+    if endcaps_list:
+        endcaps = sc.concat(endcaps_list, data.dim)
+        endcap_forward = endcaps[endcaps.coords['z_pos'] > sc.scalar(0, unit='mm')]
+        endcap_backward = endcaps[endcaps.coords['z_pos'] < sc.scalar(0, unit='mm')]
+    else:
+        endcap_forward = None
+        endcap_backward = None
 
     return {
-        'mantle': mantle.copy(),
-        'high_resolution': high_res.copy(),
-        'endcap_forward': endcap_forward.copy(),
-        'endcap_backward': endcap_backward.copy(),
+        key: val
+        for key, val in zip(
+            ('mantle', 'high_resolution', 'endcap_forward', 'endcap_backward'),
+            (mantle, high_res, endcap_forward, endcap_backward),
+        )
+        if val is not None
     }
+
+
+def _extract_detector(
+    detector_groups: sc.DataArray, detector_id_name: str, detector_id: sc.Variable
+) -> Optional[sc.DataArray]:
+    try:
+        return detector_groups[detector_id_name, detector_id].value.copy()
+    except IndexError:
+        return None
--- a/src/ess/dream/io/geant4.py
+++ b/src/ess/dream/io/geant4.py
@@ -59,8 +59,15 @@ def _split_detectors(
             dim=detector_id_name,
         )
     )
-    mantle = _extract_detector(groups, detector_id_name, MANTLE_DETECTOR_ID).copy()
-    high_res = _extract_detector(groups, detector_id_name, HIGH_RES_DETECTOR_ID).copy()
+    detectors = {}
+    if (
+        mantle := _extract_detector(groups, detector_id_name, MANTLE_DETECTOR_ID)
+    ) is not None:
+        detectors['mantle'] = mantle.copy()
+    if (
+        high_res := _extract_detector(groups, detector_id_name, HIGH_RES_DETECTOR_ID)
+    ) is not None:
+        detectors['high_resolution'] = high_res.copy()
 
     endcaps_list = [
         det
@@ -69,20 +76,14 @@ def _split_detectors(
     ]
     if endcaps_list:
         endcaps = sc.concat(endcaps_list, data.dim)
-        endcap_forward = endcaps[endcaps.coords['z_pos'] > sc.scalar(0, unit='mm')]
-        endcap_backward = endcaps[endcaps.coords['z_pos'] < sc.scalar(0, unit='mm')]
-    else:
-        endcap_forward = None
-        endcap_backward = None
+        detectors['endcap_forward'] = endcaps[
+            endcaps.coords['z_pos'] > sc.scalar(0, unit='mm')
+        ]
+        detectors['endcap_backward'] = endcaps[
+            endcaps.coords['z_pos'] < sc.scalar(0, unit='mm')
+        ]
 
-    return {
-        key: val
-        for key, val in zip(
-            ('mantle', 'high_resolution', 'endcap_forward', 'endcap_backward'),
-            (mantle, high_res, endcap_forward, endcap_backward),
-        )
-        if val is not None
-    }
+    return detectors
 
 
 def _extract_detector(
--- a/src/ess/dream/io/geant4.py
+++ b/src/ess/dream/io/geant4.py
@@ -2,6 +2,7 @@
 # Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
 
 import os
+from io import BytesIO, StringIO
 from typing import Dict, Optional, Union
 
 import numpy as np
@@ -12,7 +13,9 @@ HIGH_RES_DETECTOR_ID = sc.index(8)
 ENDCAPS_DETECTOR_IDS = tuple(map(sc.index, (3, 4, 5, 6)))
 
 
-def load_geant4_csv(filename: Union[str, os.PathLike]) -> sc.DataGroup:
+def load_geant4_csv(
+    filename: Union[str, os.PathLike, StringIO, BytesIO]
+) -> sc.DataGroup:
     """Load a GEANT4 CSV file for DREAM.
 
     Parameters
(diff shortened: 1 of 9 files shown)


celinedurniak:
Thanks for the implementation @jl-wynen .
Here are my answers to your final questions.

_Can the files contain event weight? If the test file is meant to contain events, then the detector is illuminated completely uniformly._
Not with the current version of the file writer in GEANT4. 
A voxel might be hit several times. But this will appear as a list of events whose coordinates will be contained within the boundaries of the detector voxels.

_Can we publish the test data? If not, how do we test the loader?_
The data can be published. But for testing, do you need a smaller file? I have one of 6.5MB

_Should this combine module, segment, and counter into a 'subsegment' as in
Add dream.load_nexus for latest NeXus files #199?_
Not at the moment.

_Should we compute voxel positions from the event positions (average over events per voxel). As it stands, there is no voxel coordinate._
This is not required at the moment. But once the detector_numbers are added (see reply below), it will be easy to map detected events with the corresponding detectors’ voxels

_The NeXus loader produces a detector_number coord. We cannot do this here. Is this ok?_
The CSV files contain metadata about wires, strips, counts. So, not having the detector number is not an issue. And once the mapping in the NeXus file has been approved by ECDC, detector_number could be added to the “csv” loader.

_The loader in this PR makes a position coord for each event because that is needed by the instrument view and coord transforms. Should we drop {x,y,z}_pos? Or should we not produce position?_
I do not have any preferences. What is the convention used for the other instruments?

celinedurniak (approved):

src/ess/dream/io/geant4.py
```diff
@@ -0,0 +1,89 @@
...
+
+def _adjust_coords(da: sc.DataArray) -> None:
+    da.coords['wavelength'] = da.coords['lambda']
+    da.coords['position'] = sc.spatial.as_vectors(
+        da.coords['x_pos'], da.coords['y_pos'], da.coords['z_pos']
+    )
+
+
+def _group(dg: sc.DataGroup) -> sc.DataGroup:
+    return dg.group('counter', 'segment', 'module', 'strip', 'wire')
```
SimonHeybrock: I don't think this is the correct order? I thought "strip" should be innermost (indexing voxels along a wire). Not sure about the others either.

src/ess/dream/io/geant4.py
```diff
@@ -0,0 +1,89 @@
...
+        _adjust_coords(det)
+
+    dg = sc.DataGroup({'instrument': sc.DataGroup(detectors)})
+    return _group(dg)
+
+
+def _load_raw_events(filename: Union[str, os.PathLike]) -> sc.DataArray:
+    table = sc.io.load_csv(filename, sep='\t', header_parser='bracket', data_columns=[])
+    table = table.rename_dims(row='event')
+    return sc.DataArray(sc.ones(sizes=table.sizes), coords=table.coords)
```
SimonHeybrock: @celinedurniak Should we add variances for the weights (all ones)?

src/ess/dream/io/geant4.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def _adjust_coords(da: sc.DataArray) -> None:
+    da.coords['wavelength'] = da.coords['lambda']
+    da.coords['position'] = sc.spatial.as_vectors(
+        da.coords['x_pos'], da.coords['y_pos'], da.coords['z_pos']
+    )
```
SimonHeybrock: I think we should pop in individual components. No need for duplication. Also remove `lambda`?

src/ess/dream/io/geant4.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def _group(dg: sc.DataGroup) -> sc.DataGroup:
+    return dg.group('counter', 'segment', 'module', 'strip', 'wire')
+
+
+def _split_detectors(
+    data: sc.DataArray, detector_id_name: str = 'det ID'
+) -> Dict[str, sc.DataArray]:
+    groups = data.group(detector_id_name)
```
SimonHeybrock: If you know all IDs in advance then providing `group` with it (as a variable) is faster, since it avoids having to find all unique IDs in the potentially large input data.

src/ess/dream/io/geant4.py
```diff
@@ -0,0 +1,89 @@
...
+        )
+        if val is not None
+    }
+
+
+def _extract_detector(
+    detector_groups: sc.DataArray, detector_id_name: str, detector_id: sc.Variable
+) -> Optional[sc.DataArray]:
+    try:
+        return detector_groups[detector_id_name, detector_id].value.copy()
```
SimonHeybrock: Is this copy necessary? Later you anyway group by component names, so another copy will be made.

src/ess/dream/io/geant4.py
```diff
@@ -0,0 +1,89 @@
...
+    mantle = _extract_detector(groups, detector_id_name, MANTLE_DETECTOR_ID)
+    high_res = _extract_detector(groups, detector_id_name, HIGH_RES_DETECTOR_ID)
+
+    endcaps_list = [
+        det
+        for i in ENDCAPS_DETECTOR_IDS
+        if (det := _extract_detector(groups, detector_id_name, i)) is not None
+    ]
+    if endcaps_list:
+        endcaps = sc.concat(endcaps_list, data.dim)
```
SimonHeybrock, line 97: `concat` (and thus copy) could be avoided by using `bin` instead of `group` above?

src/ess/dream/io/geant4.py
```diff
@@ -0,0 +1,89 @@
...
+        endcap_forward = endcaps[endcaps.coords['z_pos'] > sc.scalar(0, unit='mm')]
+        endcap_backward = endcaps[endcaps.coords['z_pos'] < sc.scalar(0, unit='mm')]
```
SimonHeybrock: Boolean indexing might be quite slow, have you considered using `bin`?

src/ess/dream/io/geant4.py
```diff
@@ -0,0 +1,89 @@
...
+        endcap_backward = None
+
+    return {
+        key: val
+        for key, val in zip(
+            ('mantle', 'high_resolution', 'endcap_forward', 'endcap_backward'),
+            (mantle, high_res, endcap_forward, endcap_backward),
+        )
+        if val is not None
+    }
```
SimonHeybrock: Seems simpler to just init the dict step by step above, avoiding, e.g., also the `else` for the endcap handling?

src/ess/dream/io/geant4.py
jl-wynen: I chose the order based on the respective dim-lengths where the longest is on the inside. I'm happy to change it if it makes more sense to follow a physical order. @celinedurniak ?

SimonHeybrock: I don't think `wire` is the longest actually, there are only 32 or so? Strips should be many more. Regardless, I think we should follow the logical order (I presume `detector_number` in NeXus files will also come in logical order)?

src/ess/dream/io/geant4.py
jl-wynen: Nothing gets grouped after this. In the case of mantle and high res, the object returned here is returned to the user.
I can change it such that it doesn't copy the endcaps because they are concatenated anyway.

src/ess/dream/io/geant4.py
jl-wynen: How so?

src/ess/dream/io/geant4.py
jl-wynen: No. And I would want to have a benchmark to see if this is actually faster because it needs multiple passes over the events, too.

src/ess/dream/io/geant4.py
SimonHeybrock: Instead of splitting the endcaps, and using concat to merge them again, based on:
```python
MANTLE_DETECTOR_ID = sc.index(7)
HIGH_RES_DETECTOR_ID = sc.index(8)
ENDCAPS_DETECTOR_IDS = tuple(map(sc.index, (3, 4, 5, 6)))
```
Make bin edges as `sc.array(dims=[detector_id_name], values=[3,7,8,9], unit=None_` or something like that? Then use `groups = data.bins(edges)`.

SimonHeybrock: ... but to be honest I do not know if that is really faster. So I'd say: leave it as it is, until we have evidence saying otherwise.

src/ess/dream/io/geant4.py
celinedurniak: The number of strips and wires depends on the "detector bank"
- Mantle: 256 strips, 32 wires
- Endcap backward and forward: 16 strips, 16 wires.
- HR: 32 strips, 16 wires

Strips are numbered along z axis. For the wires, it depends on the geometry (radially for the mantle and the endcaps). For the endcaps, there are also the SUMOs.

src/ess/dream/io/geant4.py
SimonHeybrock: Why does nit need multiple passes?

jl-wynen: At first I thought because it needs to find min and max. But we can just use `+/- Inf`.
Then it still needs to bin + copy each bin.

src/ess/dream/io/geant4.py
```diff
@@ -0,0 +1,114 @@
...
+        det
+        for i in ENDCAPS_DETECTOR_IDS
+        if (det := _extract_detector(groups, detector_id_name, i)) is not None
+    ]
+    if endcaps_list:
+        endcaps = sc.concat(endcaps_list, data.dim)
+        endcaps = endcaps.bin(
+            z_pos=sc.array(
+                dims=['z_pos'],
+                values=[-np.inf, 0.0, np.inf],
```
nvaytet, line 101: So this assumes that the origin is always in the middle of the detectors? i.e. the sample position. Is that always the case in Geant4?

tests/dream/io/geant4_test.py
```diff
@@ -0,0 +1,114 @@
...
+    assert isinstance(loaded, sc.DataGroup)
+    assert loaded.keys() == {'instrument'}
+
+    instrument = loaded['instrument']
+    assert isinstance(instrument, sc.DataGroup)
+    assert instrument.keys() == {
+        'mantle',
+        'high_resolution',
+        'endcap_forward',
+        'endcap_backward',
```
nvaytet, line 54: In a file that I got from @celinedurniak there were more entries than this: see https://github.com/scipp/ess/pull/184#discussion_r1227916455

Do we need to support both or is this now the new standard layout we will get?

src/ess/dream/io/geant4.py
nvaytet: @jl-wynen I think this is still missing?

src/ess/dream/io/geant4.py
celinedurniak: In GEANT4, the origin is at the sample position

tests/dream/io/geant4_test.py
jl-wynen: I think the SUMOs in that file correspond to the endcaps. Is this correct, @celinedurniak ?

celinedurniak: @jl-wynen you are correct. SUMOs are sub-divisions of the endcap detectors (i.e., the concentric partial rings) numbered from 3 to 6

nvaytet (approved):

SimonHeybrock (approved):

PR #205, opened 2023-09-14
Fix instrument view to not import plotting

jl-wynen:
I removed the return annotation because `scn.instrument_view` has no annotation either.


nvaytet (approved):

jl-wynen:
I'm going to merge this as the failures are unrelated. Will fix separately.

PR #206, opened 2023-09-14
Update scipp versions

jl-wynen:
This caused the CI workflow to fail

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -16,7 +16,7 @@ dependencies:
   - ipython==8.5.0
   - ipykernel==6.21.2
   - ipywidgets==8.0.4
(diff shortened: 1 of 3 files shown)


SimonHeybrock (approved):
Time to ditch Mantid from CI...

jl-wynen:
Mantid 6.7 and Scipp 23.08 are incompatible because they depend on different versions of TBB (courtesy of conda-forge global pinning). We need to run CI with Scipp 23.07 until the next Mantid release.
I didn't pin in the package definitions because users may want to use ess without Mantid and we want to avoid upper limits.

SimonHeybrock (approved):

PR #208, opened 2023-09-15
Ditch mantid from CI

jl-wynen:
Given the struggle in #206 and that I need Scipp 23.08.0 for #203, this PR removes Mantid from the CI builds.

There still are calls to `scn.load` in the loki and imaging modules. But they import Mantid only when called. And they don't seem to get called anywhere.


src/ess/external/powgen/data.py
```diff
@@ -14,38 +14,60 @@ def _make_pooch():
...
         registry={
+            # Files loadable with Mantid
             'PG3_4844_event.nxs': 'md5:d5ae38871d0a09a28ae01f85d969de1e',
             'PG3_4866_event.nxs': 'md5:3d543bc6a646e622b3f4542bc3435e7e',
             'PG3_5226_event.nxs': 'md5:58b386ebdfeb728d34fd3ba00a2d4f1e',
             'PG3_FERNS_d4832_2011_08_24.cal': 'md5:c181221ebef9fcf30114954268c7a6b6',
+            # Zipped Scipp HDF5 files
+            'PG3_4844_event.zip': 'md5:a644c74f5e740385469b67431b690a3e',
+            'PG3_4866_event.zip': 'md5:5bc49def987f0faeb212a406b92b548e',
+            'PG3_FERNS_d4832_2011_08_24.zip': 'md5:0fef4ed5f61465eaaa3f87a18f5bb80d',
```
SimonHeybrock, line 25: Interesting, how much of a difference does this make?

SimonHeybrock (approved):
I suppose this opens the way to remove conda from the CI (and thus a bunch of env files)? But I suppose this should be done separately, and only after we know if it is worth it (given that we are extracting technique-packages).

src/ess/external/powgen/data.py
jl-wynen: For the largest (vana, 4866), a 3x size reduction from 1.5GiB.

jl-wynen:
> I suppose this opens the way to remove conda from the CI (and thus a bunch of env files)? But I suppose this should be done separately, and only after we know if it is worth it (given that we are extracting technique-packages).

Yes, should be done separately, if at all. We may need Mantid for some sub packages it switching away from conda would make that harder in the future.

PR #214, opened 2023-10-23
Update for next scipp

jl-wynen:
`sans2d_reduction.ipynb` uses attrs explicitly in the notebook:
```python
monitors = {}
for key, da in dg.items():
    monitors[f'{key}-incident'] = da.attrs["monitor2"].value
    monitors[f'{key}-transmission'] = da.attrs["monitor4"].value
```
I did not change this yet as it requires changing the data file. Since this is being migrated into esssans anyway, I don't see a reason for updating it here.

The reflectometry notebooks store 'orso' attributes. I also dind't change this as the solution would be inferior to what can be done in essreflectometry.


docs/techniques/wfm/reducing-wfm-data.ipynb
```diff
@@ -700,7 +702,7 @@
     "    unit=events_no_wfm[\"sample\"][\"arrival_times\"].unit,\n",
     ")\n",
     "\n",
-    "ds_no_wfm = sc.Dataset(coords=coords)\n",
+    "data = {}\n",
```
nvaytet:
```suggestion
    "data_no_wfm = {}\n",
```

nvaytet (approved):

jl-wynen:
Changed it to also work with scipp 23.07 like I did for scippneutron. This should allow us to continue using mantid in CI.

nvaytet (approved):

PR #216, opened 2023-11-27
Running SANS workflow without direct_beam

wpotrzebowski:
For debugging purposes it is useful to run workflow without direct_beam. This is simple fix should enable this by setting direct_beam = None


SimonHeybrock:
Note: In `esssans` this is optional already: https://github.com/scipp/esssans/blob/a8b9ea8f901b64fd992246bbf5ef87c6d0e6399f/src/esssans/normalization.py#L114

nvaytet (approved):

